Issue #22, opened 2022-09-15
Feature request: Access EMHASS web interface via ingress


purcell-lab:
Ingress provides a means for an Home Assistant addon to expose its web service via the HA UI:

https://www.home-assistant.io/blog/2019/04/15/hassio-ingress/
https://developers.home-assistant.io/docs/add-ons/presentation/#ingress

Currently the EMHASS Web User Interface is accessed via http://homeassistant-ip:5000.

It would greatly simplify the user experience if EMHASS was configured to utilise the ingress interface.

My other add-on's are access via the Home Assistant UI and ingress interface:

ESPHome: http://odroid.local:8123/hassio/ingress/5c53de3b_esphome
Google Drive Backup: http://odroid.local:8123/hassio/ingress/cebe7a76_hassio_google_drive_backup
Terminal & SSH: http://odroid.local:8123/hassio/ingress/core_ssh

![image](https://user-images.githubusercontent.com/79175134/190319685-d9e56b5b-9ecf-426f-917b-b72bb443a3d5.png)
![image](https://user-images.githubusercontent.com/79175134/190319982-2176d3a9-5996-44db-99b3-5e9b111de88f.png)

PR #106, opened 2023-08-24
Add devcontainer and pave the way for ingress access

siku2:
Closes: #22 

image

The only way to see that it's working in this image is because the styles are loaded. But trust me, the buttons are working too.

---

This PR enables support for the Home Assistant ingress by reading the contents of the standard `X-Ingress-Path` header. This header is also used by many other reverse proxies, so this will allow them to function correctly on sub paths as well.
If there's no reverse proxy in front of emhass, it will work like before, so this won't break the non-addon setup.

There are also two other changes in this PR:

- I added a devcontainer that allows new contributors to get into the project more quickly. It's very bare-bones at the moment, but it can run the emhass web server without problems (which is how I was working on this project).
- Removed duplication from the "requirements*.txt" files by referencing the main dependencies in the web server one.

I'm of course happy to remove these changes from this PR if you wish.

There will be a separate PR over at [davidusb-geek/emhass-add-on](https://github.com/davidusb-geek/emhass-add-on) that's required before the ingress actually works.

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,0 +1,29 @@
+FROM mcr.microsoft.com/devcontainers/python:1-3.8-buster
+
+# copy the requirements file into the image
+COPY requirements.txt requirements.txt
+COPY requirements_webserver.txt requirements_webserver.txt
+
+# Setup
+RUN apt-get update \
+    && apt-get install -y --no-install-recommends \
+        coinor-cbc \
+        coinor-libcbc-dev \
+        gcc \
+        gfortran \
+        libhdf5-dev \
+        libhdf5-serial-dev \
+        libnetcdf-dev \
+        netcdf-bin \
+    && ln -s /usr/include/hdf5/serial /usr/include/hdf5/include \
+    && export HDF5_DIR=/usr/include/hdf5 \
+    && pip install netCDF4 \
+    && apt-get purge -y --auto-remove \
+        gcc \
+        libhdf5-dev \
+        libhdf5-serial-dev \
+        netcdf-bin \
+        libnetcdf-dev \
+    && rm -rf /var/lib/apt/lists/*
+
+RUN pip install -r requirements_webserver.txt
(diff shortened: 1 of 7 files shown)


davidusb-geek:
Again thanks for this.
Could you please add a quick explanation on how to run the devcontainer for testing.
I use the official Home Assistant devcontainer for testing the add-on part which actually runs a complete HA test instance. I run it from VSCode with no problem, is this something similar?

siku2:
@davidusb-geek, yes, but it doesn't include HA (yet :P). For now it gives you a full Python development environment that contains all the necessary dependencies to work on emhass.

If you want to test it out:

1. Head over to my fork, press on the friendly green "Code" button and create a new codespace: 
image
    Codespaces are based on devcontainers so this will give you a vscode instance in the browser loaded with the devcontainer that I added.

    You can, of course, also open the project locally  in vscode, but testing it with github is less of a hassle (and more secure).
2. Now you can just run `cd src && python -m emhass.web_server`. The server won't actually start as-is, because it doesn't find the configuration files in /app, but if you adjust the paths you will be able to fix it.

siku2:
Btw, there's a lot more that I want to improve wrt. the development flow and especially the user experience. I really want to use this add-on personally, but right now it's not quite in a state where I feel comfortable doing so. Keep in mind I'm only saying that because I'm actually willing to make the changes necessary to get to that point.

I'll try to find some time this weekend to write up an issue containing my thoughts to get the discussion going. I hope you're open to this idea, @davidusb-geek?

davidusb-geek (approved):

davidusb-geek:
Thanks for these changes and the instructions to test the devcontainer. I've just merged this.

davidusb-geek:
> Btw, there's a lot more that I want to improve wrt. the development flow and especially the user experience. I really want to use this add-on personally, but right now it's not quite in a state where I feel comfortable doing so. Keep in mind I'm only saying that because I'm actually willing to make the changes necessary to get to that point.
> 
> I'll try to find some time this weekend to write up an issue containing my thoughts to get the discussion going. I hope you're open to this idea, @davidusb-geek?

Yes of course this is open for contribution and improvements.
And we can discuss what you have in mind.

PR #118, opened 2023-10-02
Improvements in the documentation

g1za:
I would ask you to review the definitions of the battery_dynamics, as a bit unclear to me in regards to an explanation you gave in a post of the component thread. If I correctly understood we are just putting a constraint in the maximum allowed (dis)charge power of the battery (and not a constraint on the minimum allowed (dis)charge power), the use of the word "min" in the name of the parameter is a bit misleading.
Your input is welcome on this.

I also fixed what I think is a typo and expanded the section that explain the idea behind the reducing horizon in MPC.


davidusb-geek (approved):
Yes the definition for the dynamic power is correct for me.

PR #131, opened 2023-11-05
Upgrade to python 3.11

pail23:
- Upgrade to python version 3.11
- Upgrade numpy, scipy, pandas and tables to match python 3.11

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,4 +1,4 @@
-FROM mcr.microsoft.com/devcontainers/python:1-3.8-buster
+FROM mcr.microsoft.com/devcontainers/python:0-3.11
 
 COPY .devcontainer/setup.sh requirements.txt requirements_webserver.txt ./
 RUN ./setup.sh
--- a/.github/workflows/python-test.yml
+++ b/.github/workflows/python-test.yml
@@ -32,8 +32,8 @@ jobs:
     - name: Special dependencies for macos
       run: |
         brew install hdf5
-        pip3 install numpy==1.22.2
-        pip3 install tables==3.7.0
+        pip3 install numpy==1.26.0
+        pip3 install tables==3.9.1
       if: ${{ matrix.os == 'macos-latest' }}
         
     - name: Install dependencies
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -1,7 +1,7 @@
 wheel
-numpy==1.22.2
-pandas==1.4.1
-scipy<1.9.0
+numpy==1.26.0
+pandas==2.0.3
+scipy==1.11.3
 pvlib>=0.10.1
 protobuf>=3.0.0
 pytz>=2021.1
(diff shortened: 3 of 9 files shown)


davidusb-geek (changes requested):
This files also needs to be changed: docs/requirements.txt
But most importantly no checks were run. 
At least this workflow will need an update: .github/workflows/python-test.yml

davidusb-geek:
Tests are failing, probably related to those changes on TZ and datetimes?

pail23:
Many thanks for your reviews. The tests should be fixed now and docs/requirements.txt is updated.

davidusb-geek (approved):

davidusb-geek:
There are a lot of errors with Python 3.11. The Numpy version is not backwards compatible with Python 3.9 for example. How can we solve this? @pail23

davidusb-geek:
Python are getting these errors: https://community.home-assistant.io/t/emhass-an-energy-management-for-home-assistant/338126/1566

pail23:
As a quick thought: Would it be possible to migrate the emhass addon to python 3.11?

According this the [release notes](https://numpy.org/news/) of numpy 1.26, this version should be compatible with python 3.9-3.12.

I will have a look at the error messages an see if there are other options.

davidusb-geek:
> As a quick thought: Would it be possible to migrate the emhass addon to python 3.11?

Ok, but how to do this?

davidusb-geek:
The problem seems to be with the pvlib package using the h5py package...

pail23:
I did not test, but we might be able to update to python 3.11 by replaceing `bullseye` with `bookworm` in `
emhass-add-on/emhass/build.yaml
`

```

build_from:
  amd64: ghcr.io/home-assistant/amd64-base-debian:bookworm
  armhf: ghcr.io/home-assistant/armhf-base-debian:bookworm
  armv7: ghcr.io/home-assistant/armv7-base-debian:bookworm
  aarch64: ghcr.io/home-assistant/aarch64-base-debian:bookworm
codenotary:
  signer: notary@home-assistant.io
  base_image: notary@home-assistant.io
```

davidusb-geek:
Ok I will try that. Thanks!

pail23:
As an alternative:

According to this [article](https://stackoverflow.com/questions/74852225/attributeerror-module-numpy-has-no-attribute-typedict), it might help to explcitly install a newer h5py version (h5py==3.10.0)

davidusb-geek:
I added h5py==3.10.0 as a hard requirement but that does not seems to be the only problem.
There seems to be an issue with pip3 an PEP 668.
For now a solution was found using the flag --break-system-packages on pip3 install commands.
Let's see if it works.

davidusb-geek:
Build are failing for the armhf architectures.
It is a shame to leave those out, there is no full support for Python3.11 yet on that architecture I guess.
https://github.com/davidusb-geek/emhass-add-on/actions/runs/7241702872

PR #158, opened 2024-01-28
Make options optional

GeoDerp:
Home assistant optional parameters support


GeoDerp:
@davidusb-geek sorry was developing this and saw all you done with the pull requests. I think this code could be necessary for the emhass-add-on

GeoDerp:
@davidusb-geek sorry about that. I hope this all gets to you in time.

GeoDerp:
About my last commit. Got pytest tab inside of vscode working (inside of the vs-code virtual environment). There is also now a vs-code task to compile emhass, and two launch commands to debug emhass with and without add-on. Add-on launch requires key and url inserted in the launch.json  to operate however.

davidusb-geek:
Ok merging and then testing this locally.
I did tried the new launch.json but I did not understand how to launch a specific cofiguration inside the launch.json.
Do you know how to do this?

davidusb-geek (approved):

davidusb-geek:
> Ok merging and then testing this locally. I did tried the new launch.json but I did not understand how to launch a specific cofiguration inside the launch.json. Do you know how to do this?

Well don't about this I just found how to select the configuration that I want to run. It's in the debug pane to the left on VScode
Well explained here: https://gigi.nullneuron.net/gigilabs/working-with-vs-code-launch-configurations/

GeoDerp:
The thought would be that someone would pull the code, run it in vscode Dev container. (After vs code asks on the bottom right)
Press ctrl+ship+p, select run task, select build emhass 

Then on the left tab as you mentioned run the desired launch function and have the gui give out the results. The verible section when it fails is my favourite part.

Also the pytest panel (I believe below) should also work now too. 👍

PR #147, opened 2024-01-18
Add a function to predict a value from a csv file.

gieljnssns:
The story behind this pull request.
I keep a CSV file in which I store data from which I want to predict the number of heating hours.
I first tried to do this via a custom_component for home-assistant, but apparently it is [not possible to install](https://community.home-assistant.io/t/how-to-use-scikit-learn-with-a-custom-intigration/536939/6) `scikit-learn`.
Since the result of my prediction is to be used in emhass and the necessary dependencies are already installed in emhass, I decided to go this way.

This pull request contains a new method `csv-predict` with new parameters, here is an example of a rest command in home-assistant
```
predict_heating:
  url: http://localhost:5001/action/csv-predict
  method: POST
  content_type: "application/json"
  payload: >-
    {
      "csv_file": "prediction.csv",
      "independent_variables":["dd", "solar"],
      "dependent_variable": "hour",
      "sklearn_model": "LinearRegression",
      "csv_predict_entity_id": "sensor.voorspelde_uren",
      "csv_predict_unit_of_measurement": "h",
      "csv_predict_friendly_name": "Voorspelde uren",
      "new_values": [{{states("sensor.degree_day_prediction")}}, {{states("sensor.solcast_solaredge_forecast_today")}}]
    }
```

If you are open to accepting this pull request, I will also take the time to write some documentation. And if necessary, I would also like to try writing some tests.

Here is also a used CSV file.
[prediction.csv](https://github.com/davidusb-geek/emhass/files/13975642/prediction.csv)

PR #167, opened 2024-01-31
Utils tweak secret param handling

GeoDerp:


davidusb-geek:
Hi. Are these necessary to avoid breaking the add-on?
I'm not very fond of all these try/except, I was thinking about other ways to come around this.

GeoDerp:
These ones in particular, could get around easily by just setting a default value as the other option to get a statement (having only one get statement). 
It would just mean it won't try and take the value from the secrets file first. *(May be an issue if someone is running add-on mode not in HA. (which I'm guessing doesn't happen anyway))*
The other try statements on the lists would need to be looked into deeper for another way around if the list exists and what to replace with if else. 
Maybe a simple fix. When I get some free time I'll have a look.

davidusb-geek:
Is this related to theses changes?
https://community.home-assistant.io/t/emhass-an-energy-management-for-home-assistant/338126/1779?u=davidusb

davidusb-geek:
> Is this related to theses changes? https://community.home-assistant.io/t/emhass-an-energy-management-for-home-assistant/338126/1779?u=davidusb

Well this is more a missing key on that default dict

GeoDerp:
Both solar.forcast and solcast should be broken. That was the main reason for both pull requests initially. Would have to check if that's related to that comment however.

GeoDerp:
Closed due to #169 being merged. 
Could reopen if someone is wrong with #169.

davidusb-geek:
I did had to revert back to this solution after today's crash...

GeoDerp:
@davidusb-geek sorry to hear, what went wrong, anything I can fix?

davidusb-geek:
This issue https://github.com/davidusb-geek/emhass-add-on/issues/74#issue-2117220193

davidusb-geek:
So I went back with the option on this PR, tested the add-on properly and published a patch release.
I'm just about to release a new patched version to finally solve #154 that wasn't solved before because an error of mine.

GeoDerp:
> This issue https://github.com/davidusb-geek/emhass-add-on/issues/74#issue-2117220193

Would you like me to see if I can work out the issue?

davidusb-geek:
Well like we discussed I did found that code more proper.
Unfourtunately it failed but in the rush I didn't debugged that code completely and just switched to this PR.
It could be interesting to improve to that other more proper version. But we will need to test the add-on properly before release. I did tested the docer standalone before release but it wasn't sufficient.
So finally are you able to properly test the add-on using these dev containers?
I added this option to test an add-on-like option.json but while using the docker standalone mode which is much more easy and fast to compile and test: https://github.com/davidusb-geek/emhass/blob/0f5335ffb102ce57b34e90cded4e7a817f96bce3/src/emhass/web_server.py#L140

GeoDerp:
Personally I have been using this method to get semi close: 
https://github.com/davidusb-geek/emhass/pull/147#issuecomment-1925578995

But also there is a docker file in the pull request that also tries to emulate the add-on environment. Not as easily to use however.

GeoDerp:
Looks like the issue is with def_start_timestep/def_end_timestep. It's almost like it's being overridden somehow. I'm wondering if reverting it fixed the issue ?

GeoDerp:
Edit: I have worked out the problem and its possibly sooooo dumb

PR #168, opened 2024-01-31
New merge

PieterVO:


PieterVO:
I had opened the PR by accident.
Have been testing it for a while now, in Belgium we need to limit peak consumption below 2.5kW to avoid an extra cost.
It's been working well, it does "peak shaving" by discharging the battery if load prediction is above 2.5kW (so it makes sure to charge the batteries at the best times).
I can try to work on it during the weekend to merge the latest changes and open a new PR.

davidusb-geek:
Thanks that would be great. Interesting use case for consumption peak shaving.
However, the asked feature request is about peak shaving but for energy production, ie. to limit the kW sent to the grid. 
From what I see this is what you implemented in the modifications that you made.

PR #174, opened 2024-02-03
Issue #154: "perform_backtest": "false" has no effect

davidusb-geek:


davidusb-geek:
Hi @GeoDerp .
Will it be possible for you to check this? 
This is related to #154.
I'm able to reproduce the issue by just summoning:
```
curl -i -H "Content-Type:application/json" -X POST -d '{"perform_backtest": "False"}' http://localhost:5000/action/forecast-model-fit
```
Which don't take the passed `perform_backtest=False` and performs the back-testing nevertheless

davidusb-geek:
I added some tests to aid debug this in this PR but tests are passsing correctly

GeoDerp:
I'll see if I can have a look tomorrow (it's 1am currently)

GeoDerp:
I wonder if it's something dumb like: 

```
curl -i -H "Content-Type:application/json" -X POST -d '{"perform_backtest": "false"}' http://localhost:5000/action/forecast-model-fit
```
Or 

```
curl -i -H "Content-Type:application/json" -X POST -d '{"perform_backtest": false}' http://localhost:5000/action/forecast-model-fit
```

davidusb-geek:
I've just tested this:
```
curl -i -H "Content-Type:application/json" -X POST -d '{"perform_backtest": "false"}' http://localhost:5000/action/forecast-model-fit
```
And still using the `"perform_backtest": "true"`.

This one:
```
curl -i -H "Content-Type:application/json" -X POST -d '{"perform_backtest": False}' http://localhost:5000/action/forecast-model-fit
```
just exists as bad request

GeoDerp:
I think its the:
https://github.com/davidusb-geek/emhass/blob/d37aac3ae18fc54f340fb7d1ed4a67c353a5022f/src/emhass/machine_learning_forecaster.py#L166

Needs to be:
```
if perform_backtest is True:
```
for some reason.

davidusb-geek:
> I think its the:
> 
> https://github.com/davidusb-geek/emhass/blob/d37aac3ae18fc54f340fb7d1ed4a67c353a5022f/src/emhass/machine_learning_forecaster.py#L166
> 
> Needs to be:
> 
> ```
> if perform_backtest is True:
> ```
> 
> for some reason.

This is very strage. That is basic Python, I don't see how could that not work.
But I will test it with your PR.

PR #175, opened 2024-02-04
"perform_backtest": "false" has no effect Fix

GeoDerp:
https://github.com/davidusb-geek/emhass/issues/154
& 
https://github.com/davidusb-geek/emhass/pull/174


GeoDerp:
I have also added `is True` to all bools on the page just in case. Feel free to revert

davidusb-geek (approved):

PR #181, opened 2024-02-05
Fixed bug in #169

GeoDerp:
Related to, and possible fix for issue: https://github.com/davidusb-geek/emhass-add-on/issues/74


GeoDerp:
Ill see if I can resolve these errors

davidusb-geek:
This PR reinstates the `associations` list and dict and solves the issue that arised when we released this method. Is that right?

GeoDerp:
In theory yes 👍

GeoDerp:
This PR will need to be tested to see if its still functional and dosn't generate any bugs.

davidusb-geek:
> This PR will need to be tested to see if its still functional and dosn't generate any bugs.

Ok, we will probably integrate this in a future 0.9.0 release along with the new MLRegression class.
But yes will need to be tested.

Also requirements_addon.txt file is no longer needed right?

GeoDerp:
> > This PR will need to be tested to see if its still functional and dosn't generate any bugs.
> 
> Ok, we will probably integrate this in a future 0.9.0 release along with the new MLRegression class.
> But yes will need to be tested.
> 
> Also requirements_addon.txt file is no longer needed right?

Oh sorry. Yeah I don't believe so.
Does the web server requirements only be used for the devcontainer? If so, could that be removed also (replaced with manual run pip installs?

GeoDerp:
It may be good to keep the web server requirements if you want to specify the version. I'll change the Docker to use the file if that's the case?

davidusb-geek:
> It may be good to keep the web server requirements if you want to specify the version of flask, waitress and ploty. I'll change the Docker to use the file if that's the case? @davidusb-geek

Yes we will keep the web server requirements and erase the other requirements_addon file

GeoDerp:
> > It may be good to keep the web server requirements if you want to specify the version of flask, waitress and ploty. I'll change the Docker to use the file if that's the case? @davidusb-geek
> 
> Yes we will keep the web server requirements and erase the other requirements_addon file

Making a PR now.

PR #182, opened 2024-02-05
Optional: two additional testing environment options

GeoDerp:
This is an optional pull request uses VS-Code Dev container or Dockerfile (as an sandboxed environment) for debugging and testing the application. Both of which where designed with focus on running with Add-on mode. And both need an additional HA environment to operate *(no supervisor support for the moment)*

### Method 1: VS-Code Debug and Run
**Recommended steps to run are:**
- Pull in the git repo *(or Fork)*, open the root (`emhass`) foulder in VS-Code 
- VS-Code will asks if you want to run in a dev-container, say yes ([Docker environment](https://code.visualstudio.com/docs/devcontainers/containers) must be set up first). This give us a development environment we can test without messing with the host device. *(I believe codespaces does this by default)*
**To run and test:** *(via VS-Code Run and Debug)*
- Compile emhass by pressing `control+shift+p` > `Tasks: Run Task` > `EMHASS Install`. This has been set up in the [tasks.json](https://github.com/davidusb-geek/emhass/blob/master/.vscode/tasks.json) file. Re-run this every time a change has been made emhass before *run & debug*.
- Launch and debug the application selecting the [`Run and Debug`](https://code.visualstudio.com/docs/editor/debugging) tab  *(Ctrl+Shift+D.)* > `EMHASS run Addon`. This has been setup in the [Launch.json](https://github.com/davidusb-geek/emhass/blob/master/.vscode/launch.json). You will need to modify the `EMHASS_URL` *(http://HAIPHERE:8123/)* and `EMHASS_KEY`  *(PLACEKEYHERE)* before running to match your HA environment in order to work. you can also choose to run `EMHASS run`. This acts more like standalone mode and uses the secrets file *(however less tested)*
- If you want to change your parameters, I tend to do this in the options.json file before launch. 
- You can also run all the pytests by going to the [`Testing`](https://code.visualstudio.com/docs/python/testing) tab on the left hand side

### Method 2: Dockerfile run
**Alternatively:** This PR also adds a dockerfile that tries to emulate the Add-on environment as best as possible. This method is slower, however does not need VS-code to run *(just [Docker](https://www.docker.com/) environment)*
Run it on the emhass root dir with: *(It is recommended to run this on the host device, rather than an dev container)*

```bash
docker build -t emhass/testing -f Add-onEmulateDocker .
docker run -it -p 5000:5000 --name emhass-test docker.io/emhass/testing --url YOURHAURLHERE --key YOURHAKEYHERE
```
Delete docker container and image with:
```bash
docker rm  emhass-test
docker rmi emhass/testing
```
For Rapid testing try a command chain like this:
*Bash Example*
```bash
docker rm -f emhass-test ; docker build -t emhass/testing -f Add-onEmulateDocker . && docker run -it -p 5001:5000 --name emhass-test docker.io/emhass/testing --url YOURHAURLHERE --key YOURHAKEYHERE
```
*With example above use  `secrets_emhass.yaml` or  `options.yaml`  file to set time_zone,Lat,Lon & Alt*

#### Docker Alternative    
We can also pass location, key and url parameters via environment variables

```bash
docker build -t emhass/testing -f Add-onEmulateDocker .
docker run -it -p 5000:5000 --name emhass-test -e URL="YOURHAURLHERE" -e KEY="YOURHAKEYHERE" -e LAT="45.83" -e LON="6.86" -e ALT="4807.8" docker.io/emhass/testing
```
This allows the user to set variables prior to build
Linux Example: (in you host terminal)
``` bash
export EMHASS_URL="YOURHAURLHERE"
export EMHASS_KEY="YOURHAKEYHERE"
export TIME_ZONE="Europe/Paris"
export LAT="45.83"
export LON="45.83"
export ALT="4807.8"

docker build -t emhass/testing -f Add-onEmulateDocker .
docker run -it -p 5001:5000 --name emhass-test  -e EMHASS_KEY -e EMHASS_URL -e TIME_ZONE  -e LAT -e LON -e ALT  docker.io/emhass/testing
```

### Note:    
- both methods can use the `secrets_emhass.yaml` instead of passing secret parameters as environment variables/ arguments
- I cant vouch for the security of passing in the secrets via Environment Variables and arguments. User discretion advised. This is designed for testing with an secure host


GeoDerp:
Note this is still in development. Somethings act slightly differently then they should. However the hope is that this will be an assistance for developers that choose to use it

GeoDerp:
> This looks nice for add-on testing. To try to emulate even better the add-on environment we could use this image for docker: `ghcr.io/home-assistant/amd64-base-debian:bookworm`

Good Idea 👍

davidusb-geek:
Hi. Could consider your changes to the web_server.py.
Were they necessary to implement these new testing methods.
I like these new methods but your modifications to the web_server.py invalidates the use of the use_options helper that I'm using for debug (look at the current open review)

GeoDerp:
Yeah sorry about that. I Should be able to add it back in without a problem. Will check now however

GeoDerp:
@davidusb-geek Added support to use Environment Variables. I understand that this adds a bit of code.
Let me know if you want me to revert any aspects (ex: no ENV support for url and key)

GeoDerp:
I was also thinking about adding security with the secrets with the use of docker secrets. But since the docker container in question is for development only I thought it was outside of scope.

davidusb-geek:
> @davidusb-geek Added support to use Environment Variables. I understand that this adds a bit of code.
> Let me know if you want me to revert any aspects (ex: no ENV support for url and key)

From what I see in the code the default values are what it was before so it should be ok.
I need to take some time to test this further. This week-end...

davidusb-geek (approved):

davidusb-geek:
Merging...
Added this page to the wiki as a help following this PR: https://github.com/davidusb-geek/emhass/wiki/Contributing

PR #190, opened 2024-02-08
index/style added runtime parameter dictionary inputs

GeoDerp:
Related to: https://github.com/users/davidusb-geek/projects/1/views/1?pane=issue&itemId=52545404
### Changed
- index/style, added runtime parameter dictionary user inputs
- index/style, deleted deprecated center align tags
- style, tweaked button styling
- index/style, added comments, formatted and reordered

### Examples 
(*Dark mode*)
Input: **Box**
Screenshot from 2024-02-08 23-06-16
Input: **List**
Screenshot from 2024-02-08 23-06-04

### DIFF
--- a/src/emhass/static/style.css
+++ b/src/emhass/static/style.css
@@ -658,6 +658,7 @@ th {
   -webkit-animation: spin 2s linear infinite;
   animation: spin 2s linear infinite;
   vertical-align: text-top;
+  transition: 1s;
 }
 
 @-webkit-keyframes spin {
@@ -685,6 +686,10 @@ th {
   color: #158b00;
   vertical-align: text-top;
   font-size: 4.0em;
+  animation-name: fadeInOpacity;
+	animation-iteration-count: 1;
+	animation-timing-function: ease-in;
+	animation-duration: .5s;
 }
 
 .cross {
@@ -692,6 +697,28 @@ th {
   color: #a71515;
   vertical-align: text-top;
   font-size: 4.0em;
+  animation-name: fadeInOpacity;
+	animation-iteration-count: 1;
+	animation-timing-function: ease-in;
+	animation-duration: .5s;
+}
+
+@-webkit-keyframes fadeInOpacity {
+	0% {
+		opacity: 0;
+	}
+	100% {
+		opacity: 1;
+	}
+}
+
+@keyframes fadeInOpacity {
+	0% {
+		opacity: 0;
+	}
+	100% {
+		opacity: 1;
+	}
 }
 
 /* Alert box styling */
--- a/src/emhass/static/style.css
+++ b/src/emhass/static/style.css
@@ -408,6 +408,7 @@ h2 {
 }
 
 button:hover,
+.alert:hover,
 select:hover {
   color: darkblue;
   background-color: #e1e1e1;
@@ -430,6 +431,7 @@ input {
 }
 
 button,
+.alert,
 select {
   text-transform: none;
   border-width: 1px;
@@ -532,6 +534,7 @@ template {
 [hidden] {
   display: none;
 }
+
 /* -------------- */
 
 /* EMHASS added style */
@@ -539,6 +542,7 @@ template {
 button,
 .main-svg,
 .mystyle,
+.alert,
 table {
   border-radius: 7px;
   /* overflow: visible; */
@@ -558,12 +562,12 @@ button {
   margin: 15px 5px 0px 0px;
 }
 
-/* center align diagram and set overflow scroll*/
+/* center align diagram/table and set overflow scroll*/
 .plotly-graph-div,
 .table_div {
   overflow-x: auto !important;
   margin: 0 auto;
-  margin-bottom: 10px auto;
+  margin-bottom: 10px;
 }
 
 /* Set min size for diagrams  */
@@ -589,6 +593,7 @@ button {
   position: relative !important;
   float: right !important;
 }
+
 /* */
 
 .svg-container,
@@ -606,7 +611,6 @@ button {
   font-family: Arial;
   border-collapse: collapse;
   border-style: hidden;
-  /* box-shadow: 0 0 0 1px rgb(192, 192, 192, 50%); */
 }
 
 .mystyle td {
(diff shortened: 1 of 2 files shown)


GeoDerp:
This isn't polished yet, may still come back tomorrow and tweak.

davidusb-geek:
That actually look really really nice.
This could be very helpful for testing.
Nice work! You tell me when is ready so we can merge.

GeoDerp:
> That actually look really really nice.
> This could be very helpful for testing.
> Nice work! You tell me when is ready so we can merge.

In theory it's really to go now. However it will probably be good for me to have a look at it tomorrow morning and polish any possible bugs.

purcell-lab:
I couldn't see how to comment on the project board entry, so I'll put some thoughts here.

I help a lot of new users onboard with EMHASS.

From a user interface (UI) perspective, they load up EMHASS, then they go to this web page and start hitting the first button PerfectOpt (which doesn't appear to do anything), then they hit the second button DayAheadOpt (which also doesn't appear to do anything), then they randomly hit the ML buttons, which certainly doesn't appear to do anything. 

I think there are too many confusing options on this front page and would welcome an 'Advanced' tab that 'hides' the functions that require configuration.

The workflow in my mind would be to have a the Post Data box up front, as people need to inject values first, I like the second concept with a different box for each key. Then have a DayAheadOpt button, then have a Publish button.

Some sort of running log or user feedback when which of these steps is selected would also be really valuable, because as it stands there is zero feedback when a button is pressed..  If possible having the log running in a side bar would be ideal. I have a really ugly hack using iframes to display the log next to some buttons, but that would be far superior in the EMHASS ingress page.

image

GeoDerp:
I have actually thought about the logger on the page myself. However I have currently no idea how to achieve that. 

pseudo code thought would be:
- button gets triggered 
- logger logs inside the function triggered 
- the logs then get compiled and sent as an array as the function response. 
- python webserver reads that response and sends it as a return of some form to the webserver 
- JavaScript code interprets that code and appends it to a read only text box 
  - alternatively the python code in the webserver could loop the array like what's happening in the table generation

purcell-lab:
> However I have currently no idea how to achieve that.

Something like the Logger capability on the devices would be ideal.

Screenshot_20240209-121759

davidusb-geek:
> I couldn't see how to comment on the project board entry, so I'll put some thoughts here.

I just added you to the project members. But we can continue our discussion here is ok.

> From a user interface (UI) perspective, they load up EMHASS, then they go to this web page and start hitting the first button PerfectOpt (which doesn't appear to do anything), then they hit the second button DayAheadOpt (which also doesn't appear to do anything), then they randomly hit the ML buttons, which certainly doesn't appear to do anything.
> I think there are too many confusing options on this front page and would welcome an 'Advanced' tab that 'hides' the functions that require configuration.

Thank you Mark for the REX, this is very important to improve!

I think that yes maybe we can hide some buttons and the text box in an "advanced" tab to avoid confusing users with all these options. 

> The workflow in my mind would be to have a the Post Data box up front, as people need to inject values first, I like the second concept with a different box for each key. Then have a DayAheadOpt button, then have a Publish button.

Agree, at least the "day-ahead" and "publish" buttons and the text box to add more data shoul be up front.
Also we should take one step at a time. I think that the text box functionality that @GeoDerp just added is great as it is right now and I really like the two options: a whole dictionary of data and the list option.

> I have actually thought about the logger on the page myself. However I have currently no idea how to achieve that.

Of course that a logging functionality should be great. But I don't think it is completely necessary. I mean you can always have two browser tabs open, one for the webui and one for the logger.
But I do agree that we could add some responsiveness when pushing the buttons on the webui. An easy solution would be just to post a little message something like "Successful POST request sent to day-ahead optimization task" for example.
The logging functionality is great but requires a lot much more effort.

GeoDerp:
Will require testing and refinding.
Added the following capabilities:
- inputs keep state after refresh with localStorage, save on successful post
- error alert pops up for JSON parse error 
- removed form tags
- added status for post. tick and cross icons presented based on 201 post response 
- ml tasks have been hidden behind a button in the optimization tasks section *(can be moved)*

GeoDerp:
I may want to strengthen the null exceptions with input lists. Otherwise this works *(just forces one empty list to be on screen at all times)*
May also like to spend more time testing every scenario. Any assistance for the bug hunting will be appreciated \.

GeoDerp:
Update: found a few bugs. Will keep you posted when I fix them

GeoDerp:
I think Im pritty happy now.
For future reference. I believe we could modify the `return make_response(msg, 201)` to send back error log data if something went wrong. we could then tell the js to present the html alert box with the info.

GeoDerp:
Feel free to set this as an issue so we can keep track of it if desired.

> I think Im pritty happy now.
> For future reference. I believe we could modify the `return make_response(msg, 201)` to send back error log data if something went wrong. we could then tell the js to present the html alert box with the response error data.
> 
> js example being:
> ```js
>     async function formAction(action) {
>         var data = inputToJson()
>         if (data !== 0) { //don't run if there is an error in the data Json
>             showChangeStatus("loading") // show loading div for status
>             response = await fetch(`{{ basename }}/action/${action}`, {
>                 method: "POST",
>                 headers: {
>                     "Content-Type": "application/json",
>                 },
>                 body: JSON.stringify(data), //post can only send data via strings
>             }).then(saveStorage()) //save to storage if successful 
>             showChangeStatus(response.status) //replace loading, show tick or cross
>         }
>         else {
>             showChangeStatus("remove") //replace loading, show tick or cross with none 
>         }
>     }
> 
>     //function in control of status icons of post above
>     function showChangeStatus(status) {
>         var loading = document.getElementById("loader") //element showing status 
>         if (status === "remove") { //show loading logo
>             loading.innerHTML = "";
>             loading.classList.remove("loading"); //append class with loading animation styling
>         }
>         else if (status === "loading") { //show loading logo
>             loading.innerHTML = "";
>             loading.classList.add("loading"); //append class with loading animation styling
>         }
>         else if (status === 201) { //then show a tick 
>             loading.classList.remove("loading")
>             loading.innerHTML = `<p class=tick>&#x2713;</p>`
> 
>         }
>         else { //then show a cross 
>             loading.classList.remove("loading")
>             loading.innerHTML = `<p class=cross>&#x292C;</p>`
>         }
>     }
> ```
> 
> to 
> ```js
>     async function formAction(action) {
>         var data = inputToJson()
>         if (data !== 0) { //don't run if there is an error in the data Json
>             showChangeStatus("loading") // show loading div for status
>             response = await fetch(`{{ basename }}/action/${action}`, {
>                 method: "POST",
>                 headers: {
>                     "Content-Type": "application/json",
>                 },
>                 body: JSON.stringify(data), //post can only send data via strings
>             }).then(saveStorage()) //save to storage if successful 
>             showChangeStatus(response) //replace loading, show tick or cross
>         }
>         else {
>             showChangeStatus("remove") //replace loading, show tick or cross with none 
>         }
>     }
> 
>     //function in control of status icons of post above
>     function showChangeStatus(response) {
>         var loading = document.getElementById("loader") //element showing status 
>         if (response.status === "remove") { //show loading logo
>             loading.innerHTML = "";
>             loading.classList.remove("loading"); //append class with loading animation styling
>         }
>         else if (response.status === "loading") { //show loading logo
>             loading.innerHTML = "";
>             loading.classList.add("loading"); //append class with loading animation styling
>         }
>         else if (response.status === 201) { //then show a tick 
>             loading.classList.remove("loading")
>             loading.innerHTML = `<p class=tick>&#x2713;</p>`
> 
>         }
>         else { //then show a cross 
>             loading.classList.remove("loading")
>             loading.innerHTML = `<p class=cross>&#x292C;</p>`
>              document.getElementById("alert-text").textContent = "\r\n" + response.msg //or what the dict  response key is
>              document.getElementById("alert").style.display = "block";
>         }
>     }
> ```
>

davidusb-geek:
> Feel free to set this as an issue so we can keep track of it if desired.

Could you please open the issue your self, or better add a new card with this task to the kanban.
I'm not sure to completely understand the issue

davidusb-geek:
Merging and further testing...

davidusb-geek (approved):

Issue #191, opened 2024-02-11
Flask Return action error logs to webserver


GeoDerp:
In its current status the response message to the action is a status code Ex: https://github.com/davidusb-geek/emhass/blob/78309ad7b2a8f6f280deadec4bca378279476d00/src/emhass/web_server.py#L55
This however doesn't indicate if there was an error during action (and if there was, what was it). This only gives us the information that the action has completed *(with an 201 code)*

PR #190 implements alert box element that currently only facilitates JSON input format error:
https://github.com/davidusb-geek/emhass/blob/78309ad7b2a8f6f280deadec4bca378279476d00/src/emhass/templates/index.html#L200-L203

I believe we could modify the `return make_response(msg, 201)` to send back a 400 status code with an error log body (ex: as string[]) if something went wrong. we could then tell the JS to present the html alert box with the response error data. *(helping the user understand why it crashed without viewing the logs)*

js example being:

webserver.py (as an example)
```python
return make_response(["testa","testb","testc"], 400)
```

index.html

https://github.com/davidusb-geek/emhass/blob/78309ad7b2a8f6f280deadec4bca378279476d00/src/emhass/templates/index.html#L83-L122

to 
````js
    async function formAction(action) {
        var data = inputToJson()
        if (data !== 0) { //don't run if there is an error in the data Json
            showChangeStatus("loading") // show loading div for status
            response = await fetch(`{{ basename }}/action/${action}`, {
                method: "POST",
                headers: {
                    "Content-Type": "application/json",
                },
                body: JSON.stringify(data), //post can only send data via strings
            }).then(saveStorage()) //save to storage if successful 
            showChangeStatus(response) //replace loading, show tick or cross
        }
        else {
            showChangeStatus("remove") //replace loading, show tick or cross with none 
        }
    }

    //function in control of status icons of post above
    function showChangeStatus(response) {
        var loading = document.getElementById("loader") //element showing status 
        if (response.status === "remove") { //remove status icons
            loading.innerHTML = "";
            loading.classList.remove("loading"); 
        }
        else if (response.status === "loading") { //show loading logo
            loading.innerHTML = "";
            loading.classList.add("loading"); //append class with loading animation styling
        }
        else if (response.status === 201) { //then show a tick 
            loading.classList.remove("loading")
            loading.innerHTML = `<p class=tick>&#x2713;</p>`

        }
        else { //then show a cross 
            loading.classList.remove("loading")
[...]

Issue #186, opened 2024-02-06
Flask dynamically reload diagram after action (button press)


GeoDerp:
I can see this being done either via python as a response *(to completing the task)* or via JavaScript to refresh after X time after button press. 
I believe, it would be nice to have the latest data be presented on the webpage after pressing the respective button in the EMHASS webpage. 
The implementation of this statically would be the easiest. Telling the whole page to refresh. Ideally I would think the best method would be to refresh the diagram div's/sections dynamically on new data.

davidusb-geek:
That would be nice.
While you are at this take also a look at this if you can: https://github.com/users/davidusb-geek/projects/1?pane=issue&itemId=52141646

GeoDerp:
Had a 2 second look at this but I believe if we can redraw the {{injection_dict[plot]}} on every change or X seconds we might be able to the diagram to dynamically refresh. 🤷‍♂️

davidusb-geek:
Ok, I think should be the best option, but how to redraw on every change on web_server.py ?

GeoDerp:
That I don't know. We might be able to passively scan for changes to the file. However I haven't observed the code to see if redrawing the loop is even possible.

GeoDerp:
I'm wondering if, just before the response of a /action button, we call a function to rerun this?

```python
        injection_dict={}
    basename = request.headers.get("X-Ingress-Path", "")
    return make_response(template.render(injection_dict=injection_dict, basename=basename))
```
If this response is even responsible for drawing the diagrams. Only has a look at the code for a few mins so far

GeoDerp:
maybe something like this? 
https://stackoverflow.com/a/16881055

after button response we run a http get to get the latest html of the graphs and replace (using JS)

davidusb-geek:
Is this solved with the new version?

GeoDerp:
Sure is 👍

PR #196, opened 2024-02-13
Dynamic tables & error logger feedback 

GeoDerp:
The functional implementation of https://github.com/davidusb-geek/emhass/issues/186 & https://github.com/davidusb-geek/emhass/issues/191

Changes:
- new actionLogs.txt is generated in datapath storing sessions app.logger info
- on successful html button press, fetch is called to get  html containing latest table data 
- on html button press, If app.logger ERROR is present, send action log back and present on page. 

**Note:** As of first commit, this PR still needs refining, please wait to merge. Any suggestions appreciated.

### DIFF
--- a/src/emhass/web_server.py
+++ b/src/emhass/web_server.py
@@ -75,7 +75,7 @@ def index():
 #get actions 
 @app.route('/template/<action_name>', methods=['GET'])
 def template_action(action_name):
-    app.logger.info(" >> Sending rendered html template table data")
+    app.logger.info(" >> Sending rendered template table data")
     if action_name == 'table-template':
         file_loader = PackageLoader('emhass', 'templates')
         env = Environment(loader=file_loader)
--- a/src/emhass/web_server.py
+++ b/src/emhass/web_server.py
@@ -23,7 +23,6 @@ app = Flask(__name__)
 
 #check logfile for error, anything after string match if provided 
 def checkFileLog(refString=None):
-    isFound = False
     if (refString is not None): 
        logArray = grabLog(refString) #grab reduced log array
     else: 
@@ -45,7 +44,7 @@ def grabLog(refString):
             for x in range(len(logArray)-1): #find all matches and log key in isFound
                 if (re.search(refString,logArray[x])):
                    isFound.append(x)
-            if isFound is not None:
+            if len(isFound) != 0:
                 for x in range(isFound[-1],len(logArray)): #use isFound to extract last related action logs  
                     output.append(logArray[x])
     return output
--- a/src/emhass/web_server.py
+++ b/src/emhass/web_server.py
@@ -37,15 +37,17 @@ def checkFileLog(refString=None):
 
 #find string in logs, append all lines after to return
 def grabLog(refString): 
-    isFound = False
+    isFound = []
     output = []
     if ((data_path / 'actionLogs.txt')).exists():
             with open(str(data_path / 'actionLogs.txt'), "r") as fp:
                     logArray = fp.readlines()
-            for logString in logArray:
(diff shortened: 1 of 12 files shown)


GeoDerp:
@davidusb-geek feel free to test this and see what you think.

davidusb-geek:
Those are some great functionalities!
Ok, will do some testing this week-end.

davidusb-geek:
Ok no problem. Like I said I will only be able to test this week-end

purcell-lab:
@GeoDerp you are doing some fantastic work in extending the functionality of EMHASS.  

Thanks for your assistance! 🎉🎉

GeoDerp:
Two notes on the error logs side of things:
- it only prints app.logger outputs. 
  - If we continue going down this route we will just need to make sure we got app.logger.error for all likely fail cases. (If there isn't already)
   - else, we can modify it to print all console logs.
   - personally  I like the idea to get a little info from the webpage and get the rest via the console 
- it's currently set to app.logger.info but I may change to debug.

src/emhass/web_server.py
```diff
@@ -49,46 +102,64 @@ def action_call(action_name):
     input_data_dict = set_input_data_dict(config_path, str(data_path), costfun,
```
GeoDerp, line 102: There are a few places errors can occur if url or key is wrong, these lines are one. (then the other two are on ML Fit & Publish)

A suggested change for this line could be:
```python
    ActionStr = " >> Test HA connection..."
    app.logger.info(ActionStr)
    try:
        input_data_dict = set_input_data_dict(config_path, str(data_path), costfun, 
            params, runtimeparams, action_name, app.logger)
    except Exception as e :
        return make_response(grabLog(ActionStr), 400)
```

GeoDerp: The other method you could do is rap each action in a try and expect. Expect pasting exception in app.logger.error then returning make_response

GeoDerp:
I'm hopping that this logger implementation makes sense and can help others send informative information back to the html.

GeoDerp:
Sorry @davidusb-geek I closed the other PR as I merged those changes into this one as they are kinda related now.
Last things I added 
- url and key being incorect produces logger.error
- sensor being incorect produces logger.error
- days_to_retrieve default to 2 if number is under 2. It also produces  logger.warning

GeoDerp:
@davidusb-geek  I'm sorry in advance for all the testing

GeoDerp:
Just a heads up @davidusb-geek. The reason why i'm setting returns on the functions back to web_server, is that if there is an uncaught python error it stops the current function and runs the next in the action causing more errors (or just dosn't finish the function at all).
Whereas, if I return a string indicating an error has occurred in that function, I can stop that action and send back the log to the user on the website.

This is just how I would program it. Back when I did C in uni, we used to set returns on every function with 0 (fail) or 1 (pass). We could then check if that function worked every time and move to next. If not print issue.

GeoDerp:
@davidusb-geek  Im pritty happy with this PR now. It's not perfect but I don't believe there will be any problems.
I'm currently looking now to see how difficult it would be to set True or Fail returns on every logger.error and end of each function.

davidusb-geek:
Ok nice work!
I will at this and test it during the week-end

GeoDerp:
If you're interested; https://github.com/GeoDerp/emhass/tree/Error-Suppression-Testing is my testing branch for testing returns on all functions and if statements to make sure those functions ran without an issue. *(If not, return back to webserver and the user will see the logger.errors)*
I literally just pasted responses and if statements everywhere without thinking about it methodically 😅. 
But it might be a good idea of what my concept is. 🤔

davidusb-geek (approved):

PR #216, opened 2024-03-02
Docker, requirements and amhf support 

GeoDerp:


GeoDerp:
@davidusb-geek , will see how the test builds react to the change,

GeoDerp:
The current docker has been set up to use web_server requirements if standalone more only. Is this right?

GeoDerp:
I'll think move the web_server requirements to be used on all modes makes more sense?

GeoDerp:
If it those three packages are required for all modes, could both requirments.txt and requirements_webserver be merged like on EMHASS-add-on?

davidusb-geek:
> If it those three packages are required for all modes, could both requirments.txt and requirements_webserver be merged like on EMHASS-add-on?

Yes probably the best solution.

GeoDerp:
Sill testing. Just takes a while with pip.

GeoDerp:
Unfortunately I don't believe I can finish testing this until tomorrow. Sorry about that

GeoDerp:
This is just a test. Will continue this further possibly tomorrow. (if build works, feel free to test and pull).

davidusb-geek:
> This is just a test. Will continue this further possibly tomorrow. (if build works, feel free to test and pull). Otherwise test when I'm next available.
> 
> Requirement change: numpy<=1.26.0 to numpy<=1.26.2

No don't worry no rush.
It seems that armv7 is failing again.
A problem with the SciPy version 1.11.3?

davidusb-geek:
At some point in the past I was able to solve a lot of problems for arm arch by using this type of command to install packages:
```
RUN pip install --index-url=https://www.piwheels.org/simple --no-cache-dir -r requirements.txt
```

davidusb-geek:
@GeoDerp could you help us understand where the persistent files are stored now. They are now inside the running docker for the add-on? Before they were saved in the share folder on HA

GeoDerp:
> could you help us understand where the persistent files are stored now

The persistent data for a addon is in the `/data` folder. So  `DATA_PATH: '/data/'` https://github.com/davidusb-geek/emhass-add-on/blob/07366be0a4044dff043587b441e3d798e77d245a/emhass/config.yml#L31C2-L31C22

And the options.json is there automatically

davidusb-geek:
No this seems nice. At some point we were having issues with data persistance after reboot and add-on update. So we implemented this manual data saving on the share folder. With if your implementation works fine which seems to be the case for now is good for me. That time was just my lack of skills on docker that made use the manual approach. But I like this implementation more. We can still access the files if needed by going inside the container.

GeoDerp:
@davidusb-geek , this is still in testing. Just presenting progress. One of the issues with numpy & scipy is their requirements with OpenBLAS. For some reason (32bit) armhf doesn't have that package. Therefore we are building this manually for that architecture.

There is a option to disable the use of blas for both. However I'm unsure if EMHASS uses this package. So I'm assuming it does.

I have also bumped the versions these 2 packages for testing purposes. I assume this could be bumped down and still work however. Will need to test functionality either way *(if the build works)*.

GeoDerp:
> At some point in the past I was able to solve a lot of problems for arm arch by using this type of command to install packages:
> 
> ```
> RUN pip install --index-url=https://www.piwheels.org/simple --no-cache-dir -r requirements.txt
> ```

This may still be the way to go. When testing I found this method to be slow (if not failing to pull). So I removed it for testing,

GeoDerp:
There also also a argument for how many users are still using 32bit versions on there Raspberry pi. A lot of this could be simply fixed by removing `armhf` (which I believe is 32 bit floating point for Debian) support. Its to my understanding this will stop support for pi1,2,3(32bit) .So I understand If this architecture wants to be kept. The argument will be against how well EMHASS will run on a pi 1,2,3 , and is it worth keeping support for the time being. Happy to work around whatever.

davidusb-geek:
> There is also an argument for how many users are still using 32bit versions on their Raspberry pi. A lot of these errors could be simply fixed by removing `armhf` support. (which I believe is 32 bit floating point for Debian)
> 
> Its to my understanding this will stop support for pi1,2,3(32bit). So I understand If this architecture wants to be kept.
> 
> The argument will be against how well EMHASS will run on a pi zero, 1,2,3 , and is it worth keeping support for the time being. Happy to work around whatever. (This pr tries to keep armhf)

Yes I understand the argument. It there is no simple solution we will have no other choice than drop support for armhf.
But let's try to keep at least armv7 alive ;-)

GeoDerp:
@davidusb-geek , realized why I was so confused. the armhf image of `home-assistant/armhf-base-debian` was set to use packages from `armel`. I don't know why. but the latest commits added a line to check  armhf and change package architecture  to armhf.

GeoDerp:
Last commit, found that setting `--index-url=`  overrides pypy making the non arm architectures to struggle. Setting an if statement to only use `https://www.piwheels.org/simple` on `armhf` and `armv7`. Fixes that solution.

It looks like all architectures work now. both arm builds take only 10m now also. (I think that's thanks to the new numpy/scipy versions)

davidusb-geek:
Great! Nice finding of that solution

GeoDerp:
Just doing final testing now and ill do the last commit. *(ill do a pr for EMHASS-Add-on also)*

davidusb-geek:
> Just doing final testing now and ill do the last commit. _(ill do a pr for EMHASS-Add-on also)_

Ok no problem, we will merge when you say it's ready

GeoDerp:
fyi, still testing on: https://github.com/GeoDerp/emhass

GeoDerp:
Testing  process:
- armhf [ ]
  - standalone [ ]
  - Add-On 
- arm/v7 [ ]
  - standalone [ ]
  - Add-On 
- amd64 [ ]
  - standalone [ ]
  - Add-On    
- aarch64 [ ]  
  - standalone [ ]
  - Add-On

GeoDerp:
Ill see if I can test the other two tomorrow.

davidusb-geek:
> Ill see if I can test the other two tomorrow. @davidusb-geek if you have a good MPC example that could work for testing send it my way and ill use it to test MPC also.
> 
> Something like this?
> 
> ```shell
> curl -i -H 'Content-Type:application/json' -X POST -d '{"pv_power_forecast":[0, 70, 141.22, 246.18, 513.5, 753.27, 1049.89, 1797.93, 1697.3, 3078.93], "prediction_horizon":10, "soc_init":0.5,"soc_final":0.6}' http://192.168.3.159:5000/action/naive-mpc-optim
> ```
> 
> Let me know if there is anything else you will like me to test with above.

This example test is good enough for me for MPC.

GeoDerp:
Was unable to work on this today. Will try and get it done for the time you usually merge on the weekend.

davidusb-geek:
> Was unable to work on this today. Will try and get it done for the time you usually merge on the weekend.

Yes exactly, I will be able to take a more in depth look at all this this weekend. Thanks for these efforts!

GeoDerp:
All goes well the test will be done tomorrow. Otherwise the day after (Aus time). 👍

GeoDerp:
@davidusb-geek , im probably pretty happy with these changes. Ill see If I can test the other two by the time you test your self.

davidusb-geek:
> @davidusb-geek , im probably pretty happy with these changes. Ill see If I can test the other two by the time you test your self.

So the color-code table is updated?

GeoDerp:
im looking into virtualization of aarch64 and arm/v7 now.

GeoDerp:
@davidusb-geek , feel free to start testing this.

davidusb-geek (approved):
There are some typos on develop.md but I will fix them on my side.

Issue #236, opened 2024-03-13
WARNING list_hp_periods / list_peak_hours_periods


davidusb-geek:
I have this warning:
```
web_server WARNING list_hp_periods / list_peak_hours_periods_(start&end)_hours does not match number in num_def_loads, adding default values to parameter
```
But actually these are lists that should not necessarily need to be equal in length to the number of deferrable loads.
@GeoDerp Do you think you can take a look at this?

davidusb-geek:
Solved as per #237 
Closing

PR #237, opened 2024-03-14
utils.py removed list_hp_periods def load no check

GeoDerp:
Related to issue: https://github.com/davidusb-geek/emhass/issues/236
@davidusb-geek, I think this is what you meant?


davidusb-geek:
Exactly that was it

davidusb-geek (approved):

Issue #230, opened 2024-03-11
Huge icons on webui


davidusb-geek:
Hi @GeoDerp 
I don't know if you can see this same issue but in my webui there are now huge icons:

![image](https://github.com/davidusb-geek/emhass/assets/32926550/8ca11437-3a7c-4015-ba37-c95d95221029)

These icons where normal size on upper right corner on v0.8.1

GeoDerp:
Interesting, definitely nothing changed from 0.8.1.
By webui, it's like a PWA? 
I'll have a look to see if there is any max sizing on  the buttons. If not I'll set some. 
It may be a good idea to have a PR that just focuses on improved PWA support.

purcell-lab:
I get large icons on mobile but not my laptop.

![Screenshot_20240312-122546](https://github.com/davidusb-geek/emhass/assets/79175134/48c9f15a-3cb5-44d6-a8a1-73c9709a35e9)

GeoDerp:
Thanks @michaelpiron,  It looks like the same issue with some sort of styling not being propperly enabled. Im having a look at this now.

GeoDerp:
Is it possible that one of your can test this out: https://github.com/GeoDerp/emhass/tree/index-icon-fix

GeoDerp:
Just added an JS function that checks if stylsheet exists, if not adds it. *(redundancy)*

GeoDerp:
Dad had the idea it could be a DNS issue. He might not be wrong. Ie. If you're using DNS it may not be getting called probably inside the webui. The JS function may help with this I'm not sure.

GeoDerp:
Let me know if this has been fixed for you guys.  👍

michaelpiron:
> Thanks @michaelpiron, It looks like the same issue with some sort of styling not being propperly enabled. Im having a look at this now.
> 
> Its kinda annoying that this doesn't seem to be an issue on my end. Does the other elements styling look correct? _(buttons have shadows, etc)_. If you run http://yourhaip:5000, does it still look that same.
> 
> Also @michaelpiron , is your mobile an IPhone? Mine is a Android.

Hi @GeoDerp , could it be that you addressed the wrong person? I don’t recall this issue.

purcell-lab:
Yes I do get styling from the webUI interface.

![Screenshot_20240314-070110](https://github.com/davidusb-geek/emhass/assets/79175134/018c98b5-1a97-4543-989c-08be06cad4a2)

[...]

PR #239, opened 2024-03-16
Simplified fetch urls to relatives

GeoDerp:
This attempts to fix the issue in: https://github.com/davidusb-geek/emhass/issues/230
This should be checked before Pulled. Should validate with proxies. 
From testing with Nabu Casa, everything seems to work. Will wait for response with @davidusb-geek or @purcell-lab.

### DIFF
--- a/src/emhass/static/script.js
+++ b/src/emhass/static/script.js
@@ -1,8 +1,3 @@
-//before page load check for stylesheet
-document.onreadystatechange = async function() {
-    checkStyleSheets()
-}
-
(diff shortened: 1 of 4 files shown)


GeoDerp:
This may be a PR to revert/remove if there is any issues with Proxies. As `{{ basename }}` was added in the first place to resolve those issues. From what I have testes it seems to work. However it will be good to validate before merge.

davidusb-geek (approved):

PR #247, opened 2024-03-29
Dynamic paths with emhass config dict, some mlforcaster error suppression

GeoDerp:
Will do more checking tomorrow. Here early for progress and help.

### Added Dictionary var containing EMHASS paths
`emhass_config` dictionary has been created *(currently, containing `data` and `config` paths)* to try and add directory structure consistency though the program.

Currently, if the `data` path is outside the root EMHASS directory, certain errors may occur from some hard coded paths.
emhass_config tried to change this by  passing in a dictionary to *(I believe)* all sub models.
 
### MLForcaster error suppression 
Added some more error suppression. (more testing should be done)
MLForcaster wise, I'm still not supper confident with how MLForcaster works. My assumption is, if there is missing data in the sensor, that days worth gets rejected, Causing the "ValueError: All arrays must be of the same length" error. (between `forecast_dates` and `forecast_out.`). 
I have added  an if statement to notify the user of the issue, replacing the ValueError. This however doesn't fix the issue.

PR #249, opened 2024-04-01
patch fix cec path issue

GeoDerp:
@davidusb-geek


GeoDerp:
This should work. This is because the root path is `/data` for forecast.py

GeoDerp:
This is not ideal. As its placed in the data directory. I may still change to use parent paths to fix this.

davidusb-geek:
Ok I see. Many thanks!

GeoDerp:
#247 Tries to fix these types of issues.

davidusb-geek (approved):

PR #267, opened 2024-05-02
Readme Buttons

GeoDerp:
Displays some of the most used EMHASS links.
This PR is just a quick concept, can modify.

Screenshot from 2024-05-02 13-33-28


GeoDerp:
If this is liked we can also place these buttons for EMHASS-Add-on. The concept is that these buttons would assist in new users for gaining info, asking questions and creating issues.

GeoDerp:
The last commit tests with style tags instead of images. This unfortunately doesn't work in GitHub. Back to the drawing board.

GeoDerp:
Replaced buttons with SVG files. these SVG files are pretty mutch just the buttons inside a SVG tag.

davidusb-geek (approved):

Issue #261, opened 2024-04-23
Feature request: startup penalty


werdnum:
I am using emhass to control my pool pump based on Amber electricity prices. Often, the solution EMHASS comes up with switches the pump on and off repeatedly, which is probably a bit more switching than I'd like. The pump also runs faster (2x energy consumption) for the first few minutes, which probably negates the benefit from running at 22c/kWh rather than 23c/kWh.

I think I'd like to alter the algorithm to allow specifying a penalty (in minutes or cents) for each startup. That way, the optimiser can split the usage in two if it really helps, but won't stop and start every 30 minutes when the power price fluctuates slightly over the day.

davidusb-geek:
Turning on the option to make it a continuous load with just one startup is not an option for you?

werdnum:
It’s an option, but not the preferred one. Starting up twice is fine, but
it needs to be a big enough win to justify it.

— Andrew Garrett

werdnum:
I have this somewhat working in [a branch](https://github.com/werdnum/emhass/tree/start-penalty). Want to clean it up before sending a PR.

davidusb-geek:
Hi, ok nice feature!
The annoying part will be to add a new user-define input variable to control this.
I think that by default no penalty should be applied on the number of startups. Then if the user desires as it was your case, this option can be activated. 
This has been a constant struggle, add parameters so that this optimization can be highly configurable, but at the same each new parameter adds more complexity and the people complain about how all this is hard to setup.
But go ahead these are nice features. Try to add some lines to `test_optimization.py` to quickly test the desired/expected behavior of the feature.

davidusb-geek:
This is included in the just released new version. 
Closing this for now but reopen if eneded.

PR #291, opened 2024-05-23
Implement start penalty functionality

werdnum:
How to use:

* Set `def_start_penalty` to a list of floats. For each deferrable load with a penalty `P`, each time the deferrable load turns on will incur an additional cost of `P` • `P_deferrable_nom` • `cost of electricity at that time`. Intuitively, how many extra hours would the load have to run to avoid starting up again.
* Pass `def_current_state` in `runtime_params` to avoid incorrectly penalizing a start if a forecast is run when the load is already running.

I also made a couple of other supporting changes:

* Added unit tests for the relevant scenarios in `test_optimization.py`, plus one for the `set_def_constant` function. In doing so, I found a missing constraint for `set_def_constant` and added it to the optimizer (`P_def_bin2` was not constrained to be zero when `P_deferrable` was zero).
* Added comments for all the start-related constraints to help myself understand it.

### DIFF
--- a/src/emhass/optimization.py
+++ b/src/emhass/optimization.py
@@ -240,9 +240,15 @@ class Optimization:
                     self.optim_conf['weight_battery_charge']*P_sto_neg[i]) for i in set_I)
 
         # Add term penalizing each startup where configured
-        if 'def_start_penalty' in self.optim_conf and self.optim_conf['def_start_penalty']:
-            for k in range(self.optim_conf['num_def_loads']):
-                if k in self.optim_conf['def_start_penalty'] and self.optim_conf['def_start_penalty'][k]:
+        if (
+            "def_start_penalty" in self.optim_conf
+            and self.optim_conf["def_start_penalty"]
+        ):
+            for k in range(self.optim_conf["num_def_loads"]):
+                if (
+                    len(self.optim_conf["def_start_penalty"]) > k
+                    and self.optim_conf["def_start_penalty"][k]
+                ):
                     objective = objective + plp.lpSum(
                         -0.001
                         * self.timeStep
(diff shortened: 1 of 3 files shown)


werdnum:
Solves #261

davidusb-geek:
Hi @werdnum, is this ready to merge?
Probably some documentation needed somewhere.
From what I see this is only usable from the runtime parameters dictionary, but not from the configuration file?

werdnum:
I think it works in the configuration file. I’ll have to check to make sure.

Will take a look at documentation.

— Andrew Garrett

davidusb-geek:
Ok, I'm merging. But from what I see this will only works when passing `def_start_penalty` and `def_current_state` at runtime. For `def_current_state` it makes completely sense, but for `def_start_penalty` this could be set directly in the configuration file.
A small entry on the documentation would be needed don't you think.
I will merge and then try to update the documentation later on another PR

davidusb-geek (approved):

davidusb-geek:
Hi, this was merged but not correctly reviewed I guess.
I was trying to update the documentation an stumbled on this.
There are some  important things that I don't fully understand:

Let's start with this code for `def_start_penalty`:
```
        # Add term penalizing each startup where configured
        if ("def_start_penalty" in self.optim_conf and self.optim_conf["def_start_penalty"]):
            for k in range(self.optim_conf["num_def_loads"]):
                if (len(self.optim_conf["def_start_penalty"]) > k and self.optim_conf["def_start_penalty"][k]):
                    objective = objective + plp.lpSum(
                        -0.001 * self.timeStep * self.optim_conf["def_start_penalty"][k] * P_def_start[k][i] *\
                            unit_load_cost[i] * self.optim_conf['P_deferrable_nom'][k]
                        for i in set_I)
```
This `self.optim_conf["def_start_penalty"]` is supposed to be a list of floats. 
But it seems to me that you are treating this a list of `bool`. 
For example here `if ("def_start_penalty" in self.optim_conf and self.optim_conf["def_start_penalty"])`.
Another term I don't understand is `len(self.optim_conf["def_start_penalty"]) > k`. If that parameter is a list of floats with a single float for each deferrable then when is this condition supposed to `True`? And again this is treating `self.optim_conf["def_start_penalty"][k]` as if it were a `bool`.

Same comment for the `def_current_state` parameter.
This condition can never hold if that parameter is a list of `bool` with a single `bool` for each deferrable load: `if ("def_current_state" in self.optim_conf and len(self.optim_conf["def_current_state"]) > k):`
I'm a bit lost here.

werdnum:
Hi David, thanks for taking a look and sorry I haven't got to the
documentation yet!

Happy to help you understand the code and of course if there's something
I've done that's hard to understand, please let me know how I can make it
better.

I am relatively confident in the basic functionality because of the
expanded unit tests, but there could be small bugs of course.

— Andrew Garrett


On Mon, 10 Jun 2024 at 05:34, David ***@***.***> wrote:

> Hi, this was merged but not correctly reviewed I guess.
> I was trying to update the documentation an stumbled on this.
> There are some important things that I don't fully understand:
>
> Let's start with this code for def_start_penalty:
>
>         # Add term penalizing each startup where configured
>         if ("def_start_penalty" in self.optim_conf and self.optim_conf["def_start_penalty"]):
>             for k in range(self.optim_conf["num_def_loads"]):
>                 if (len(self.optim_conf["def_start_penalty"]) > k and self.optim_conf["def_start_penalty"][k]):
>                     objective = objective + plp.lpSum(
>                         -0.001 * self.timeStep * self.optim_conf["def_start_penalty"][k] * P_def_start[k][i] *\
>                             unit_load_cost[i] * self.optim_conf['P_deferrable_nom'][k]
>                         for i in set_I)
>
> This self.optim_conf["def_start_penalty"] is supposed to be a list of
> floats.
> But it seems to me that you are treating this a list of bool.
> For example here if ("def_start_penalty" in self.optim_conf and
> self.optim_conf["def_start_penalty"]).
>
This is a common idiom in python - a nonzero value evaluates to true, a
zero value (or empty string, None, etc) evaluates to false. I'm checking
whether there is any value (so I don't just assume it's a list and crash if
there's nothing there). This is what's recommended by the Google style
guide (
https://google.github.io/styleguide/pyguide.html) but happy to follow your
lead if you have another style preference.

>
> Another term I don't understand is len(self.optim_conf["def_start_penalty"])
> > k. If that parameter is a list of floats with a single float for each
> deferrable then when is this condition supposed to True?
>
Here I'm trying to check whether the list of start penalties is long enough
to have a value for the current one. I don't want to try to get a staff
penalty value for load 2 (3rd in the list) if there are only 2 items in the
list. Len() returns the number of values, and k is the zero based index of
the deferrable load. So if there are 2 deferrable loads, K could be 0 or 1.

> And again this is treating self.optim_conf["def_start_penalty"][k] as if
> it were a bool.
>
Right, I don't want to do any processing for start penalty if it's set to
zero.

> Same comment for the def_current_state parameter.
> This condition can never hold if that parameter is a list of bool with a
> single bool for each deferrable load: if ("def_current_state" in
> self.optim_conf and len(self.optim_conf["def_current_state"]) > k):
>
Remember K will take all values from 0 to num_deferrable_loads exclusive,
so this will be false only when the def_current_state list if under sized.

PR #334, opened 2024-08-02
Standalone and Add-on merge

GeoDerp:
Related to https://github.com/davidusb-geek/emhass/issues/269

PR #336, opened 2024-08-03
Fix issue where thermal runtime parameters were not being propagated into optim_conf

werdnum:
Seems this got lost at some point in submitting the thermal configuration.

It's needed to specify the start temp and desired temperatures in runtime params.


werdnum:
Not sure why the tests aren't passing, will have to look into that...

davidusb-geek:
Hi @werdnum, thanks for this.
The tests seem to be passing. There is just a random build error that happens from time to time on Python 3.10 test.
Could you convert this to a final PR?

davidusb-geek:
Hi @werdnum, we have just merged a huge PR #334 and it probably messed up this branch.
Can you please update/rebase and fix the eventual conflict so I can merge your PR?

davidusb-geek:
@GeoDerp These tests are failing possibly by how we now will deal with the configuration parameters.
There is the message "config.json does not exist, or has not been passed" so test parameters are not being passed correctly?

GeoDerp:
> @GeoDerp These tests are failing possibly by how we now will deal with the configuration parameters.
> There is the message "config.json does not exist, or has not been passed" so test parameters are not being passed correctly?

Ill see if I can have a look into this today, The message about config.json not existing is more of a redundant message for the tests. The code falls back to the default Params in confid_defaults.json. the message was designed to let users know that that there is no config.json but in result sends a warning on every test.

PR #357, opened 2024-10-22
ADD OptimalConfiguration as class instead of dict

abriotde:
Bonjour,

J'aimerai rendre le code un poeu plus "objet" cela améliorera la lisibilité. Il faut y aller progressivement. C'est pourquoi je propose ceci qui est assez simple.

PS : Pour moi, cela me permet après l'avoir installé via PIP de changer dynamiquement les "deferrables" sans devoir changer le fichier de configuration/recharer la configuration. (Cf : https://github.com/abriotde/openhems-sample/blob/main/scripts/emhass/test_emhass.py)

Cordialement,
Albéric

### DIFF
--- a/src/emhass/optimal_configuration.py
+++ b/src/emhass/optimal_configuration.py
@@ -1,0 +1,58 @@
+
+class Deferrable:
+
+	def __init__(self, nominal_power, total_hours, start_timestep=0, end_timestep=0, constant = False, start_penalty = 0.0, as_semi_cont = True, load_config = False):
+		self.power = nominal_power
+		self.total_hours = total_hours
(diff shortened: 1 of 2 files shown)


davidusb-geek:
There is a current HUGE ongoing work by @GeoDerp that will greatly affect and improve how we deal with the setup phase. We need to see how this could possibly fit into all that.

davidusb-geek:
I don't think that we will be moving in this direction anytime soon.
Thanks for your contribution

abriotde:
Ok, never mind, I realized a great integration of Emhass from source witch I will maintain.

PR #367, opened 2024-10-31
Recommendation, bump get_legacy_yaml_parse to get_yaml_parse

GeoDerp:
Just an idea. The commented lines will take the config_emhass.yaml (instead of config.json) as the config and auto format the parameters to the new definitions.


davidusb-geek:
Yes, I was (kind of) aware that it was not the smartest move.
I did that to quickly fix the `test_retrieve_hass.py` that was broken for the case when `get_data_from_file=False` that I actually use sometimes for quick debugging.
I will test this recommendation

davidusb-geek (approved):

PR #374, opened 2024-11-13
Thermal model updates: feasibility & unit tests

werdnum:
Three significant changes:

1. Add an option (to become the default?) to use penalties rather than constraints for thermal conditions. This will replace a bunch of hacks I'm using to avoid trying to ask for something impossible from the solver.
2. Rework overshoot temperature - instead of "never exceed this temperature", we want "if we do exceed this temperature, it shouldn't be our fault" - implemented as "load must be off if we overshoot"
3. Added relatively comprehensive unit tests to cover the above behaviours plus more.


davidusb-geek:
Hi @werdnum, I finally gathered some small spare time to look at this. It seems like a very nice contribution, this may help with unexplainable infeasible results we get some time. I should probably think about moving to this type of reasoning and implementation: penalizing instead of constraining. On downside of this is that if you ask the "imposible" to the algorithm then it will converge but the result may possible have no real physical meaning.

What about this PR, it is in a Draft version? Do you think that you could package this into a meargeable PR?

Micr0mega:
Any updates on dropping the draft status? I'd love this, since I'm using the thermal model to control my heat pump, which is connected to a solar boiler. The temperatures will, in summer time, for sure rise beyond the overshoot temperature, without running the heat pump itself. This currently will mean my model will be infeasible very often.

GeoDerp:
@sourcery-ai review

davidusb-geek:
@sourcery-ai review

PR #399, opened 2024-12-24
Added support to retrieve configuration data directly from HA

davidusb-geek:
In this PR two main new features are introduced:

1. Retrieving basic configuration data directly from Home Assistant

This will avoid the need to define some of this parameters in the secrets file.
For now implemented automatic currency and temperature units.
But in the future we could also use the imported altitude, latitude, longitude, etc.

2. A new load forecast method based on typical household consumption

This new method will be useful to new installations as it is based on just a typical household consumption and some basic statistic methods. So a new user installing EMHASS don't need to provide data from a sensor to use the current default method which is the naive persistance.
To completely make it very easy to install we need to do something similar with the PV. A first option could be to add a on/off switch to consider a PV installation or not (as we are doing for the battery), then make the no PV installation option as the default. 
With these two changes a new install needs minimal configuration and we can obtain optimization results quickly.


davidusb-geek:
Hi @GeoDerp, I've just added in this PR a new load forecast method (look at the description in the first post.
Could you please quickly check if I'm not forgetting anything? What I want to do is make this new method the default method, as it does not require any external data retrieved from HA or any other source. The hope is that this will improve the experience for new users and make it easier to setup

I will open a second PR to implement the second part of this "make-it-easier" approach: make PV installation optional

PR #407, opened 2024-12-29
configuration page init pv switch

GeoDerp:


GeoDerp:
@davidusb-geek , My biggest concern is that his will likley be a braking change for current users who use PV (since the default is false ).

GeoDerp:
Last commit is a possible idea arround the issue. It still passes `get_weather_forecast` if the user passes data via runtime.

davidusb-geek:
> @davidusb-geek , My biggest concern is that his will likley be a braking change for current users who use PV (since the default is false ). Im thinking now for a possible solution.
> 
> We could hardcode in a function that checks if `pv_module_model` or `pv_inverter_model` has been defined. _(dont equal the config_defaults.json default values)_ If true, set pv to true and save.

Yes this will be a breaking change but we will be transparent about it and the fix will just need to turn on that new switch.
The new hardcoded function will be the elegant solution. If you can spend some time on that just tell me, otherwise I will release with the explicit breaking change

PR #408, opened 2024-12-29
generate test image from master

GeoDerp:
In theory, this additional github action will run every time a commit has been pushed into the master branch.
It will automatically generate a `test` tagged image, which could be used to test the latest changes on emhass-add-on / docker, prior to pushing a new release. 

*It also makes it easier for other users to check the changes and add comments before releasing.*

### DIFF
--- a/.github/workflows/docker-build-test.yaml
+++ b/.github/workflows/docker-build-test.yaml
@@ -1,5 +1,5 @@
-#This action test builds EMHASS standalone Docker images, in each architecture.
-name: "Test Docker"
+#This action test builds EMHASS Docker images, in each architecture.
+name: "Test Building Docker Image"
 
 on:
   push:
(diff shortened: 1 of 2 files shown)


davidusb-geek:
And how can we launch this image?
Something like this?
```
docker run -it -p 5000:5000 --name emhass-test -v ./options.json:/data/options.json emhass/test
```
With some secrets defined in `options.json` to access the HA instance

GeoDerp:
Should be the exact same as the original. however you will be pulling from `ghcr.io/davidusb-geek/emhass:test` instead of `ghcr.io/davidusb-geek/emhass:v0.11.4`
you can also select the `emhass-test` version from emhass-add-on

just hold for two secs before merging however, I think I found a tag bug.

GeoDerp:
looks good now 

This is what it looks like on emhass-add-on

davidusb-geek:
> looks good now 👍
> 
> p.s. This is what it looks like on emhass-add-on Screenshot From 2024-12-29 23-55-53

Great, thanks for this!

davidusb-geek (approved):

PR #411, opened 2024-12-30
get_ha_config add error handling

GeoDerp:
@davidusb-geek my guess is that this will help with the current bug (testing now)


GeoDerp:
I think this might have fixed it.

davidusb-geek (approved):

PR #419, opened 2025-01-04
CEC append custom EMHASS modules and inverters

GeoDerp:
Ability to add inverters / modules that otherwise wont be added to SAM.
*(Will review my work tommorow when im more awake)*

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -43,7 +43,7 @@ dependencies = [
 
 [project.optional-dependencies]
 docs = ["sphinx", "sphinx-rtd-theme", "myst-parser"]
-test = ["requests_mock", "pytest", "coverage", "snakeviz", "ruff"]
+test = ["requests_mock", "pytest", "coverage", "snakeviz", "ruff", "tabulate"]
 
 [tool.setuptools.packages]
 find = { where = ["src"] }
--- a/scripts/save_pvlib_module_inverter_database.py
+++ b/scripts/save_pvlib_module_inverter_database.py
@@ -32,9 +32,11 @@ logger, ch = get_logger(__name__, emhass_conf, save_to_file=False)
 
 if __name__ == "__main__":
     save_new_files = True
+
     logger.info("Reading original outdated database from PVLib")
     cec_modules_0 = pvlib.pvsystem.retrieve_sam("CECMod")
     cec_inverters_0 = pvlib.pvsystem.retrieve_sam("cecinverter")
+
     logger.info("Reading the downloaded database from SAM")
     cec_modules = pvlib.pvsystem.retrieve_sam(
         path=str(emhass_conf["data_path"] / "CEC Modules.csv")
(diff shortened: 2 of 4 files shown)


GeoDerp:
I could look into automating this procedure also. But that may be a future goal.

davidusb-geek:
Good idea.
My question is: where do you obtain the parameters for the new inverters?

RafAustralia:
> Good idea. My question is: where do you obtain the parameters for the new inverters?

Hi David

The parameters could be obtained either from manufacturers directly if they are willing to share / provide or simply google / chatgpt as most parameters are publicly searchable - all but certs such as C0, C1, C2, C3 which can be approximated for now. 

Ive just emailed Sungrow to confirm these so we can have a better idea if what Ive found is correct.
Given we deal with China they are on holidays and may take some time to revert back. But they will. 

That will update Sungrow inverters to new data and make emhass much much more accurate. 

Raf

GeoDerp:
@davidusb-geek @RafAustralia , I have added an git action that will now *(or should)* automatically update cec_inverters.pbz2 and cec_modules.pbz2 files.
This will mean you should not need to do it manually in the future @davidusb-geek

GeoDerp:
Sorry about all the commits. Thats what I get for going too fast.

GeoDerp:
PR is good to go 👍

RafAustralia:
Let me just say that is bloody exciting! 
well done both of u lot.

davidusb-geek:
> > Good idea. My question is: where do you obtain the parameters for the new inverters?
> 
> Hi David
> 
> The parameters could be obtained either from manufacturers directly if they are willing to share / provide or simply google / chatgpt as most parameters are publicly searchable - all but certs such as C0, C1, C2, C3 which can be approximated for now.
> 
> Ive just emailed Sungrow to confirm these so we can have a better idea if what Ive found is correct. Given we deal with China they are on holidays and may take some time to revert back. But they will.
> 
> That will update Sungrow inverters to new data and make emhass much much more accurate.
> 
> Raf

The parameters are obtained my using real measurements in a laboratory test bench. So yes either the manufacturer or a lab entity as the NREL. These are not easy parameters to obtain. But yes they can be approximated.

davidusb-geek:
> PR is good to go 👍

Great!

davidusb-geek (approved):

GeoDerp:
> It seems that the "Update EMHASS PVLib databases" job is failing when trying to git push at the end

@davidusb-geek 's completely fine.
Sorry I was going to mention that this action will fail if no update needs to be done to the database.

Since u have already updated them it will fail

GeoDerp:
> > It seems that the "Update EMHASS PVLib databases" job is failing when trying to git push at the end
> 
> @davidusb-geek 's completely fine.
> Sorry I was going to mention that this action will fail if no update needs to be done to the database.
> 
> Since u have already updated them it will fail

I might look into the future to edit the action so that it passes even tho it failed

RafAustralia:
> > > Good idea. My question is: where do you obtain the parameters for the new inverters?
> > 
> > 
> > Hi David
> > The parameters could be obtained either from manufacturers directly if they are willing to share / provide or simply google / chatgpt as most parameters are publicly searchable - all but certs such as C0, C1, C2, C3 which can be approximated for now.
> > Ive just emailed Sungrow to confirm these so we can have a better idea if what Ive found is correct. Given we deal with China they are on holidays and may take some time to revert back. But they will.
> > That will update Sungrow inverters to new data and make emhass much much more accurate.
> > Raf
> 
> The parameters are obtained my using real measurements in a laboratory test bench. So yes either the manufacturer or a lab entity as the NREL. These are not easy parameters to obtain. But yes they can be approximated.

100% agree. 
but actually more then half of these are publicly available if you search with exception of C values. But Im trying to get these confirmed also as well. 
so with a bit of effort we will update the list to new data which is exciting

PR #423, opened 2025-01-07
Multiple roof_id support

radim2tej:
Multiple roof_id support. The gains of each roof are added together. Individual Roof_id are separated by commas or white characters.


davidusb-geek:
Hi @radim2tej. First of all thanks for the contribution.
Now could you please add a small test in test_forecast.py?
Just add a new method treating this case, follow the same approach as the other test methods in that file. You have everything you need to test from the attributes in the SetUp method.

davidusb-geek (approved):

PR #450, opened 2025-02-07
Fix SonarQube error

GeoDerp:


GeoDerp:
@davidusb-geek , given time ill look into SonarQube web recommendations later. Im also looking into how we can send more data via rest. ie: `curl ... -H "Transfer-Encoding: chunked"`

davidusb-geek:
Ok 

> Im also looking into how we can send more data via rest. ie: `curl ... -H "Transfer-Encoding: chunked"`

What do you mean by this? A new method to make it simpler to send data un a curl command?

GeoDerp:
> Ok
> 
> > Im also looking into how we can send more data via rest. ie: `curl ... -H "Transfer-Encoding: chunked"`
> 
> What do you mean by this? A new method to make it simpler to send data un a curl command?

So currently users are hitting the limitations of sending too much data in one packet. using `-H "Transfer-Encoding: chunked"` in the curl command tells the call to split the data into chunks. However I defiantly still need to look into it.

GeoDerp:
This is an example of this: https://github.com/davidusb-geek/emhass/pull/450/commits/7ca105f8d4c4273efd7e3105d30f5162d2528114

GeoDerp:
> Hi @GeoDerp, is this ready to merge? 

Should be, would like to do one more double check that nothing is broken in the js files however.

GeoDerp:
@davidusb-geek , cant change google osv just yet. They havent released a new version. Will keep en eye out for when it happens.

GeoDerp:
@davidusb-geek there may be a cpu maxed out issue from the most reagent release. I not 100% sure what is the cause however.

davidusb-geek:
> @davidusb-geek there may be a cpu maxed out issue from the most reagent release. I not 100% sure what the cause is. But it sounds like it's a CBC package issue.

Did you saw this on your own install? Or it was reported to you?
My install is running fine on the latest release, no maxed out CPU issue

PR #458, opened 2025-02-12
fallback to what is set in the config 

scruysberghs:
Attempt to fix the "ERROR - emhass.web_server - The retrieved JSON is empty, A sensor:sensor.power_load_no_var_loads may have 0 days of history, passed sensor may not be correct, or days to retrieve is set too heigh" 

 even if you have another sensor set in your config file.


davidusb-geek (changes requested):
That need to be changed to:
```
var_model = params["retrieve_hass_conf"]["sensor_power_load_no_var_loads"]
```
Because `self` is not defined here.

PR #476, opened 2025-02-27
Adjust PV forecasts

davidusb-geek:
This PR will propose a new method to adjust the current prediction for PV power production forecast.
This can be very useful to adapt the prediction to local conditions.
This should work for either specific conditions with your solar panel such as special shading problems, or even provide a correction based on your specific micro-climate conditions of your environment.
We will use the already available methods using machine learning


davidusb-geek:
What is still needed for PV adjust
	- Need to create a common method to retrieve data from HA in `command_line.py`
		- Before we retrieved only for perfect optim and MPC (for the now functionality)
		- Now we will need to retrieve on the 3 cases: perfect, dayahead and MPC
		- See if the current implementations on perfect and MPC matches
	- Define the new needed parameters >> see `test_forecast.py` >> `test_pv_forecast_adjust`
	- The `adjust_pv_forecast` method in forecast.py still needs a data preparation method, similar to what is done on `test_pv_forecast_adjust`

scruysberghs:
@davidusb-geek A thing that popped during an exeptionally sunny March week here in Belgium last week. If you set your inverter to curtail, the value in your sensor_power_photovoltaics will go down (for instance if you set your inverter to zero grid export when curtailmant has been foreseen by EMHASS). I guess that is something you have to consider when doing ML predictions on the difference between measured and predicted solar yield. 
I noticed I had to change my pv_power_forecast runtime parameter for mpc for this. I started with the actual value of sensor_power_photovoltaics as first value in the list but had to change that to solar+actual consuption (which should normally be the same as actual solar + predicted curtailment) to prevent flip flopping (curtailment on/off)

davidusb-geek:
Added attempt to fix issue #495

Issue #480, opened 2025-03-02
open-meteo negative PV forecast


g1za:
**Describe the bug**
I'm testing the new open-meteo PV forecast. In particular I'm noticing that, if compared to my current method based on forecast.solar, this always results in an infeasible optimization. The one based on forecast.solar (where I manually provide the data base and make sure the values make sense) is always "optimal". Even if the profile of the PV forecast, load, costs and all the rest are exactly the same and also values really really close.
What I noticed and I'm wondering if may be the cause of this behaviour, is that I see some odd values of PV forecast at night when using open-meteo. I see small negative values at night where I actually should see 0 and a tiny PV production for the current hour.
  

**To Reproduce**
Run the open-meteo forecast method. Check for any negative PV forecast values.

**Expected behavior**
No PV production at night, for sure no negative values.

**Screenshots**
Comparing here the charts from open-meteo and forecast.solar:
![Image](https://github.com/user-attachments/assets/440fad5d-a118-428f-b36a-50395b2f8eb2)
![Image](https://github.com/user-attachments/assets/ba5a1650-c56f-487a-a810-d4e0f3d40761)

Here the table with the computation using open-meteo. As you can see there are negative values at night for PV production. I guess this, as part of the passed values, is not affected by the outcome of the optimization but rather used as basis to compute the results.

![Image](https://github.com/user-attachments/assets/e625d1c9-7103-4567-8755-8258e3bde346)
![Image](https://github.com/user-attachments/assets/a326c175-f951-48ac-ba31-46e06558b459)

**Home Assistant installation type**
 - Home Assistant OS

**Your hardware**
- OS: HA OS
- Architecture: amd64

**EMHASS installation type**
 - Add-on

davidusb-geek:
Those negative values are certainly odd, I'll check that. But for me they are not likely to root cause for the infeasible outcome.
For now I cannot explain it.
Would it be possible to provide the payloads that you are passing in for optimal and infeasible cases?

purcell-lab:
With version 0.13 I am no longer seeings negative values from PV forecasts overnight.

g1za:
Correct, it's been fixed

PR #494, opened 2025-03-18
Replace any overnight negative PV power values with zero

paulhomes:
Using emhass with the open-meteo weather forecast, I get small negative forecast PV power values overnight (-75W) which look odd in charts.  This is a small PR to replace any negative PV forecast values with zero.

### DIFF
--- a/src/emhass/forecast.py
+++ b/src/emhass/forecast.py
@@ -537,7 +537,7 @@ class Forecast(object):
         else:
             self.logger.error("Method %r is not valid", method)
             data = None
-        self.logger.debug("get_weather_forecast returning:\n%s", data.to_string())
+        self.logger.debug("get_weather_forecast returning:\n%s", data)
         return data
 
     def cloud_cover_to_irradiance(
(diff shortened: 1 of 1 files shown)


paulhomes:
@davidusb-geek I'm a pandas newbie so please can you check if the negatives-to-zero replacement line I added is ok. I've assumed that the function always returns a single numeric column containing PV values so let me know if that assumption is incorrect. I've not worked with unnamed columns before otherwise I would have only applied the replacement to a named column.

paulhomes:
The logging lines I added are breaking the tests - I'll either remove or fix them.

paulhomes:
I noticed that negative PV values were occurring in issue #480 too.

davidusb-geek (approved):

davidusb-geek:
Thanks for this fix. And yes that function always return a Pandas series and not a Dataframe, so the negative value treatment seems just fine

paulhomes:
Thanks David!

Issue #492, opened 2025-03-17
Feature request: refresh stale weather cache when required rather than error and wait for next iteration


paulhomes:
If I understand this correctly, it looks like when using /action/naive-mpc-optim with weather_forecast_cache: true, and the weather cache is stale, it logs an error/warning and returns HTTP 400. The next call will then refreshes the weather cache and continues as normal.

What do you think of refreshing the weather cache immediately on the first call (assuming weather_forecast_cache: true is passed) instead of logging an error?

In the end it all works out ok as-is, but refreshing automatically when required might help keep the logs clear of errors and avoid subsequent investigations and questions.

This is a docker log showing the initial cache error at 13:15 and subsequent fetch/cache at 13:20 
```
emhass  | [2025-03-17 13:15:30 +1000] [25] [INFO]  >> Obtaining params: 
emhass  | [2025-03-17 13:15:30 +1000] [25] [INFO] Passed runtime parameters: {'load_cost_forecast': [0.06, 0.06, 0.13, 0.15, 0.13, 0.1, 0.28, 0.33, 0.34, 0.45, 0.45, 0.43, 0.37, 0.34, 0.33, 0.31, 0.22, 0.21, 0.23, 0.22, 0.21, 0.2, 0.2, 0.2, 0.2, 0.2, 0.2, 0.2, 0.2, 0.2, 0.2, 0.2, 0.23, 0.23, 0.24, 0.24, 0.23, 0.2, 0.18, 0.18, 0.15, 0.13, 0.13, 0.1, 0.09, 0.07, 0.06, 0.07], 'prod_price_forecast': [-0.03, -0.02, 0.04, 0.06, 0.04, 0.01, 0.04, 0.08, 0.09, 0.19, 0.19, 0.18, 0.12, 0.09, 0.08, 0.07, 0.09, 0.09, 0.1, 0.1, 0.09, 0.08, 0.08, 0.08, 0.08, 0.08, 0.08, 0.08, 0.08, 0.08, 0.08, 0.08, 0.1, 0.11, 0.12, 0.11, 0.1, 0.08, 0.06, 0.06, 0.06, 0.04, 0.04, 0.01, -0.0, -0.02, -0.02, -0.02], 'prediction_horizon': 48, 'soc_init': 0.98, 'soc_final': 0.98, 'weather_forecast_cache': True}
emhass  | [2025-03-17 13:15:30 +1000] [25] [INFO]  >> Setting input data dict
[...]

PR #503, opened 2025-04-04
Cache Open-Meteo JSON reponse locally to reduce the number of API calls and add resilience

paulhomes:
This change caches the default 3-day Open-Meteo JSON response locally with a configurable time limit (30 minutes by default). It is intended to resolve issue https://github.com/davidusb-geek/emhass/issues/492 where the post-subset data stored in the pickle weather_forecast_cache becomes stale quickly and, with regular scheduled 5 min MPCs, it results in frequent Open-Meteo API calls for weather data that does not need to be updated that frequently.  I am currently running this in my test emhass docker instance where the Open-Meteo JSON response is cached using this change and I have turned off the pickle weather_forecast_cache.

It is only caching the JSON response from Open-Meteo so there are no changes to the existing pickle weather_forecast_cache code and it should not impact caching for any of the other weather methods.

By keeping the full 3 days in the JSON cache, and continuing to use it if there are network errors, it also means that emhass will be more resilient to any transient connectivity issues with Open-Meteo.

If you like this suggested change then I can try adding a configuration parameter for max_age for those that might want to change the default (via REST and/or config.json)

What are your preferences for a default? I picked 30 minutes to start with but it could be longer like an hour or maybe 3 or 4 hours.


davidusb-geek:
Hi, there is a pending revision on this PR, can you please check it?

paulhomes:
Hi David, sorry I am not sure I understand what the pending revision is. I did ask a question about if you would like a configuration parameter for max_age (via REST and/or config.json). If so then I will change this PR back to draft and keep working on it.

paulhomes:
Hi David, that's really strange as when I look at this PR, the first message I see from you is the one about the pending revision. I don't see any prior messages other than my initial opening comment and the commits (see screenshot below). Could this be a github bug?

I can see the commented code block I removed though. Sorry about that. I'll add it back in today.

image

davidusb-geek:
Oh I see, that's very strange indeed you cannot see any review comment. Seems like a github bug

paulhomes:
I also added a configuration option for the max age of the cached open-meteo json so it can be set differently to the default of 30 minutes if required.

davidusb-geek (approved):

Issue #519, opened 2025-04-23
0.13.1 breaks my mpc optimiser : [ERROR] The retrieved JSON is empty, A sensor:sensor.p_pv_forecast may have 0 days of history


scruysberghs:
**Describe the bug**
After the changes in 0.13 I had this in my mpc runtime parameters:
'load_forecast_method': 'mlforecaster', 'model_type': 'load_forecast', 'set_use_adjusted_pv': False

That worken fine but after upgrading to 0.13.1 I now get :

[ERROR] The retrieved JSON is empty, A sensor:sensor.p_pv_forecast may have 0 days of history, passed sensor may not be correct, or days to retrieve is set too heigh

Probably because I only publish with the mpc_ prefix. So my ha has a sensor called sensor.mpc_p_pv_forecast but no sensor called sensor.p_pv_forecast

paulhomes:
This sounds similar to https://github.com/davidusb-geek/emhass/issues/517

Is it resolved if you add the following to your config.json?

```
    "sensor_power_photovoltaics_forecast": "sensor.mpc_p_pv_forecast",
```

paulhomes:
@davidusb-geek I think this may be due to the following lines of code in command_line.py:

```
        if optim_conf.get("set_use_pv", True):
            var_list.append(retrieve_hass_conf["sensor_power_photovoltaics"])
            var_list.append(retrieve_hass_conf["sensor_power_photovoltaics_forecast"])
```

For those that are not using the new PV adjustment feature in 0.13 it adds the default config value for `sensor_power_photovoltaics_forecast` (`sensor.p_pv_forecast`) to the list of sensors to fetch from HA and fails if that sensor doesn't exist in HA.  Do you think it would be a good idea to change this to the following?

```
        if optim_conf.get("set_use_pv", True):
            var_list.append(retrieve_hass_conf["sensor_power_photovoltaics"])
        if optim_conf.get("set_use_adjusted_pv", True):
            var_list.append(retrieve_hass_conf["sensor_power_photovoltaics_forecast"])
```

scruysberghs:
adding "sensor_power_photovoltaics_forecast": "sensor.mpc_p_pv_forecast", to the runtime paramters does fix it.

Changing the default in the config would also have fixed it. 

But as you noticed since I had "set_use_adjusted_pv" set to false it wasn't supposed to be requested from HA. 

--
[...]

Issue #517, opened 2025-04-22
Error since last update


gieljnssns:
**Describe the bug**
Since the last update I'm seeing this error in the emhass logs
```
[2025-04-22 09:15:00 +0200] [25] [ERROR] The retrieved JSON is empty, A sensor:sensor.p_pv_forecast may have 0 days of history, passed sensor may not be correct, or days to retrieve is set too heigh
```

**To Reproduce**
Update to latest version

**Expected behavior**
No errors in the log

**Screenshots**
emhass config:
```
{
  "adjusted_pv_regression_model": "LassoRegression",
  "adjusted_pv_solar_elevation_threshold": 10,
  "battery_charge_efficiency": 0.95,
  "battery_charge_power_max": 1000,
  "battery_discharge_efficiency": 0.95,
  "battery_discharge_power_max": 1000,
  "battery_dynamic_max": 0.9,
  "battery_dynamic_min": -0.9,
  "battery_maximum_state_of_charge": 0.9,
  "battery_minimum_state_of_charge": 0.3,
  "battery_nominal_energy_capacity": 5000,
  "battery_target_state_of_charge": 0.6,
  "compute_curtailment": false,
  "continual_publish": false,
  "costfun": "profit",
  "delta_forecast_daily": 1,
  "end_timesteps_of_each_deferrable_load": [
    0,
    0,
    0,
    0,
    0
  ],
  "historic_days_to_retrieve": 360,
  "inverter_is_hybrid": false,
  "load_cost_forecast_method": "hp_hc_periods",
  "load_forecast_method": "mlforecaster",
  "load_negative": false,
  "load_offpeak_hours_cost": 0.164497535,
  "load_peak_hour_periods": {
    "period_hp_1": [
      {
        "start": "08:00"
      },
      {
        "end": "09:00"
      }
    ],
    "period_hp_2": [
      {
        "start": "09:00"
      },
      {
        "end": "10:00"
      }
    ],
    "period_hp_3": [
      {
        "start": "11:00"
      },
      {
        "end": "12:00"
      }
    ],
    "period_hp_4": [
      {
        "start": "13:00"
      },
      {
        "end": "14:00"
      }
    ],
    "period_hp_5": [
      {
        "start": "15:00"
      },
      {
        "end": "16:00"
      }
    ]
  },
  "load_peak_hours_cost": 0.164497535,
  "logging_level": "DEBUG",
  "lp_solver": "COIN_CMD",
[...]

PR #520, opened 2025-04-23
Ignore sensor_power_photovoltaics_forecast when set_use_adjusted_pv is false

paulhomes:
This is a proposed fix for the issue encountered in https://github.com/davidusb-geek/emhass/issues/519 and https://github.com/davidusb-geek/emhass/issues/517 where the sensor default value `sensor.p_pv_forecast` for config option `sensor_power_photovoltaics_forecast` is added to the var_list to be fetched from Home Assistant when `set_use_pv` is true but the new `set_use_adjusted_pv` is false. Is that default sensor doesn't exist in HA then you get the error _The retrieved JSON is empty, A sensor:sensor.p_pv_forecast may have 0 days of history, ..._

This change is to only add `sensor_power_photovoltaics_forecast` to the var_list when both `set_use_pv` and `set_use_adjusted_pv` are true .

## Summary by Sourcery

Bug Fixes:
- Fix an issue where the PV forecast sensor was being fetched even when adjusted PV was disabled, potentially causing errors if the sensor does not exist

### DIFF
--- a/src/emhass/command_line.py
+++ b/src/emhass/command_line.py
@@ -71,8 +71,8 @@ def retrieve_home_assistant_data(
         var_list = [retrieve_hass_conf["sensor_power_load_no_var_loads"]]
         if optim_conf.get("set_use_pv", True):
             var_list.append(retrieve_hass_conf["sensor_power_photovoltaics"])
-        if optim_conf.get("set_use_adjusted_pv", True):
-            var_list.append(retrieve_hass_conf["sensor_power_photovoltaics_forecast"])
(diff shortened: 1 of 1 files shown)


paulhomes:
I was able to replicate the issue and resolve it with this change.  Whilst writing the PR I realized I've added sensor_power_photovoltaics_forecast only when set_use_adjusted_pv is true but it would probably be better to do it when both set_use_pv and set_use_adjusted_pv are true. I'll adjust, retest and commit that change.

davidusb-geek (approved):

PR #530, opened 2025-04-30
HiGHS solver, fix glpk solver, config num_threads

Crosenhain:
- Introduce new `num_threads` option (0 = auto-detect) across:
  - special_options.json, config_defaults.json, associations.csv
  - param_definitions.json (with description & default)
- Wire `num_threads` into the Optimization class:
  - detect CPU count when set to 0
  - debug log the thread count
  - pass threads to CBC and COIN solvers (since they support it)
  - removed threads option from `GLPK_CMD` since it does not support it.
- Add HiGHS as a solver:
  - import `HiGHS` interface from pulp, add `highspy` dependency
  - branch in perform_optimization() to call HiGHS(msg=0, timeLimit)
  - include `HiGHS` in `lp_solver` options everywhere (docs, README, scripts, tests)
- Update Dockerfile to install glpk-utils for GLPK support

## Summary by Sourcery

Enhance optimization solver capabilities by adding HiGHS solver, improving thread configuration, and updating solver support across the project

New Features:
- Add HiGHS solver as a new linear programming solver option
- Introduce configurable `num_threads` option for supported solvers

Enhancements:
- Improve solver configuration flexibility
- Auto-detect CPU thread count when num_threads is set to 0

Deployment:
- Update Dockerfile to install glpk-utils for GLPK support

Documentation:
- Update documentation to include HiGHS solver in solver options
- Add documentation for new `num_threads` configuration parameter

Chores:
- Update project dependencies to include highspy
- Modify configuration files to support new solver and thread options

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -37,7 +37,9 @@ RUN apt-get update \
     libhdf5-serial-dev \
     # cbc
     coinor-cbc \
-    coinor-libcbc-dev
+    coinor-libcbc-dev \
+    # glpk
+    glpk-utils
 
 # add build packadges (just in case wheel does not exist)
 RUN apt-get update \
--- a/README.md
+++ b/README.md
@@ -625,7 +625,7 @@ Pull requests are very much accepted on this project. For development, you can f
 
 Some problems may arise from solver-related issues in the Pulp package. It was found that for arm64 architectures (ie. Raspberry Pi4, 64 bits) the default solver is not available. A workaround is to use another solver. The `glpk` solver is an option.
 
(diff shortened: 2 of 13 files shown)


davidusb-geek:
Hi, thanks for this PR and for contributing to improve the code.
One comment, is the new dependency really necessary? I'm maybe wrong but HIGHS isn't already shipped when installing scipy? (https://docs.scipy.org/doc/scipy/reference/optimize.linprog-highs.html)

Crosenhain:
Good catch, I didn't consider that. 

It seems like they don't expose HiGHS as an importable package or an executable callable from `HiGHS_CMD`, making it only useful if you're calling `scipy.linprog(method='highs')` directly. I would also hesitate to try importing HiGHS from within SciPy as any changes to how SciPy utilises the HiGHS submodule could break HiGHS in EMHASS, whereas PuLP provides us with that supported, stable interface.

PR #541, opened 2025-05-13
Update development documentation to use uv

polyfloyd:
The current documentation refers to `requirements.txt` but there is not such file (anymore?).

pyproject.toml has some references to uv which makes me think it is the preferred package manager.

This patch updates the development documentation with instructions on how to get started using uv.

### DIFF
--- a/docs/develop.md
+++ b/docs/develop.md
@@ -21,10 +21,10 @@ Here you may also wish to add the add the original/upstream repository as a remo
 A command example may be:
 ```bash
 # on GitHub, Fork url, then:
-git clone https://github.com/<YOURUSERNAME>/emhass.git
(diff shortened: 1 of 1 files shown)


davidusb-geek:
Yes uv was adopted recently. However everything can be installed and testes using the bare instructions that were there but that were erased on this PR. So that someone that is not using uv can setup everything the classic way: `python -m venv .venv` etc...
Would it possible the make the two options oc-exist? Thx

polyfloyd:
That should be doable. Exporting the requirements.txt from uv should be sufficient for that

polyfloyd:
Thanks! I have updated the patch. I have omitted the `uv run` in the examples, in favor of activating the uv env itself, but people can decide for themselves how to use uv.

PR #583, opened 2025-09-18
Fix operating_timesteps_of_each_deferrable_load validation formula

scruysberghs:
## Summary
Fix the validation formula for `operating_timesteps_of_each_deferrable_load` parameter that was causing optimization to return "infeasible" results.

## Problem
The runtime parameter `operating_timesteps_of_each_deferrable_load` caused infeasible optimization results in MPC calls, while the equivalent `operating_hours_of_each_deferrable_load` parameter worked correctly.

**Example:**
- ✅ `operating_hours_of_each_deferrable_load: [2.0, 0.0, 0.0, 12.0]` → "Optimal"  
- ❌ `operating_timesteps_of_each_deferrable_load: [8.0, 0.0, 0.0, 48.0]` → "Infeasible"

Both should be equivalent (8 timesteps = 2 hours at 15min intervals).

## Root Cause
Located in `src/emhass/optimization.py` around line 908, the validation formula performed an incorrect double-conversion:

```python
# BUGGY (before):
ceil((60 / ((self.freq.seconds / 60) * def_total_timestep[k])) / self.timeStep)

# CORRECT (after):  
ceil(def_total_timestep[k])
```

The parameter `def_total_timestep[k]` is already in timesteps, so no conversion is needed.

## Solution
Simplified the formula to directly use the timestep value, matching the logic used for the hours parameter:
- Hours: `ceil(def_total_hours[k] / self.timeStep)` 
- Timesteps: `ceil(def_total_timestep[k])`

## Testing
Verified the fix with example values:
- Old buggy formula: 2 timesteps (incorrect)
- New correct formula: 8 timesteps (correct)
- Hours equivalent: 8 timesteps (matches)

## Impact
- ✅ Fixes `operating_timesteps_of_each_deferrable_load` parameter for sub-hour load durations
- ✅ No impact on existing `operating_hours_of_each_deferrable_load` functionality
- ✅ Makes the parameter usable as documented ("better for setting under 1 hr")

🤖 Generated with [Claude Code](https://claude.ai/code)

## Summary by Sourcery

Bug Fixes:
- Correct the timestep validation formula by using ceil(def_total_timestep) instead of a double-conversion expression to ensure valid optimization inputs for sub-hour loads

### DIFF
--- a/src/emhass/optimization.py
+++ b/src/emhass/optimization.py
@@ -904,10 +904,7 @@ class Optimization:
                 def_start, def_end, warning = Optimization.validate_def_timewindow(
                     def_start_timestep[k],
                     def_end_timestep[k],
-                    ceil(
-                        (60 / ((self.freq.seconds / 60) * def_total_timestep[k]))
-                        / self.timeStep
-                    ),
+                    ceil(def_total_timestep[k]),
                     n,
                 )
             else:
--- a/src/emhass/optimization.py
+++ b/src/emhass/optimization.py
@@ -1039,16 +1039,7 @@ class Optimization:
                             f"constraint_pdef{k}_start5": plp.LpConstraint(
                                 e=plp.lpSum(P_def_bin2[k][i] for i in set_I),
                                 sense=plp.LpConstraintEQ,
-                                rhs=(
-                                    (
-                                        60
-                                        / (
-                                            (self.freq.seconds / 60)
-                                            * def_total_timestep[k]
-                                        )
-                                    )
-                                    / self.timeStep
-                                ),
+                                rhs=def_total_timestep[k],
                             )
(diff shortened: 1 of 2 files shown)


scruysberghs:
after testing the initial PR (manually edited the code in the running container) I still got infeasible result. After changing this second occurrence of the same calculation it now seems to work fine on my system

davidusb-geek:
> after testing the initial PR (manually edited the code in the running container) I still got infeasible result. After changing this second occurrence of the same calculation it now seems to work fine on my system

Almost there for the unit tests. The default is 30min. I don't see where it is defined as 15min. And after the optimization we need to check somehow that the timesteps where taken into account, check it numerically

PR #594, opened 2025-10-11
Improvements to timezone handling and testing, use ISO8601 date format, fixes to pass linting checks

Crosenhain:
Removed a bit of the confusing handling of timezones that seemed to have a negative effect on the mlforecaster producing sensible datestamps. Also, switched to using ISO8601 for HA. 

I don't really like how it is yet - I think the better approach is to do everything in UTC and let the UI libraries take care of the conversion to local time. Then you never have to worry about daylight savings times, at least from the emhass perspective.

## Summary by Sourcery

Simplify and standardize timezone and timestamp handling by using UTC internally, return ISO8601-formatted dates, introduce a mockable current-time helper, fix related bugs in forecasting and data retrieval, clean up code formatting, and add tests covering DST boundary cases.

New Features:
- Return forecast dates as ISO8601 strings from get_forecast_dates

Bug Fixes:
- Fix prediction backtest column assignment in machine_learning_forecaster
- Resolve inconsistent timestamp formatting and DST edge-case handling in get_forecast_dates

Enhancements:
- Centralize current time access via helper _get_now for easier mocking
- Enforce ISO8601 formatting for all timestamp outputs and conversions
- Simplify timezone handling by standardizing on UTC and deferring local conversion to UI
- Clean up code formatting and quoting for lint compliance across modules

Build:
- Add pytz to test dependencies in pyproject.toml

Tests:
- Add unit tests for get_forecast_dates under standard and DST-crossing days

### DIFF
--- a/tests/test_retrieve_hass.py
+++ b/tests/test_retrieve_hass.py
@@ -347,7 +347,7 @@ class TestRetrieveHass(unittest.TestCase):
                 "influxdb_host": "fake-host",
                 "influxdb_port": 8086,
                 "influxdb_username": "fake-user",
-                "influxdb_password": "fake-pass",
+                "influxdb_password": None,
                 "influxdb_database": "fake-db",
                 "influxdb_measurement": "W",
                 # Add other necessary keys from the original conf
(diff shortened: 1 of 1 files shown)


Crosenhain:
Yeah let's go with this for now, I'll take a look at using UTC for everything in another PR

PR #599, opened 2025-10-16
Async + websocket data retrieval

gieljnssns:
Main Changes
Asynchronous Refactor

The codebase has been refactored to use asynchronous (async/await) operations throughout, enabling non-blocking I/O and improved performance, especially for network-bound tasks.
Websocket Data Retrieval

Added the option to retrieve data from Home Assistant (HA) using a persistent websocket connection. This allows for real-time data updates without polling, improving efficiency and responsiveness.
Major File Modifications

32 files changed, with over 13,600 lines added and 200+ lines removed.
Significant files affected likely include main application modules, data retrieval clients, configuration, and utility scripts to support async and websockets.
API and Internal Changes

Many function definitions and calls were updated from synchronous to asynchronous.
Integration points with Home Assistant now support both traditional REST polling and websocket-based data streaming.
Likely new helper functions or classes created to manage websocket connections, state handling, and reconnection logic.
Configuration and Settings

New configuration options introduced to enable/disable websocket-based retrieval.
Documentation (README, config examples) may be updated to describe new async and websocket features.
Test and Dependency Updates

Test cases updated or added to cover async workflows and websocket scenarios.
Dependency files (e.g., requirements.txt, pyproject.toml) may reflect new packages for async HTTP/websocket support (such as aiohttp, websockets, etc.).
This PR represents a major architectural shift, aiming to modernize EMHASS for better scalability, efficiency, and integration with Home Assistant’s real-time APIs.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -124,23 +124,11 @@ ENV WORKER_CLASS=uvicorn.workers.UvicornWorker
 ENV PORT=5000
 ENV IP=0.0.0.0
 
-# Entrypoint script inline - async only
+# Entrypoint script inline
 ENTRYPOINT ["/bin/bash", "-c", "set -e && \
 WORKER_CLASS=${WORKER_CLASS:-uvicorn.workers.UvicornWorker} && \
 PORT=${PORT:-5000} && \
 IP=${IP:-0.0.0.0} && \
-echo "=================================================="\
-echo "   Starting EMHASS Energy Management System"\
-echo "=================================================="\
-echo "   Configuration:"\
-echo "   Server Type: $EMHASS_SERVER_TYPE"\
-echo "   Worker Class: $WORKER_CLASS"\
-echo "   Bind Address: $IP:$PORT"\
-echo "   Python Path: $(which python3 2>/dev/null || echo 'N/A')"\
-echo "   UV Version: $(uv --version 2>/dev/null || echo 'N/A')"\
-echo "=================================================="\
-echo 'Using asynchronous Quart server with gunicorn + uvicorn workers' && \
-echo \"Starting gunicorn with $WORKER_CLASS workers...\" && \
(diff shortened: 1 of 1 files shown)


davidusb-geek:
> @scruysberghs What version of influxdb library are you using? v1, v2 or v3? influxdb>=5.3.1 is only sync [influxdb-client-python](https://github.com/influxdata/influxdb-client-python) is for async, but it uses v2

We will need to set the working version for Influxdb, from what I have been using and have seen, the v2 seems like the best compromise.

davidusb-geek:
> @davidusb-geek Have you had a chance to review this pull request yet? Would you prefer to keep the original sync code, or can it all be removed? And maybe you have a proposed test function for the websocket?

This is an extensive refactor. If it's for the best performance wise and it does not affect how the tool is used today then I don't have any problem.
This being said I honestly don't completely understand the core implications of this sync/async flow. Even after reading your original first message on this thread is not completely clear. Is just that I'm not an expert on this and I have not had the time to review the entire changes.
Could you bother to explain what is the advantage of this or the main problem that it solves and how could this possibly change how we use emhass today?

scruysberghs:
>  I see INFO in retrieve_hass_async: InfluxDB integration enabled: localhost:8086/homeassistant in my logs, even though this is set to False

This worries me, as I have seen another report abount influxdb being used even when set to false, but am unable to reproduce myself. Could be caused by use_influxdb: being seen as a string "false" instead of a boolean False but it is handled as any other boolean in the config as far as i can see...
` .get("param_name", False)` then `if param_name`

davidusb-geek:
> Wouldn't it be best to add `“editor.formatOnSave”: true` to .vscode/settings.json

Yes might be a good idea.

> What length should we choose?

The default I guess? whatever that is

gieljnssns:
> The default I guess? whatever that is

The default is 79, I think this is to narrow for this project, there is a lot of nested code. Maybe we should go for 120?

gieljnssns:
@sourcery-ai review

PR #628, opened 2025-11-27
Feature: HA independant configuration

Jahper:
This PR incorporates the feature to allow users to not use HA for configuration data. Like the already existing logs suggested, setting the url and/or token to 'empty' will skip getting configuration data from HA.

## Summary by Sourcery

Allow EMHASS to run without retrieving configuration from Home Assistant when HA connection details are not provided.

New Features:
- Support using only addon configuration by skipping Home Assistant configuration retrieval when HA URL or token are not set.

Bug Fixes:
- Handle the Home Assistant configuration retrieval result explicitly to avoid updating parameters when configuration fetching fails.

### DIFF
--- a/src/emhass/command_line.py
+++ b/src/emhass/command_line.py
@@ -227,6 +227,7 @@ def set_input_data_dict(
     if get_data_from_file:
         with open(emhass_conf["data_path"] / test_df_literal, "rb") as inp:
             _, _, _, rh.ha_config = pickle.load(inp)
+    # Call to get HA configuration
     else:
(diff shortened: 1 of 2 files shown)


davidusb-geek:
@Jahper Thanks for your contribution.
However, the logged message is completely misleading as we will be on standalone mode, not add-one mode.
Also code-wise I don't think that is a good idea to introduce a tri-state. 
I will merge this and then refactor but the idea is similar to what you propose here, so thanks for that.

PR #638, opened 2025-11-30
Add feature to cache pv regression model

sokorn:
added a parameter to cache the adjusted pv regression model. It works by checking the file age and only running a new fit, when the file is older than the given hours (default 24h). This could improve the performance of the day ahead and mpc calls. Especially, when using a high frequency.

## Summary by Sourcery

Add caching and age-based reuse of the adjusted PV regression model to avoid unnecessary re-training while retaining a fallback path when the cached model is invalid.

New Features:
- Introduce an age-based caching mechanism for the adjusted PV regression model, reusing a saved model when it is recent enough and retraining otherwise.

Bug Fixes:
- Ensure adjust_pv_forecast can recover from corrupted or incompatible cached model files by refitting a new model when loading fails.

Enhancements:
- Allow the adjusted PV model maximum age to be configured via optim_conf and overridden at runtime for individual optimization calls.

Documentation:
- Document the adjusted PV model caching behavior, including the adjusted_pv_model_max_age parameter and its runtime override usage in the configuration and forecasts documentation.

Tests:
- Add tests covering the model age check helper, runtime overriding of adjusted_pv_model_max_age, and recovery behavior when a cached adjusted PV model file is corrupted.

### DIFF
--- a/src/emhass/command_line.py
+++ b/src/emhass/command_line.py
@@ -129,7 +129,6 @@ def is_model_outdated(model_path: pathlib.Path, max_age_hours: int, logger: logg
 
 
 async def _retrieve_and_fit_pv_model(
-    logger: logging.Logger,
     fcst: Forecast,
     get_data_from_file: bool,
     retrieve_hass_conf: dict,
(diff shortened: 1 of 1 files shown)


davidusb-geek:
Hi @sokorn, please update your code to solve current conflict so that we can merge this new feature. Thanks

PR #676, opened 2026-01-07
Fix/influxdb skip renaming

sokorn:
When using InfluxDB as data source for ML model training (forecast_model_fit),
the code was incorrectly renaming the load sensor column to add a "_positive"
suffix even when skip_renaming=True was set. This caused a KeyError because
the variable list still referenced the original column name.

Changes:
- Wrap column renaming logic in 'if not skip_renaming:' check
- Only rename and drop columns when skip_renaming is False
- Add debug logging to track skip_renaming parameter and column names
- Improve error message to show actual DataFrame columns for debugging

This fix ensures that when skip_renaming=True (used for ML model training),
the sensor column names remain unchanged and match the variable list.

Fixes model-fit workflows when using InfluxDB data retrieval.

## Summary by Sourcery

Support per-timestep thermal battery temperature bounds and fix load sensor renaming behavior when skip_renaming is enabled.

New Features:
- Allow thermal battery configurations to specify per-timestep min and max temperature bounds, including support for varying limits over time.

Bug Fixes:
- Respect the skip_renaming flag when preparing load data so that sensor column names are not renamed unexpectedly, and improve error logging when load variables are missing.

Enhancements:
- Refine thermal battery physics-based optimization to use temperature lists in constraints, with validation for non-empty bounds and support for partial or None entries in the lists.
- Adjust thermal battery tests to use list-based temperature bounds and add coverage for variable, short, and partially unconstrained temperature lists.

Documentation:
- Update thermal battery documentation and configuration examples to describe list-based min_temperatures/max_temperatures and show how to configure variable comfort ranges over time.

### DIFF
--- a/src/emhass/retrieve_hass.py
+++ b/src/emhass/retrieve_hass.py
@@ -822,26 +822,9 @@ class RetrieveHass:
         self.logger.debug("prepare_data var_interp=%s", var_interp)
         self.logger.debug("prepare_data skip_renaming=%s", skip_renaming)
         self.logger.debug(f"prepare_data df_final columns before rename: {list(self.df_final.columns)}")
-        try:
-            # Only rename column if skip_renaming is False
-            if not skip_renaming:
-                if load_negative:  # Apply the correct sign to load power
-                    self.df_final[var_load + "_positive"] = -self.df_final[var_load]
-                else:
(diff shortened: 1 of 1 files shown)


sokorn:
this is ready to go

davidusb-geek (approved):

PR #679, opened 2026-01-16
Fix missing outdoor temperature data for thermal model

davidusb-geek:
## Summary by Sourcery

Improve robustness of thermal optimization against missing outdoor temperature inputs and align runtime thermal configuration handling.

Bug Fixes:
- Handle missing or incomplete outdoor temperature data in thermal load and thermal battery models by falling back to alternative weather fields or default values.
- Ensure deferrable load runtime configuration correctly updates the number of deferrable loads for optimization.

Enhancements:
- Add logging around thermal load parameter setup to aid debugging of thermal model configuration.
- Relax thermal battery configuration to derive outdoor temperature series from multiple sources with interpolation and patching.

Documentation:
- Update changelog with an unreleased entry describing the thermal model outdoor temperature fix.

### DIFF
--- a/src/emhass/optimization.py
+++ b/src/emhass/optimization.py
@@ -189,6 +189,13 @@ class Optimization:
                         rhs=0,
                     )
 
+    # Helper to extract list from DataFrame/Series/List
+    def _get_clean_list(self, key, data_opt):
+        val = data_opt.get(key)
+        if hasattr(val, "values"):
+            return val.values.tolist()
+        return val if isinstance(val, list) else []
+
     def perform_optimization(
         self,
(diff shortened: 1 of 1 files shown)


purcell-lab:
Looking forward to this as it will fallback to the open metro outside temp forecasts that might already be available/ calculated.

davidusb-geek:
@scruysberghs That is a nice template.
As @purcell-lab says, if the emhass user is already using open-meteo the temperature column is already there. There was a bug in the code where this temperature column was not properly passed to the optimization DataFrame. This PR solves that

PR #688, opened 2026-01-19
Refactor optimization.py to use cvxpy

davidusb-geek:


sokorn:
@davidusb-geek I am testing this PR and ran into user_limit warnings. my 3days 15 min MPC call finished with the old configuration in 47 seconds and the new version runs for 5 minutes.

davidusb-geek:
> @davidusb-geek I am testing this PR and ran into user_limit warnings. my 3days 15 min MPC call finished with the old configuration in 47 seconds and the new version runs for 5 minutes.

That's some bad news. I had tested and compared using a dedicated script in "scripts/benchmark_performance.py" and consistently obtained better faster results with the new code.

Can you help debug this?
Can we search from where that user_limit warning is coming from?
Any chance that it is related to your PV forecasting method? What is your current config?

PR #690, opened 2026-01-21
Missing unit convertion for load_power

sokorn:
@davidusb-geek I am really sorry, just realized that I forgot to do a unit convertion from W to kW in the inter_gains calculation. Just a small PR :)

## Summary by Sourcery

Correct internal gains unit handling by treating electrical load forecasts as Watts and converting to kW in the heating demand physics calculation.

Bug Fixes:
- Fix internal gains calculation by converting electrical load power from Watts to kilowatts before subtracting from heating demand.

Documentation:
- Clarify the calculate_heating_demand_physics parameter documentation to state that internal gains forecasts are expected in Watts.

Tests:
- Update heating demand physics tests to provide electrical load forecasts in Watts instead of kilowatts and maintain consistency with the new unit convention.

### DIFF
--- a/src/emhass/utils.py
+++ b/src/emhass/utils.py
@@ -23,6 +23,9 @@ if TYPE_CHECKING:
 
 pd.options.plotting.backend = "plotly"
 
+# Unit conversion constants
+W_TO_KW = 1000  # Watts to kilowatts conversion factor
+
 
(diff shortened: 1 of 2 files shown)


davidusb-geek:
This is way easier as it only involves utils.py

PR #693, opened 2026-01-21
Solve secrets handling and MILP big-M solve hang issues

davidusb-geek:
## Summary by Sourcery

Improve deferrable load MILP formulation for better solver performance and extend and clarify secrets handling for Solcast and InfluxDB across code and docs.

New Features:
- Support reading Solcast and InfluxDB credentials from Home Assistant config and options.json into the EMHASS secrets map.
- Document Solcast and InfluxDB secret parameters and how to provide them via the add-on configuration or secrets_emhass.yaml.

Bug Fixes:
- Prevent MILP solver hangs by replacing a large constant Big-M with a load-specific dynamic Big-M for deferrable load constraints.
- Avoid conflicting constraints on sequence-based deferrable loads by skipping single-constant and semi-continuous logic for these loads.
- Ensure InfluxDB username and password are consistently treated as secrets instead of regular configuration parameters.

Enhancements:
- Refine deferrable load modeling by separating sequence-based loads from standard loads and tightening related constraints.
- Expand default secrets template to include InfluxDB and Solcast-related fields for easier configuration.

Documentation:
- Update configuration and data passing documentation to describe all secret parameters, including Solcast and InfluxDB credentials, and how to supply them via different installation methods.

### DIFF
--- a/tests/test_optimization.py
+++ b/tests/test_optimization.py
@@ -2339,9 +2339,9 @@ class TestOptimization(unittest.IsolatedAsyncioTestCase):
             True,
             True,
             True,
-            False,
             True,
-            False,
+            True,
+            True,
         ]
         complex_optim_conf["set_deferrable_load_single_constant"] = [
             True,
@@ -2399,7 +2399,7 @@ class TestOptimization(unittest.IsolatedAsyncioTestCase):
         )
 
         # Runtime Definitions
-        def_total_timestep = [24, 8, 0, 0, 0, 0, 0]
+        def_total_timestep = [24, 8, 12, 5, 9, 27, 18]
         def_start_timestep = [0] * 7
         def_end_timestep = [n_steps] * 7
 
@@ -2416,13 +2416,30 @@ class TestOptimization(unittest.IsolatedAsyncioTestCase):
             )
 
             # Assertions
-            self.assertEqual(opt_res["optim_status"].iloc[0], "Optimal")
+            self.assertEqual(opt_res["optim_status"].iloc[0], "Optimal (Relaxed)")
 
             # Check Load 0
             p_def_0 = opt_res["P_deferrable0"]
-            steps_active = (p_def_0 > 1.0).sum()
-            self.assertEqual(steps_active, 24, "Load 0 should be active for exactly 24 timesteps")
 
+            # Since we might have fallen back to relaxed LP, strict timestep counting
+            # can be off by +/- 1 due to continuous "smearing" of energy.
+            # Instead, verify the Total Energy matches the requirement.
+            # Target: 24 steps * 3000 W
+            total_energy_delivered = p_def_0.sum()
+            target_energy = 24 * 3000
+
+            # Allow small numerical tolerance (e.g. 1%)
+            self.assertAlmostEqual(
+                total_energy_delivered, target_energy, delta=target_energy * 0.01
+            )
+
+            # Optional: Relaxed check for steps (allow 24 or 25)
+            steps_active = (p_def_0 > 10.0).sum()  # Use higher threshold (10W) to ignore noise
+            self.assertTrue(
+                23 <= steps_active <= 25, f"Expected ~24 active steps, got {steps_active}"
+            )
+
+            # Check Max Power matches nominal (approx)
             max_power = p_def_0.max()
             self.assertAlmostEqual(max_power, 3000.0, delta=1.0)
 
--- a/src/emhass/optimization.py
+++ b/src/emhass/optimization.py
@@ -1404,12 +1404,14 @@ class Optimization:
         def_start_timestep: list | None = None,
         def_end_timestep: list | None = None,
         def_init_temp: list | None = None,
+        min_power_of_deferrable_loads: list | None = None,
         debug: bool | None = False,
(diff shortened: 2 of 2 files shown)


sokorn:
> @sokorn I've added a special test in test_optimization.py: test_perform_naive_mpc_optim_complex_case This tries to replicate your configuration. It solves very quicky, < 1second?

just tested this branch, but it still took nearly 6 minutes this time. Could this be related to the thermal_config and my long prediction_horizon? I know that a few others also use nearly 3 days of horizon.
Additionally I got the following log line:

`HighsMipSolverData::transformNewIntegerFeasibleSolution tmpSolver.run();`

tests/test_optimization.py
```diff
@@ -2280,6 +2281,156 @@ def test_power_limit_columns_are_integers(self):
...
+                "ventilation_rate": 0.41,
+                "heated_volume": 356.0,
+                "indoor_target_temp": 21,
+                "window_area": 29.0,
+                "shgc": 0.50,
+                "start_temperature": 20,
+                "min_temperatures": [18] * n_steps,
+                "max_temperatures": [24] * n_steps,
+                "supply_temperature": 35.0,
+                "cop_carnot_efficiency": 0.4,
```
sokorn: have you renamed this parameter? shouldn't it be named `carnot_efficiency`

sokorn:
did some more testing:
- **no** thermal_config and **no** thermal_battery: 1 second
- thermal_config and **no** thermal_battery: 33 seconds
- **no** thermal_config and thermal_battery: 33 seconds
- thermal_config **and** thermal_battery: up to 6 minutes
- two thermal_batteries: up to 6 minutes
- two thermal_configs: up tp 6 minutes

The runtime issue occurs if more than one thermal load is used.

tests/test_optimization.py
davidusb-geek: No I did not renamed anything, that's just an error on that test

davidusb-geek:
> did some more testing:
> 
> * **no** thermal_config and **no** thermal_battery: 1 second
> * thermal_config and **no** thermal_battery: 33 seconds
> * **no** thermal_config and thermal_battery: 33 seconds
> * thermal_config **and** thermal_battery: up to 6 minutes
> * two thermal_batteries: up to 6 minutes
> * two thermal_configs: up tp 6 minutes
> 
> The runtime issue occurs if more than one thermal load is used.

There must be something else going on, this does not seems related to the new optimization arch. 
As you can see my includes thermal_config **and** thermal_battery. And it runs very fast. That particular test runs in less than a second. Actually in my dev machine I ran the whole test_optimization.py that's 40 tests in 14.6s.
So like I said there must be something else blocking and slowing the optimization.
It could be a RAM issue?
Also the benchmark has proven that the new code is more efficient and fast on all tests.

davidusb-geek:
> The new commit only works with adjustment on my configuration. I think the issue is, that my thermal_config for DHW had startup_penalty and is semi-continuous. When I changed to semi_cont=false the problem solves in 4 seconds. This config solves really fast (load 4 and 6 are DHW and thermal_battery): `"num_def_loads":7,"def_current_state":[0,0,0,0,0,0,1],"operating_timesteps_of_each_deferrable_load":[24,8,12,5,9,27,18],"start_timesteps_of_each_deferrable_load":[139,174,0,0,0,0,0],"end_timesteps_of_each_deferrable_load":[174,186,41,41,0,0,0],"treat_deferrable_load_as_semi_cont":[true,true,true,true,false,true,false],"set_deferrable_load_single_constant":[true,true,true,false,false,true,false],"def_start_penalty":[1,1,1,1,0,1,0],"nominal_power_of_deferrable_loads":[600,500,450,10500,2000,250,899],"minimum_power_of_deferrable_loads":[0,0,0,0,0,0,580]`
> 
> and this config runs into a 360 second timeout: `"num_def_loads":7,"def_current_state":[0,0,0,0,0,0,1],"operating_timesteps_of_each_deferrable_load":[24,8,12,5,9,27,18],"start_timesteps_of_each_deferrable_load":[139,174,0,0,0,0,0],"end_timesteps_of_each_deferrable_load":[174,186,41,41,0,0,0],"treat_deferrable_load_as_semi_cont":[true,true,true,true,true,true,false],"set_deferrable_load_single_constant":[true,true,true,false,false,true,false],"def_start_penalty":[1,1,1,1,0,1,0],"nominal_power_of_deferrable_loads":[600,500,450,10500,2000,250,899],"minimum_power_of_deferrable_loads":[0,0,0,0,0,0,580]`

Thanks for that detailed feedback. i'm now able to properly and consistently replicate this issue. You can check again the "test_perform_naive_mpc_optim_complex_case" test method.
I have proposed now a fix for this based on relaxation by drooping the binary constraints, so the problem becomes purely LP and solves very quickly. This is only activated as a fallback if an initial solve fails with "user_limit" or "infeasible".
The previous pulp-based code might have applied a similar solution also based on some advanced meta-heuristics. In any case this is the limit of the MILP problem: the curse of dimensionality. This special case results in a huge problem to solve. For better results either reduced the number of steps or the number of deferrable loads.
I will merge this and treat this as solved for now.

PR #711, opened 2026-02-02
Added caching for the Optimization object to use warm start

sokorn:
## Summary by Sourcery

Introduce a warm-start capable optimization cache and parameterization of deferrable load constraints to allow reusing Optimization instances across MPC runs while remaining robust to configuration changes.

Enhancements:
- Add an in-memory OptimizationCache with deterministic cache keys to reuse Optimization objects across runs when structure-affecting configuration is unchanged.
- Parameterize deferrable load window and energy constraints (including Big-M formulations and masks) so operating hours and time windows can change without rebuilding the optimization problem.
- Extend deferrable load constraint handling to recognize thermal and thermal-battery configurations and exclude them from generic energy constraints when appropriate.

Tests:
- Add a comprehensive OptimizationCache test suite covering cache hits/misses for various configuration changes, key determinism, stats reporting, timestamp updates, and handling of None and def_load_config variations.

### DIFF
--- a/src/emhass/command_line.py
+++ b/src/emhass/command_line.py
@@ -93,6 +93,32 @@ class PublishContext:
         return self.input_data_dict["fcst"]
 
 
+@dataclass(frozen=True)
+class OptimizationCacheKey:
+    """
+    Frozen dataclass representing configuration fields that affect optimization structure.
+
+    Changes to any of these fields require rebuilding the optimization problem.
+    Using a frozen dataclass makes the cache key explicit, hashable, and easy to extend.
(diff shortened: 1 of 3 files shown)


sokorn:
I have to look into the parametrization once again, seeing some odd behaviour of the heating demand calculation since i added it. I will leave a comment here, if I have finished it. Same with the other PR.

sokorn:
> @sokorn is this good to merge or you intent to provide further improvements?

the two PR now look good. You can merge them :)

PR #715, opened 2026-02-06
Pin cvxpy version to <1.8.0

rmounce:
Pin cvxpy to <1.8.0 to avoid build issues with version 1.8.1 on CPython 3.12.


davidusb-geek:
Hi @rmounce, where did you saw those issues with cpython ?

PR #739, opened 2026-03-05
fix: Solcast Accept header and sub-30min resampling

kallegrens:
**DISCLAIMER**:
This is a Claude-generated PR, the code was however tested live on my emhass instance and seemed to work. Feel free to close if this is against policy.

---

## Summary

Two bugs in `_get_weather_solcast()` prevent Solcast forecasts from working correctly:

### Bug 1: Wrong HTTP header
The request uses `"content-type": "application/json"` instead of `"Accept": "application/json"`. The `content-type` header describes the *request body* (which is empty for a GET), while `Accept` tells the server what format the client wants. Without the correct `Accept` header, Solcast returns HTML instead of JSON, causing a parse error.

### Bug 2: No resampling for sub-30min timesteps
Solcast returns data at 30-minute intervals. When `optimization_time_step < 30` (e.g. 15 min), the code compared `len(data_list)` against `len(self.forecast_dates)` and always found "not enough data" because 48 Solcast points < 96 forecast slots.

The fix builds a timestamped DataFrame from Solcast's `period_end` timestamps and resamples using `reindex() + interpolate(method='time')`, matching the pattern already used in `_get_weather_solar_forecast()`.

### Also
- Removes the now-unused `zip_longest` import

### Testing

**Unit tests** (pytest, local Python 3.12):
- All 4 Solcast tests pass (3 existing + 1 new)
- New: `test_get_weather_forecast_solcast_15min_resampling_mock`

**Production pod validation** (v0.17.0 on Kubernetes):
Patched `forecast.py` was `kubectl cp`'d into the running pod and validated:

1. **30-min timestep** (default): 48 slots, peak 9000 W — direct mapping ✅
2. **15-min timestep** (production): 96 slots, peak 9000 W — interpolated correctly ✅
   - e.g. between 08:00 (100 W) and 08:30 (500 W), 08:15 → 300 W (linear midpoint)
3. **Failure modes** — all return `False` gracefully:
   - 429 rate limit ✅
   - 403 bad API key ✅
   - Empty forecasts ✅
   - Missing API key ✅

Live Solcast API call also confirmed working (got 200 + valid JSON before hitting free-tier daily quota).

## Summary by Sourcery

Fix Solcast weather forecast retrieval and resampling to support sub-30-minute optimization time steps and ensure valid JSON responses.

Bug Fixes:
- Correct Solcast HTTP request header to use the Accept header so the API returns JSON instead of HTML.
- Handle empty Solcast forecast responses and return a graceful failure instead of proceeding with invalid data.
- Fix Solcast data length mismatch for sub-30-minute optimization time steps by resampling provider data to match forecast dates.

Enhancements:
- Aggregate multi-rooftop Solcast data using a timestamped DataFrame with time-based interpolation and zero-filled edges instead of manual list accumulation.

Tests:
- Add a Solcast unit test verifying 30-minute input data is correctly resampled to a 15-minute optimization time step and contains no NaN values.

### DIFF
--- a/tests/test_forecast.py
+++ b/tests/test_forecast.py
@@ -476,6 +476,27 @@ class TestForecast(unittest.IsolatedAsyncioTestCase):
             # Verify no NaN values after interpolation
             self.assertFalse(df_weather_scrap["yhat"].isna().any())
 
+            # Verify interpolation correctness at a midpoint between two 30-min source timestamps
+            # Pick a midpoint index to avoid edge effects
+            midpoint_idx = len(df_weather_scrap.index) // 2
+            ts_mid = df_weather_scrap.index[midpoint_idx]
+            ts_prev = ts_mid - pd.Timedelta(minutes=15)
+            ts_next = ts_mid + pd.Timedelta(minutes=15)
+
+            # Ensure the neighboring timestamps exist in the index
+            self.assertIn(ts_prev, df_weather_scrap.index)
+            self.assertIn(ts_next, df_weather_scrap.index)
+
+            y_prev = df_weather_scrap.loc[ts_prev, "yhat"]
+            y_mid = df_weather_scrap.loc[ts_mid, "yhat"]
+            y_next = df_weather_scrap.loc[ts_next, "yhat"]
+
+            # Expected linear interpolation at the midpoint
+            expected_mid = (y_prev + y_next) / 2.0
+
+            # Check that the interpolated midpoint matches the expected linear value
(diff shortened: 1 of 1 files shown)


davidusb-geek:
> This is a Claude-generated PR, the code was however tested live on my emhass instance and seemed to work. Feel free to close if this is against policy.

Hi thanks for your contribution and no problem accepting PR created using AI. As long as they are well constructed and the code is tested. The best way is to propose unit tests to directly test the functionalities of your new code

PR #745, opened 2026-03-10
fix: prevent Solcast rate limit exhaustion with daily call cap and stale cache reuse

kallegrens:
## Problem

When the Solcast weather forecast cache expires or is invalidated, every MPC optimization call (every ~4 minutes) falls through to a live Solcast API call. The free tier allows only 10 API calls/day, so the quota is burned within ~40 minutes. After that, every call fails and the EMHASS web UI shows stale data for the rest of the day.

Two interacting issues:
1. `get_cached_forecast_data()` deletes the cache file when the requested timeframe doesn't fully overlap, causing subsequent calls to hit Solcast directly
2. No rate limiting on live Solcast API calls

### 1. Daily Solcast API call rate limiter
- Added `_solcast_rate_limit_ok()` static method with a file-based daily counter (`/tmp/solcast_calls_<date>.count`)
- Caps live API calls to 8/day (free tier allows 10, leaving 2 as buffer)
- Inserted before any live Solcast API call in `_get_weather_solcast()`

### 2. Stale cache reuse instead of deletion
- `get_cached_forecast_data()`: when the cache doesn't fully cover the requested timeframe, instead of deleting it and returning False, now serves best-effort data by reindexing to the requested dates, interpolating gaps, and zero-filling edges
- Cache is only deleted when truly corrupt (not a DataFrame or empty)
- This prevents the cascade: stale cache deleted → every call hits API → quota exhausted → all calls fail

## Testing

Deployed and validated on a live system:
- MPC calls with `weather_forecast_cache_only: true` correctly use cache or fall back to ML-based forecast without hitting Solcast
- Rate limiter counter correctly tracks API calls
- Stale cache reuse provides reasonable PV forecasts when cache partially covers the window

## Summary by Sourcery

Add protections against exhausting Solcast API quota by limiting daily live calls and reusing stale cached forecasts instead of deleting them.

Bug Fixes:
- Prevent repeated Solcast API calls when the cache timeframe does not fully match the requested window by serving best-effort data instead of deleting the cache.
- Avoid exhausting the Solcast daily API quota by adding a file-based daily call cap before performing live Solcast requests.

Enhancements:
- Improve cache corruption detection and logging guidance for refreshing forecast cache data.

### DIFF
--- a/src/emhass/command_line.py
+++ b/src/emhass/command_line.py
@@ -18,12 +18,12 @@ import orjson
 import pandas as pd
 
 from emhass import utils
-from emhass.utils import log_runtime_banner, stage_timer
 from emhass.forecast import Forecast
 from emhass.machine_learning_forecaster import MLForecaster
 from emhass.machine_learning_regressor import MLRegressor
 from emhass.optimization import Optimization
 from emhass.retrieve_hass import RetrieveHass
+from emhass.utils import log_runtime_banner, stage_timer
 
 default_csv_filename = "opt_res_latest.csv"
 default_pkl_suffix = "_mlf.pkl"
--- a/src/emhass/command_line.py
+++ b/src/emhass/command_line.py
@@ -569,7 +569,13 @@ async def _retrieve_and_fit_pv_model(
         emhass_conf,
         test_df_literal,
     )
-    if not success:
+    if not success or df_input_data is None or df_input_data.empty:
+        return False
+    if fcst.var_pv_forecast not in df_input_data.columns:
+        fcst.logger.warning(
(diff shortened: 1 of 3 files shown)


kallegrens:
@davidusb-geek Hey! I addressed the Sourcery comments now. It marked all as resolved.

davidusb-geek:
> @davidusb-geek Hey! I addressed the Sourcery comments now. It marked all as resolved.

Great @kallegrens, unfortunately as you can see some tests are failing. 
We need to fix those before merging.
You might need to rebase this branch to update with the current master

davidusb-geek (approved):

Issue #585, opened 2025-09-25
[Feature Request] Use Actual SOC as Starting Point in Optimization


martinarva:
**Feature Request: Use Actual SOC as Starting Point in Optimization**
------------------------------------------------------------------------

### **Problem**

Currently, EMHASS uses the configured battery\_minimum\_state\_of\_charge and battery\_maximum\_state\_of\_chargevalues both as **bounds** and as **hard limits for the initial SOC** at the beginning of optimization.

This becomes problematic when the actual SOC is **outside** these limits — EMHASS assumes it’s at battery\_minimum\_state\_of\_charge or battery\_maximum\_state\_of\_charge, rather than using the **true SOC** from the battery.

### **Why this matters**

In setups where EMHASS isn’t the only controller for the battery, this behavior results in **incorrect forecasts** and potentially **suboptimal scheduling decisions**.

### **Example use case**

I use EMHASS to optimize my battery within the range of **30%–80% SOC**. However, I’m also participating in the **MFFR (Manual Frequency Restoration Reserve)** market via **Qilowatt/Kratt**. These signals override EMHASS control:

*   **MFFR UP** events may discharge the battery below 30% (e.g., down to 5%)
    
*   **MFFR DOWN** events may charge the battery up to 100%
    

This is intentional — I want to **reserve 0–30% for UP** and **80–100% for DOWN** commands.

But when this happens, EMHASS ignores the actual SOC and **treats it as 30% or 80%**, leading to an **incorrect optimization baseline**.

### **Visual Example**

See attached chart:
<img width="408" height="370" alt="Image" src="https://github.com/user-attachments/assets/4eb20f19-ab17-4a9d-a288-2c3dd91d003c" />

*   **Green**: actual SOC
    
*   **Red**: forecast history
    
*   **Blue**: forecast going forward
    

You can clearly see that EMHASS assumes the SOC is at 30% when in reality it’s closer to 5%.

### **Desired Behavior**

*   EMHASS should always **use the real actual SOC** from the sensor as the starting point for optimization — **even if it’s outside the configured min/max SOC bounds**.
    
*   The battery\_minimum\_state\_of\_charge and battery\_maximum\_state\_of\_charge should still be enforced as **constraints on the future steps**, but **not** as assumptions about the **starting point**.
    

### **Why this change is important**

This improvement benefits all users who:

*   Manually charge or discharge the battery (e.g. for testing or off-grid prep)
    
[...]

PR #761, opened 2026-03-20
fix: support out-of-band initial SOC recovery in naive MPC

martinarva:
## Summary
- preserve the real `soc_init` even when it starts outside the configured SOC limits
- allow temporary out-of-band initial SOC in naive MPC and enforce recovery back into the valid SOC band
- keep recovery monotonic before first re-entry and prevent going out of bounds again within the same optimization horizon
- add targeted regression tests for low/high SOC recovery and runtime parameter handling

## Motivation
This addresses the feature request and discussion here:
- https://github.com/davidusb-geek/emhass/issues/585

The main goal is to let EMHASS work with the real battery SOC even when the current SOC is temporarily below `battery_minimum_state_of_charge` or above `battery_maximum_state_of_charge`, instead of clipping `soc_init` before optimization.

## What Changed
- stop clamping `soc_init` to `soc_min/soc_max` in runtime parameter handling
- add recovery-aware SOC constraints in the optimizer
- require SOC to move only toward the valid band until first re-entry
- once SOC is back inside the configured band, enforce the normal hard SOC bounds again
- add regression tests for:
  - preserving out-of-band `soc_init`
  - low-SOC recovery
  - high-SOC recovery
  - monotonic recovery before re-entry
  - non-ideal battery efficiency case

## Testing
- ran targeted SOC recovery tests locally
- tested against a real Home Assistant setup with real runtime payloads
- verified both:
  - `soc_init > soc_max`
  - `soc_init < soc_min`

## Summary by Sourcery

Support temporary out-of-band initial battery SOC in naive MPC while enforcing controlled recovery back into configured SOC limits.

Bug Fixes:
- Preserve out-of-band initial SOC from runtime parameters instead of clamping it to configured minimum/maximum limits.

Enhancements:
- Add recovery-aware SOC constraints and state tracking to ensure monotonic SOC movement back into the valid band and prevent further violations within the same optimization horizon.

Tests:
- Add regression and integration-style tests covering low/high SOC recovery, monotonic behavior with and without non-ideal efficiencies, and preservation of out-of-band soc_init in runtime parameter handling.

### DIFF
--- a/src/emhass/optimization.py
+++ b/src/emhass/optimization.py
@@ -1103,8 +1103,9 @@ class Optimization:
         soc_high_recovered = self.vars["soc_high_recovered"]
         min_energy = self.plant_conf["battery_minimum_state_of_charge"] * cap
         max_energy = self.plant_conf["battery_maximum_state_of_charge"] * cap
-        recovery_big_m = cap
         recovery_margin = max(cap * 1e-6, 1e-3)
+        recovery_big_m_low = cap - min_energy + recovery_margin
+        recovery_big_m_high = max_energy + recovery_margin
 
         # Grid Interaction Constraints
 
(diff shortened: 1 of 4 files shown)


davidusb-geek:
@martinarva 
The test suite catched an issue here.

It seems that the relaxation terms designed to "disable" the recovery logic during normal operations (when the battery is within limits) are slightly too tight, causing an Infeasible case there in one of the tests.

A proposed solution is to update two specific constraints in the_init_deferrable_load_params method to this:
```
constraints.append(
            current_stored_energy
            <= min_energy
            - recovery_margin
            + recovery_big_m * soc_low_recovered
            + (recovery_big_m + recovery_margin) * (1 - self.param_soc_low_required)
        )
        constraints.append(
            current_stored_energy
            >= max_energy
            + recovery_margin
            - recovery_big_m * soc_high_recovered
            - (recovery_big_m + recovery_margin) * (1 - self.param_soc_high_required)
        )
```

martinarva:
Thanks, that helped pinpoint the issue. I tested that direction locally and ended up applying a closely related fix on this branch. Instead of using a single `recovery_big_m = cap`, I switched to side-specific bounds for the recovery gating terms:

- `recovery_big_m_low = cap - min_energy + recovery_margin`
- `recovery_big_m_high = max_energy + recovery_margin`

That keeps the out-of-band recovery behavior intact, but fully relaxes the corresponding low/high gating constraints in normal in-band operation and in the previously failing hybrid inverter case.

I tested locally with:
- `tests/test_soc_recovery_prototype.py` -> 5 passed
- `tests/test_utils.py -k 'preserves_out_of_band_soc_init or preserves_high_out_of_band_soc_init'` -> 2 passed
- a local reproduction of the previously failing `test_inverter_stress_cost_discharge_spread` setup -> `Optimal` for both `stress=0.0` and `stress=1.0`

The updated commit is now on this PR branch, so CI should show whether this resolves the regression cleanly across the full suite.

PR #766, opened 2026-03-24
Fix DST-boundary crashes in forecast date range computation

carposio:
For a 7-day forecast, I started to get issues on Sunday saying that array sizes didn't match in `_get_weather_list`: 668 vs. 672. Then I realized that DST is changing. With the help of claude, I came up with the following fix which seems to work fine for me now

`Forecast.__init__` was adding `delta_forecast_daily` (a `pd.Timedelta`) to `start_forecast`, giving a fixed-duration end time that does not account for DST. `utils.get_forecast_dates` uses `pd.DateOffset` (calendar days), so the two independently computed date ranges diverged by one hour (4 × 15-min slots) whenever the forecast window crossed a DST transition. This caused downstream length mismatches and crashes in `_get_weather_list`, `_load_forecast_data`, and `_apply_df_freq_horizon`.

- Replace `Timedelta` additions with `pd.DateOffset` in all three `end_forecast` computations inside Forecast (including the ×2 cache variant and the CSV/cache-trim helpers), making them DST-aware and consistent with `get_forecast_dates`.
- Clamp the `prediction_horizon` slice in `_apply_df_freq_horizon` to the actual `df` length to guard against any remaining off-by-one at DST boundaries.
- Parse `opt_res_latest.index` with `utc=True` then `tz_convert`, so mixed UTC-offset strings produced by a DST-spanning result CSV are handled correctly.
- Normalise `_get_weather_list` to resolve `self.forecast_dates` to the configured `time_zone` before the length check and index assignment.

## Summary by Sourcery

Make forecast date range handling robust across DST transitions and align it with other time-based processing.

Bug Fixes:
- Ensure forecast end times are computed using calendar-day offsets so weather cache and CSV ranges stay consistent across DST changes.
- Prevent crashes in weather list generation by timezone-normalizing forecast dates before length checks and index assignment.
- Avoid index out-of-bounds when applying a prediction horizon by clamping the slice to the available dataframe length.
- Correct parsing of optimization result timestamps from CSV by treating them as UTC and converting to the configured time zone to handle DST-spanning data.

Enhancements:
- Unify forecast end-time calculations in core, CSV, and cache-trimming paths using a shared calendar-day offset configuration.

### DIFF
--- a/src/emhass/command_line.py
+++ b/src/emhass/command_line.py
@@ -726,7 +726,7 @@ def _apply_df_freq_horizon(
     # Handle Prediction Horizon
     if prediction_horizon:
         # Slice the dataframe up to the horizon
-        df = copy.deepcopy(df)[df.index[0] : df.index[prediction_horizon - 1]]
+        df = copy.deepcopy(df)[df.index[0] : df.index[min(prediction_horizon, len(df)) - 1]]
     return df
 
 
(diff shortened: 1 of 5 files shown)


scruysberghs:
Some input for the coding agent : emhass could do with DST related test(s) as wel. 
Have the feeling this was fixed in the past and regressed somehow since previous DST switch

davidusb-geek:
@carposio Thanks for the contribution.
Please fix the failing tests to be able to merge

carposio:
I added another fix and test cases just now, but I think this might need a bit more testing in production

davidusb-geek (approved):

PR #775, opened 2026-04-02
fix: add state_class to published sensors and prevent ML crash on short history

kallegrens:
This PR introduces two fixes:

1. **HA Statistics tracked correctly:** Adds `state_class` to the sensors published back to Home Assistant. Without this, Home Assistant refuses to store the data in its long-term statistics database, which EMHASS relies on for fetching historical data via WebSockets.
2. **Graceful ML fallback:** If the historical PV forecast data retrieved is too short to fulfill the required cross-validation folds (e.g. `< 6` samples when history is just enabled/cleared), it gracefully falls back to the unadjusted PV forecast rather than crashing scikit-learn with a `ValueError`.

## Summary by Sourcery

Ensure Home Assistant sensors publish long-term statistics-compatible attributes and make PV forecast adjustment more robust when historical data is insufficient.

Bug Fixes:
- Add appropriate state_class to published sensor attributes so Home Assistant stores long-term statistics used by EMHASS.
- Prevent PV model training from failing when historical data is too short by falling back to the unadjusted PV forecast instead of raising errors.

Enhancements:
- Centralize construction of sensor attributes, including device_class, units, friendly name, and forecast payload, for different sensor types.

### DIFF
--- a/src/emhass/command_line.py
+++ b/src/emhass/command_line.py
@@ -620,7 +620,8 @@ async def adjust_pv_forecast(
             test_df_literal,
         )
         if not success:
-            return False
+            logger.warning("Could not train adjusted PV model, falling back to unadjusted PV forecast.")
+            return p_pv_forecast
     else:
         # Load existing model
         logger.info("Loading existing adjusted PV model from file")
@@ -644,15 +645,15 @@ async def adjust_pv_forecast(
                 test_df_literal,
             )
             if not success:
-                logger.error("Failed to retrieve data for model re-fit after load error")
-                return False
+                logger.error("Failed to retrieve data for model re-fit after load error. Falling back to unadjusted forecast.")
+                return p_pv_forecast
             logger.info("Successfully re-fitted model after load failure")
         except Exception as e:
             logger.error(
                 f"Unexpected error loading adjusted PV model: {type(e).__name__}: {str(e)}"
             )
-            logger.error("Cannot recover from this error")
-            return False
+            logger.error("Cannot recover from this error. Falling back to unadjusted forecast.")
+            return p_pv_forecast
     # Call the predict method
     p_pv_forecast = p_pv_forecast.rename("forecast").to_frame()
     p_pv_forecast = fcst.adjust_pv_forecast_predict(forecasted_pv=p_pv_forecast)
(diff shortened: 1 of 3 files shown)


davidusb-geek:
Thanks for these improvements.
Could you fix the failing tests to be able to merge?

davidusb-geek:
Hi @kallegrens, you need to rebase this branch and solve the conflicts so that I can merge your contribution

kallegrens:
I'm fairly convinced that this single test failure does not originate from the code changes coming from this PR, but are pre-existing. For example, the exact same failure is also in this open PR: https://github.com/davidusb-geek/emhass/actions/runs/25889088781/job/76757375043?pr=851

kallegrens:
I fixed that test anyway, although it wasn't related to this PR. AI summary follows below:

---

I tracked the CI failure down to solver residue in the optimization result export, not to an actual scheduling regression in this PR.

What was happening:
- The deferrable-load constraints were producing the correct binary schedule.
- In some runs, CVXPY/HiGHS returned a tiny positive residue in the exported power series (for example `1.4e-13` instead of `0.0`).
- The tests count active timesteps with a strict `> 0` check, so that residue made one extra timestep look active and caused the failure.

What I changed:
- In `Optimization._build_results_dataframe`, I now normalize values within `1e-9` of zero to `0.0` before building the output DataFrame.

Why this is safe:
- This does not relax or change the optimization model.
- It only cleans up numerical noise at the result boundary, so downstream tests and consumers see the intended schedule instead of floating-point residue.

Validation:
- `uv run pytest tests/test_optimization.py -k 'load_deactivation_zero_operating_timesteps or load_deactivation_reactivation_on_cache_hit' -vv`
- `uv run pytest tests/test_optimization.py -q`

This looks incidental to the main purpose of the PR, but it unblocks the branch and makes the exported optimization results more robust.

kallegrens:
Perhaps we can get a new release soon so I can return to the upstream image instead of self-hosting my fork? :)

davidusb-geek:
> Perhaps we can get a new release soon so I can return to the upstream image instead of self-hosting my fork? :)

Ok thanks!
And yes I'll put a release very soon, probably tonight

davidusb-geek (approved):

PR #796, opened 2026-04-21
Update optimization.py Fix logic flaw in grid/battery interaction constraints.

### LINKED ISSUE #795: Logical flaw with set_nodischarge_to_grid=True: Battery drains to cover Load to maximize PV export, causing desync with physical inverters.
### Description
I have encountered a logical flaw in how EMHASS handles the `set_nodischarge_to_grid = True` constraint. The current constraint seems to only limit grid export to the total PV production (`P_export <= P_pv`). However, it ignores the physical routing priorities of hybrid inverters, leading to mathematically "correct" but physically impossible plans.

### The Problem
When the selling price is high, EMHASS wants to maximize export. To do this without violating the `P_export <= P_pv` constraint, it commands the battery to discharge and cover 100% of the house load (`P_load`). This allows 100% of the PV production to bypass the house and go straight to the grid. 

**Example from my optimization:**
* **PV Production:** 3607 W
* **House Load:** 2538 W
* **Grid Export:** 3607 W
* **Battery Discharge:** 2728 W *(exactly 2538 W after efficiency loss)*

### Why this breaks the system (The Negative Price Scenario)
While this works in the linear programming math, physical hybrid inverters (like Deye) prioritize PV to Load. If PV > Load, the inverter will *never* discharge the battery to cover the load. 

This causes a severe desync between EMHASS and reality, especially on days with negative prices:
1. EMHASS sees negative prices coming at 13:00.
2. It tries to empty the battery before 13:00 to make room for free solar energy.
3. To empty the battery, it uses the "trick" mentioned above (discharging to load while exporting PV).
4. The physical inverter ignores this command, because PV is already covering the load.
5. `naive_mpc_optim` runs again, sees the battery is still full, and plans the exact same impossible trick.
6. When 13:00 and negative prices hit, the battery is still half-full, solar production goes to waste or is exported at negative prices, ruining the profit.

### Suggested Solution
The mathematical model needs a constraint preventing the battery from discharging to the load if PV is sufficient to cover the load (unless `grid_forbid` is false and it's actually doing grid arbitrage). PV must satisfy the load *before* battery discharge is considered.

Image

Image

torsteinelv:
### Description
This PR addresses a logic flaw in the `set_nodischarge_to_grid` and `set_nocharge_from_grid` constraints, as discussed in issue #795.

The previous implementation allowed the battery to discharge to cover the house load while simultaneously exporting 100% of PV production to the grid. While mathematically "correct" under the constraint $P_{export} \le P_{pv}$, it is physically impossible for most hybrid inverters (like Deye) which prioritize PV to Load. This led to a desynchronization between EMHASS plans and real-world execution, especially critical during negative price periods.

### Changes
* Updated `src/emhass/optimization.py`: Loaded the grid direction binary variable `D` into the `_add_battery_constraints` function.
* **No discharge to grid:** Replaced the previous power-balance check with a strict binary constraint: `E <= D`. This ensures the battery cannot discharge ($E=1$) if the grid is exporting ($D=0$).
* **No charge from grid:** Replaced the check with `D <= E`. This ensures the battery cannot charge ($E=0$) if the grid is importing ($D=1$).

### Testing Performed
* Validated the logic by running the full optimization test suite: `pytest tests/test_optimization.py`.
* **Result:** 81 passed.
* *Note:* I have verified that the mathematical model is consistent with existing test cases. I have not performed extensive live production testing over several weeks, but the change directly aligns the model with the physical behavior of hybrid inverters.

Fixes #795

## Summary by Sourcery

Bug Fixes:
- Prevent the optimizer from planning simultaneous PV export and battery discharge or charge states that are infeasible for typical hybrid inverters by tightening grid/battery interaction constraints.

### DIFF
--- a/src/emhass/optimization.py
+++ b/src/emhass/optimization.py
@@ -1149,9 +1149,9 @@ class Optimization:
 
         # Grid Interaction Constraints
 
-        # No charge from grid: Battery cannot charge (E=0) while grid is importing (D=1)
+        # No charge from grid: The battery charge power cannot exceed the PV production
         if self.optim_conf["set_nocharge_from_grid"]:
-            constraints.append(D <= E)
+            constraints.append(p_sto_neg + p_pv >= 0)
 
         # No discharge to grid: Battery cannot discharge (E=1) while grid is exporting (D=0)
         if self.optim_conf["set_nodischarge_to_grid"]:
--- a/src/emhass/optimization.py
+++ b/src/emhass/optimization.py
@@ -1126,6 +1126,7 @@ class Optimization:
         p_sto_neg = self.vars["p_sto_neg"]
         p_grid_neg = self.vars["p_grid_neg"]
         E = self.vars["E"]  # Binary: 1=Discharge, 0=Charge
+        D = self.vars["D"]  # Binary: 1=Import, 0=Export
         p_pv = self.param_pv_forecast
 
         # Parameters (Scalars)
@@ -1148,13 +1149,13 @@ class Optimization:
 
         # Grid Interaction Constraints
 
-        # No charge from grid: Charging power (neg) + PV must be positive (net surplus)
+        # No charge from grid: Battery cannot charge (E=0) while grid is importing (D=1)
         if self.optim_conf["set_nocharge_from_grid"]:
-            constraints.append(p_sto_neg + p_pv >= 0)
+            constraints.append(D <= E)
 
-        # No discharge to grid: Grid Export (neg) + PV must be positive
+        # No discharge to grid: Battery cannot discharge (E=1) while grid is exporting (D=0)
         if self.optim_conf["set_nodischarge_to_grid"]:
-            constraints.append(p_grid_neg + p_pv >= 0)
+            constraints.append(E <= D)
 
         # Dynamic Power Limits (Ramp Rate)
         if self.optim_conf["set_battery_dynamic"]:


davidusb-geek:
Hi.

The proposed $E \le D$ for set_nodischarge_to_grid seems just fine and thanks for catching this.

However the other part for set_nocharge_from_grid seems flawed.

The proposed Constraint is $D \le E$, so if importing $D=1$, battery must be discharging/idle $E=1$.

Maybe this was added for mathematical symmetry, but tying the variables this way breaks a physically valid and strategically important optimization path.

The proof here.
Let's imagine a scenario where it is mid-morning:
- Load: 5000 W (Oven and AC are running)
- PV Production: 2000 W
- Grid Prices: Currently very cheap, but will spike to negative/expensive later in the day.

The mathematically optimal (and physically possible) move here is to route the 2000 W of PV directly into the battery to save it for later, and import 5000 W from the grid to cover the heavy house load. Many inverters support this via "Battery First" or scheduled charging modes.

Under the old constraint: ($P_{charge} \le P_{pv}$) -> You are allowed to charge the battery up to 2000 W. The solver imports 5000 W to cover the load. This works perfectly.

Under your new constraint: ($D \le E$) -> Because you are importing 5000 W, $D = 1$. The constraint forces $1 \le E$, meaning $E$ must be $1$. The battery is forced into the discharge state. The Result is that your new constraint makes it mathematically impossible to charge the battery from PV if the house is simultaneously pulling any power from the grid. You lose the ability to harvest solar while importing.

davidusb-geek:
I think the possible fix could look something like this:

```
if self.optim_conf["set_nocharge_from_grid"]:
    constraints.append(p_sto_neg + p_pv >= 0)

if self.optim_conf["set_nodischarge_to_grid"]:
    constraints.append(E <= D)
```

torsteinelv:
yes, i updated it and testing it in production now. Looks good so far.

davidusb-geek (approved):

PR #800, opened 2026-04-22
Add sticky first row and column to optimization results table

carposio:
Currently, the table header as well as the index column scroll away, which makes reviewing details in a big optimization results table difficult.

This PR adds sticky header and sticky index column with the help of claude code.

## Approach
Sticky first column: CSS position:sticky on td:first-child. The column can use CSS because the horizontal scroll container (overflow-x:auto on .table_div) does not trap sticky:left anchoring.

Sticky header row: JS fixed-clone overlay. CSS sticky:top cannot be used because the same overflow-x:auto scroll container also acts as a scroll boundary for the vertical axis, trapping sticky:top inside it. A position:fixed clone is used instead, syncing horizontal scroll via wrapper.scrollLeft.

border-collapse:separate; border-spacing:0 ensures sticky cells retain their own borders (with collapse, borders are shared between adjacent cells and lost when a sticky cell creates its own stacking context).

## Result

Here a number of screenshots. The sticky column makes reduces horizontal space on mobile devices, but I think it's still much more useful.

| Dark mode | Light mode |
|-----------|------------|
|  |   |
|  |   |
|  |  |
|  |   |
|  |   |

## Summary by Sourcery

Add sticky first row and column behavior to optimization results tables to keep headers and index visible while scrolling.

New Features:
- Introduce a sticky header row for optimization results tables using a fixed cloned header synchronized with scrolling.
- Make the first column of optimization results tables sticky so row labels remain visible during horizontal scrolling.

Enhancements:
- Adjust table border and spacing styles so sticky header and first column render correctly in both light and dark themes.
- Version the main stylesheet to ensure clients load the updated table styling.

### DIFF
--- a/src/emhass/static/script.js
+++ b/src/emhass/static/script.js
@@ -8,6 +8,7 @@
 //on page reload get saved data
 window.onload = async function () {
   await loadBasicOrAdvanced();
+  initStickyTables();
 
   //add listener for basic and advanced html switch
   document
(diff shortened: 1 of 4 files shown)


davidusb-geek:
Very nice improvement, thanks for this.
Not well versed at all in html/css stuff

davidusb-geek (approved):

PR #803, opened 2026-04-23
feat: Support arithmetic expressions in InfluxDB var_list entries

dewi-ny-je:
Original Pull with full Sorcery review and diagrams in: https://github.com/dewi-ny-je/emhass/pull/1

**NOT TESTED YET, ONLY PROPOSAL FOR MAINTAINER**
It looks quite reasonable though!

Also, **test files not updated**, to be done at later stage.

### Motivation

- Enable `var_list` entries to express simple arithmetic between multiple InfluxDB time series (e.g. `{{'sensor.a' - 'sensor.b' * 1000}}`) so arithmetic can be performed server-side while retrieving data. 

### Description

- Added expression detection and parsing utilities: `_is_influx_expression`, `_extract_influx_expression_entities`, and `_evaluate_influx_expression` to detect `{{ ... }}` entries, extract quoted entity IDs, map them to safe tokens, and evaluate arithmetic over pandas `Series` using a restricted AST whitelist. 
- Updated `get_data_influxdb` to fetch each referenced entity once (per-entry caching via `sensor_cache`), build a `series_mapping` of tokens to fetched `Series`, evaluate expressions into a `DataFrame` column, and keep backward compatibility for plain sensor strings. 
- Added imports for `ast` and `re` and guarded expression evaluation to allow only arithmetic and numeric constants for safety. 

### Testing

- Ran a compilation check with `python -m compileall src/emhass/retrieve_hass.py`, which completed successfully. 
- No additional automated unit tests were added in this change.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_69cb833f8e988332a115406855de727b)

## Summary by Sourcery

Support arithmetic expressions in InfluxDB variable definitions when retrieving data from InfluxDB and ensure failures are surfaced cleanly.

New Features:
- Allow InfluxDB var_list entries to be defined as arithmetic expressions over multiple time series using a {{ ... }} syntax.
- Introduce helpers to detect, parse, and safely evaluate arithmetic expressions that reference quoted entity IDs and operate on pandas Series.

Bug Fixes:
- Report and abort InfluxDB retrieval when any requested variables (sensors or expressions) fail to return data.

Enhancements:
- Cache InfluxDB sensor queries per entity to avoid redundant fetches when the same entity is used multiple times or across expressions.

## Summary by Sourcery

Support arithmetic expressions in InfluxDB variable retrieval and ensure retrieval fails cleanly when any requested variable has no data.

New Features:
- Allow InfluxDB var_list entries to be defined as arithmetic expressions over multiple time series using a {{ ... }} syntax.
- Provide helpers to detect, parse, and safely evaluate arithmetic expressions over pandas Series when fetching InfluxDB data.

Bug Fixes:
- Abort InfluxDB retrieval and log an error when any requested variables, including expressions, return no data.

Enhancements:
- Cache InfluxDB sensor queries per entity to avoid redundant fetches when entities are reused across variables and expressions.

### DIFF
--- a/src/emhass/retrieve_hass.py
+++ b/src/emhass/retrieve_hass.py
@@ -1,8 +1,10 @@
 import asyncio
+import ast
 import copy
 import logging
 import os
 import pathlib
+import re
 import time
 from datetime import datetime, timezone
 from typing import Any, Self
@@ -647,18 +649,51 @@ class RetrieveHass:
         if end_time < requested_end:
             self.logger.debug(f"End time capped at current time (requested: {requested_end})")
 
-        # Collect sensor dataframes
+        # Collect sensor dataframes (raw sensors and expression outputs)
         sensor_dfs = []
+        sensor_cache: dict[str, pd.DataFrame | None] = {}
         global_min_time = None
         global_max_time = None
 
-        for sensor in filter(None, var_list):
-            df_sensor = self._fetch_sensor_data(client, sensor, start_time, end_time)
-            if df_sensor is not None:
-                sensor_dfs.append(df_sensor)
+        for variable in filter(None, var_list):
+            df_variable = None
+
+            if self._is_influx_expression(variable):
+                try:
+                    parsed_expression, entities, token_to_entity = self._extract_influx_expression_entities(variable)
+                    series_mapping: dict[str, pd.Series] = {}
+
+                    for token, entity in token_to_entity.items():
+                        if entity not in sensor_cache:
+                            sensor_cache[entity] = self._fetch_sensor_data(
+                                client, entity, start_time, end_time
+                            )
+                        df_entity = sensor_cache[entity]
+                        if df_entity is None:
+                            raise ValueError(f"No data available for entity '{entity}'")
+                        series_mapping[token] = df_entity[entity]
+
+                    expression_result = self._evaluate_influx_expression(
+                        parsed_expression, series_mapping
+                    )
+                    df_variable = pd.DataFrame({variable: expression_result})
+                    self.logger.debug(
+                        f"Evaluated InfluxDB expression '{variable}' with entities: {entities}"
+                    )
+                except Exception as e:
+                    self.logger.error(f"Failed to evaluate InfluxDB expression '{variable}': {e}")
+            else:
+                if variable not in sensor_cache:
+                    sensor_cache[variable] = self._fetch_sensor_data(
+                        client, variable, start_time, end_time
+                    )
+                df_variable = sensor_cache[variable]
+
+            if df_variable is not None and not df_variable.empty:
+                sensor_dfs.append(df_variable)
                 # Track global time range
-                sensor_min = df_sensor.index.min()
-                sensor_max = df_sensor.index.max()
+                sensor_min = df_variable.index.min()
+                sensor_max = df_variable.index.max()
                 global_min_time = min(global_min_time or sensor_min, sensor_min)
                 global_max_time = max(global_max_time or sensor_max, sensor_max)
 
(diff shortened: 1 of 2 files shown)


dewi-ny-je:
Before I go further with taking into account all the suggestions by Sourcery, and CodeCov which I'm not even sure what it is, I would prefer to get an initial opinion by you on the implementation.

davidusb-geek:
> Before I go further with taking into account all the suggestions by Sourcery, and CodeCov which I'm not even sure what it is, I would prefer to get an initial opinion by you on the implementation.

Yes sure thing, and thanks for the efforts developing this.

First security-wise is ok, specially using Python's `ast` with a strict whitelist of operators as this prevents code injection.

On thing that one should be careful here is: Time-Series Index Alignment
A possible problem will arise if `sensor.a` and `sensor.b` are raw data pulled from InfluxDB and their timestamps (the pandas index) is not perfectly match down to the second/millisecond. If you add two pandas Series with different indices, pandas will return a result containing all indices from both series, but with NaN values wherever the timestamps don't perfectly overlap.

So we need to be sure that the dataframes coming out of `_fetch_sensor_data` are already resampled to a common time grid before this arithmetic happens.

dewi-ny-je:
I think GROUP BY takes care of it but I'll check.

davidusb-geek:
> Before I go further with taking into account all the suggestions by Sourcery, and CodeCov which I'm not even sure what it is, I would prefer to get an initial opinion by you on the implementation.

Ok, just hit a message here when you dev is done.
Don't worry too much about Codecov, as long as you have introduced test methods for your new devs thats ok for me

LesIT1:
@dewi-ny-je are you still planning to take this forward? It's a genuinely useful feature and David was positive on the approach, so it'd be a shame to let it go stale.

The two things left from his review are:

1. The time-index alignment he raised: two InfluxDB series can land on slightly different timestamps and inject `NaN` when you do the arithmetic. The existing `GROUP BY time() FILL(previous)` handles most of it, but a leading gap (a bucket with no earlier value to fill from) can still drop a row, so an explicit reindex of the series to a shared grid before evaluating would make it robust.
2. Unit tests for the new expression helpers (detection, entity extraction, the safe-eval whitelist including rejecting non-arithmetic input, and a deliberately misaligned-index case).

If you're not planning to continue, I'm happy to carry it forward on current master with the alignment fix and tests, crediting you and this PR. Didn't want to step on it if you're still on it. Either way, no rush.

dewi-ny-je:
@LesIT1 

> are you still planning to take this forward? It's a genuinely useful feature and David was positive on the approach, so it'd be a shame to let it go stale

Currently this project is low priority in my case, because it requires quite some time and my current automations for planning battery (simulated battery via battery_sim) are covering the day and next day quite well.

Also, I wanted to be clear about me not being a developer (I can read code decently, but I can't write it). It would be a continuous back and forth with LLMs.
My purpose was to show the level of complexity involved, since I was confident it was not too high, while the benefits for the community are significant (install EMHASS and bam! immediately have it able to use historical data, not waitin a YEAR before being able to get each month stats).

> If you're not planning to continue, I'm happy to carry it forward on current master with the alignment fix and tests, crediting you and this PR. Didn't want to step on it if you're still on it. Either way, no rush

If you want to continue, you surely have my blessing. When crediting me, please limit it to "Idea by", since I didn't code it really myself.

> The two things left from his review are:
> 
> 1. The time-index alignment he raised: two InfluxDB series can land on slightly different timestamps and inject `NaN` when you do the arithmetic. The existing `GROUP BY time() FILL(previous)` handles most of it, but a leading gap (a bucket with no earlier value to fill from) can still drop a row, so an explicit reindex of the series to a shared grid before evaluating would make it robust.
> 2. Unit tests for the new expression helpers (detection, entity extraction, the safe-eval whitelist including rejecting non-arithmetic input, and a deliberately misaligned-index case).

Tests have been added, at least a proposal for them, untested by me.

About time alignment: since we know the "group by" duration, maybe an option would be to extend it by one (or more) unit(s) in the past, and then drop the first row(s) via python? there would always be a "previous" one. For example, my solar inverter (which uses zigbee communication) publishes data every 5 minutes. If EMHASS asks for 1 minute buckets, the series should start not 1, but 5 buckets early to miss an empty data point.

Another issue is that timeseries data won't match bucket intervals: a timeseries might store data points at each 00:30 seconds and we group by 1:00 (full minutes), the values provided would be too old/new and lead to a (likely) very small bias. Does influxdb (v1!) offer a way to interpolate, so that for example it would provide in the 1:00 bucket the time-weighed average of the two existing values at 00:30 and  1:30?
Again for my solar inverter example, it's even worse: if emhass uses 1 minute buckets, unless some interpolation takes place, the results would have constant readings for solar power for 5 buckets at time. Maybe not an issue given the regression which takes place, it's just something that has to be kept in mind.

Influxdb V2 using Flux kinds of solves ALL issues here, but it's not even supported anymore in V3 (even if it was really powerful!), so the choice of using V1 is good, but the extra buckets at the beginning will be needed.

LesIT1:
Thanks, that's all really helpful and the blessing is appreciated. I'll pick it up and credit it as "Idea by @dewi-ny-je".

The leading-bucket idea is a clean fix for the FILL(previous) gap: pad the query by enough buckets to guarantee a prior value and drop the extras after. Good call on the sub-bucket timestamp bias too. InfluxDB v1 can't do a time-weighted average inside GROUP BY, so for a source that publishes off-grid (your 5-minute inverter into 1-minute buckets) the early buckets just repeat the last value. The regression downstream should absorb most of that, but I'll note it in the docs so it isn't a surprise.

I've got a couple of other PRs in flight with David at the moment, so I'll start this once those settle rather than pile too much on at once. I'll build on current master with the alignment fix and your proposed tests as the starting point, and tag you on the PR when it's up in case you can sanity-check it against a real InfluxDB setup. No rush on your side.

dewi-ny-je:
I'll be glad to test it of course, since it doesn't break any existing setup in my case.

Quick note: keep in mind that a possible use case of this feature is taking care of changes in entity names over time: I don't remember how, but I know that some power-related timeseries changed name in my setup so in Grafana I usually use two entities in OR mode for graphs which I might want to plot in the far past.
In emhass I would simply put two entities summed together (no overlap is expected, after all).

So, the queries to InfluxDB should provide error messages in the logs when one fails, but it should not be blocking because it might very well be that one empty query is expected to be empty.

LesIT1:
Carried this forward in #973 (I could not push to your fork's `master`, so it is a fresh branch on mine). It keeps your expression parser and safe evaluator, adds the leading-bucket query padding for the time-alignment point David raised, and brings tests plus docs. I credited you as the idea author.

The alignment fix is your leading-bucket-padding idea: expression entities are queried one day earlier than the requested window and sliced back, so `FILL(previous)` can seed the first buckets from the most recent prior reading. I checked it against a real InfluxDB 1.8 container before settling on it.

If you ever get a chance to point it at your real setup that would be a great sanity check. Thanks again for the green light.

PR #811, opened 2026-04-25
Generalize Load cost forecast value

m8ram:
Minor update to use currency instead of EUR as in the other forecast.

## Summary by Sourcery

Documentation:
- Update the load cost forecast documentation to describe prices in generic currency per kWh instead of EUR per kWh.


davidusb-geek:
Thanks for your contribution

davidusb-geek (approved):

PR #830, opened 2026-04-29
fix(schema): correct 5 default mismatches in param_definitions.json

OptimalNothing90:
## Summary

Per maintainer Option-B decision (2026-05-12 review), this PR now contains a single schema normalization: the `input` field for `ignore_pv_feedback_during_curtailment` changes from `"bool"` to `"boolean"` to match the rest of the file (all other booleans use `"boolean"`).

The 4 default-value mismatches originally proposed here have been reverted. `param_definitions.json` is now treated as source-of-truth; a follow-up PR will update `config_defaults.json` to align with it for these 4 params:

- `historic_days_to_retrieve`: 9 → 2
- `inverter_ac_output_max`: 5000 → 0
- `inverter_ac_input_max`: 5000 → 0
- `load_forecast_method`: `"naive"` → `"typical"`

Behavior change for Headless users on those 4 params — release-notes-heads-up to follow in the `config_defaults` PR.

## Verification

- `git diff` against pre-revert state shows the 4 default-value reverts.
- The boolean-type fix stays (`ignore_pv_feedback_during_curtailment.input = "boolean"`).
- `pytest tests/` passes.
- `uvx ruff check .` clean.

## Sibling PR

Follow-up PR for `config_defaults.json` value changes + release notes will be filed separately. This PR is now minimal: 1 line of schema-normalization.

Closes follow-up dependency direction agreed in this thread.

### DIFF
--- a/src/emhass/static/data/param_definitions.json
+++ b/src/emhass/static/data/param_definitions.json
@@ -120,9 +120,9 @@
     },
     "historic_days_to_retrieve": {
       "friendly_name": "Historic days to retrieve",
-      "Description": "We will retrieve data from now to days_to_retrieve days. Defaults to 9",
+      "Description": "We will retrieve data from now to days_to_retrieve days. Defaults to 2",
       "input": "int",
-      "default_value": 9
+      "default_value": 2
     },
     "load_negative": {
       "friendly_name": "Load negative values",
@@ -154,7 +154,7 @@
       "Description": "The load forecast method that will be used. The options are ‘csv’ to load a CSV file or ‘naive’ for a simple 1-day persistence model.",
       "input": "select",
       "select_options": ["typical", "naive", "mlforecaster", "csv"],
-      "default_value": "naive"
+      "default_value": "typical"
     },
     "set_total_pv_sell": {
       "friendly_name": "PV straight to grid",
@@ -215,13 +215,13 @@
       "friendly_name": "Max hybrid inverter AC output power",
       "Description": "Maximum hybrid inverter output power from combined PV and battery discharge.",
       "input": "int",
-      "default_value": 5000
+      "default_value": 0
     },
     "inverter_ac_input_max": {
       "friendly_name": "Max hybrid inverter AC input power",
       "Description": "Maximum hybrid inverter input power from grid to charge battery.",
       "input": "int",
-      "default_value": 5000
+      "default_value": 0
     },
     "inverter_efficiency_dc_ac": {
       "friendly_name": "Hybrid inverter efficency DC to AC",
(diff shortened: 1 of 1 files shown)


davidusb-geek:
Hi, `historic_days_to_retrieve` should default to 2 (not 9 as proposed but yes there was a mismtach that needed fix).
`load_forecast_method` default should be "typical" (not "naive")
I can't remember why but I think that `inverter_ac_output_max` and `inverter_ac_input_max` should be dafaulted to 0, need to check this one

OptimalNothing90:
Can do, but this will change code default's as well.

For inverter, it seemed kind of philosophical to me, will change it too

davidusb-geek:
> Can do, but this will change code default's as well.
> 
> For inverter, it seemed kind of philosophical to me, will change it too

@OptimalNothing90  So it seems that this PR's is updating `param_definitions.json` based on values from `config_defaults.json`. But it should be the other way around, update the `config_defaults.json` based on `param_definitions.json`.

OptimalNothing90:
@davidusb-geek I checked the code path before flipping. `utils.py:1599-1638` in `build_config_emhass` loads `config_defaults.json` first, then user `config.json` overrides. `param_definitions.json` `default_value` is never read as a fallback in code — only displayed in the GUI configuration form. `develop.md` notes "should act as last-resort fallback if default_config.json is not found" but that path returns False today (line 1605).

So today there are two effective defaults per param:

- **Headless install** (Docker + `options.json`, no GUI save): `config_defaults` wins
- **GUI install** (user opens `/configuration`, saves): `param_def` value gets written into `config.json` and wins

That's the inconsistency. Direction matters because flipping changes behavior for Headless users:

| Param | config_defaults today | param_def today | Headless impact if flipped to param_def |
|---|---|---|---|
| `historic_days_to_retrieve` | 9 | 2 | ML forecaster trains on 7 fewer days |
| `load_forecast_method` | "naive" | "typical" | switches default to ML-trained model for users who never set it |
| `inverter_ac_output_max` / `_input_max` | 5000 | 0 | optimizer treats as unconstrained (no cap) |

The boolean fix (`bool` → `boolean` for `ignore_pv_feedback_during_curtailment`) is unaffected — pure normalization to match the rest of the file.

Two options:

**A)** Treat `config_defaults` as source-of-truth (PR #830 as-is) — fix `param_def` to show what code does. Headless behavior unchanged. GUI users see the actual runtime values.

**B)** Treat `param_def` as source-of-truth — revert this PR's value changes, open follow-up PR updating `config_defaults`. **Behavior changes for Headless users on those four params.** Worth a heads-up in release notes.

Your call. Either way I'll keep the boolean type fix.

davidusb-geek:
> **B)** Treat `param_def` as source-of-truth — revert this PR's value changes, open follow-up PR updating `config_defaults`. **Behavior changes for Headless users on those four params.** Worth a heads-up in release notes.

Option B is the way to go

davidusb-geek (approved):

Issue #846, opened 2026-05-13
[Feature request] Allow a "soft" minimum on battery SOC by introducing a virtual cost entry


lutorm:
It seems there are two kinds of minimums on battery SOC. One is a hard "you must not discharge the battery more than this" type of limit that depends on the battery chemistry, and the other is "I'd like to always keep this much energy in the battery for e.g. backup purposes." 
The rationale for the second limit being soft is that I'd be willing to let the battery drop below my preferred low if it's only for a short time because it's selling power at the morning price spike and I know PV will charge it up soon anyway than if it's going to sit all night or maybe several days at that level in the middle of winter. At that point the risk of having a power cut and not sufficient charge in the battery is much greater.

This type of limit could be implemented by adding a virtual cost to every e.g kWh*hour the battery spends below some specified SOC. It would be very similar to the existing virtual cost for inverter power, so should not be very hard to implement.

PR #859, opened 2026-05-15
Add a penalty to the battery SOC being below a threshold.

lutorm:
Created a PR for the feature suggested in https://github.com/davidusb-geek/emhass/issues/846. This is my first time working in this repo so I'd appreciate some guidance if this is OK. It appears to do what it's supposed to, but I'm not sure how to create test cases.

Ideally, I'd only like to apply this penalty cost if P_PV<P_Load, since if there's sufficient PV to immediately carry the load if the battery goes down, the risk is small. However, I could not figure out how to do this without getting infeasible results or the solver complaining about DCP.

## Summary by Sourcery

Introduce a configurable penalty in the optimization for operating the battery below a state-of-charge threshold and expose its results.

New Features:
- Add an SOC deficit cost term to the optimization objective when battery usage is enabled.
- Allow configuration of SOC deficit threshold and cost rate via plant configuration parameters.
- Expose the SOC deficit cost values in the optimization results for downstream use.

Enhancements:
- Extend battery constraint handling to compute SOC deficit cost based on current stored energy relative to the configured threshold.


lutorm:
Thanks for the points, ai bot. I added the cost as a variable to be able to see what it was calculating. I'll remove the >=0 constraint though.

davidusb-geek:
Thanks for this contribution

lutorm:
@davidusb-geek Can you give me some pointers for the test standards, do I have to design a test case with a known optimization result?

davidusb-geek:
> @davidusb-geek Can you give me some pointers for the test standards, do I have to design a test case with a known optimization result?

Yes exactly. A test that check what your method is returning and that the values are what you expect

davidusb-geek:
Ok, so will merge after unit test are added. Just drop a message here when you think its good to go.
It will be for the release after the one going out tonight

lutorm:
@davidusb-geek I think this should be good if the tests pass, but please take a look. Just to double check: the "common normalization" scale of 0.001 on line 916 of optimization.py is a kW -> W conversion, right? I had included it in my term of the objective function, but I'd already converted my cost rate to /Wh and was getting smaller penalties than anticipated with the scale included.

davidusb-geek:
> @davidusb-geek I think this should be good if the tests pass, but please take a look. Just to double check: the "common normalization" scale of 0.001 on line 916 of optimization.py is a kW -> W conversion, right? I had included it in my term of the objective function, but I'd already converted my cost rate to /Wh and was getting smaller penalties than anticipated with the scale included.

Yes that is the conversion of W to kW

davidusb-geek (approved):

Issue #539, opened 2025-05-13
[Feature request] Hybrid heating source


snauwaertc:
Is it possible to build functionality to control a hybrid heating system (e.g., gas and heat pump), taking into account gas EUR/kWh and implied electrical EUR/kWh (based on predicted internal price, COP/heat differential etc.)? 
There may be a circular dependency for heat source and implied electrical energy price.

snauwaertc:
## Hybrid heating PR series

I've split the hybrid-heating work into a small chain of focused PRs.
Each one is additive, opt-in, and independently testable. All 4 are
ready to push at the same time, but they stack so that PR2 builds on
PR1, etc. When PR1 merges, PR2's base auto-updates to master and so on.

| # | Branch | Title | Net | Risk |
|---|---|---|---|---|
| 1 | #860 | `feat: constant-efficiency mode for thermal_battery (gas / oil / district)` | ~120 lines | low |
| 2 | #861 | `feat: per-load cost forecasts and surface solar gain` | ~280 lines | low |
| 3 | #862 | `feat: heat_topology graph model for hybrid heating` | ~1,200 lines | medium |
| 4 | #863 | `feat: weather-compensated supply T and buffer floor` | ~690 lines | low |

### Why split

`AGENTS.md` section 5 warns "touches optimization.py beyond about three
lines -> file an issue first". This issue is the issue. The cumulative
diff against master is ~2,950 lines across 6 files in `src/emhass/`,
which is too big to review in one shot. Splitting along feature
boundaries keeps each PR scoped to one capability with its own tests
and docs.

### Architecture story

The thread through all four PRs is "make thermal_battery composable":

1. **PR 1** unlocks non-heat-pump sources (flat efficiency).
2. **PR 2** lets each source bill on its own cost track and gain heat
   from the sun.
3. **PR 3** introduces a graph model so users can declare arbitrary
   topologies (multi-source tanks, shared actuators, mixed electric /
   commodity sources) instead of hand-aligning half a dozen
   `optim_conf` arrays. Adds the `is_electric_load` flag so non-electric
   loads don't load `p_def_sum`.
4. **PR 4** adds weather compensation: heating curve on source, min-T
   curve on storage, soft-comfort fields plumbed through the compiler.
   Documents the direct-heating pattern (no buffer; building IS the
   storage).

### Invariants preserved (per `AGENTS.md` section 3)

- `action_logs.txt` format unchanged
- `get_logger` handler-accumulation guarded
[...]

PR #860, opened 2026-05-16
feat: constant-efficiency mode for thermal_battery (gas / oil / district)

snauwaertc:
## Problem

`thermal_battery` currently assumes a heat-pump source: the COP is computed
with Carnot physics based on `outdoor_temperature_forecast` and
`supply_temperature`. Non-electric sources (gas boilers, oil burners, district
heating substations) don't have a Carnot COP - they convert primary energy at
a roughly flat efficiency. To model hybrid heating where one tank is fed by a
heat pump and another by a gas boiler, EMHASS needs to support both source
types behind the same `thermal_battery` interface.

## Change

Adds an optional `efficiency` field on the `thermal_battery` sub-config. When
set, EMHASS uses a flat per-timestep conversion factor instead of computing
Carnot COP. Existing configs (Carnot path) are unchanged.

A new helper `utils.resolve_thermal_battery_cop()` dispatches between the two
modes:

- if `efficiency` is present → flat array `[efficiency] * n_timesteps`
- else → existing `calculate_cop_heatpump()` path

All three COP call sites in `optimization.py` (`_add_thermal_battery_*`
methods, both per-load and shared paths) route through the helper.
`supply_temperature` is no longer hard-required; either it or `efficiency`
must be present, otherwise a clear `ValueError` is raised.

## Compatibility

- Existing configs with `supply_temperature` (and optional
  `carnot_efficiency`) keep working exactly as before. The helper picks the
  Carnot path when `efficiency` is missing.
- All 27 existing `thermal_battery` tests pass.
- No changes to optim_conf cache key shape.

## Test plan

- [x] `pytest tests/test_utils.py::TestResolveThermalBatteryCop` (9 unit
      tests: flat efficiency, both fields present, missing both,
      negative-value validation, etc.)
- [x] `pytest tests/test_optimization.py -k "thermal_battery and efficiency"`
      (3 integration tests: optimizer solves with flat efficiency,
      efficiency overrides Carnot when both set, missing-both raises)
- [x] `pytest tests/test_optimization.py` - all 32 thermal_battery tests pass
- [x] No new ruff errors (pre-existing F841 + I001 in upstream code are
      unchanged)

## Docs

New "Constant-efficiency mode" section in `docs/thermal_battery.md` with
typical efficiency values (gas 0.92, oil 0.85, district 0.95) and a note on
modelling hybrid systems via per-target `thermal_battery` configs combined
with `deferrable_load_groups`.

## Related

This is the first in a small chain of additive changes that, together,
enable hybrid heating modelling (heat pump + gas boiler, shared tanks,
weather-compensated supply temperature). Each change is independently
useful and shipped as a focused PR.

Refs #539.

## Summary by Sourcery

Add support for constant-efficiency heat sources in the thermal_battery model alongside existing Carnot-based heat-pump mode.

New Features:
- Allow thermal_battery to use a flat efficiency parameter as a constant energy-conversion factor for non-heat-pump sources such as gas, oil, or district heating.
- Introduce a resolve_thermal_battery_cop helper that selects between Carnot COP and flat-efficiency modes based on the thermal_battery configuration.

Enhancements:
- Relax thermal_battery configuration so that either supply_temperature or efficiency is required, improving flexibility and error messaging.
- Update optimization logic to use the unified COP resolver for both per-load and shared thermal_battery constraints.
- Extend thermal_battery documentation with guidance and typical values for constant-efficiency sources and hybrid system modelling.

Tests:
- Add unit tests for the COP resolver helper covering flat-efficiency, Carnot, length handling, and validation cases.
- Add integration tests to ensure constant-efficiency thermal_battery configurations solve correctly, override Carnot when both are set, and raise errors when required fields are missing.

### DIFF
--- a/docs/thermal_battery.md
+++ b/docs/thermal_battery.md
@@ -68,6 +68,22 @@ These parameters define the basic thermal battery system:
     * Example: `0.42`
     * See the calibration section below for how to determine this
 
+### Constant-efficiency mode (gas boiler, oil burner, district heating)
+
+The default thermal battery model uses a Carnot-based COP that varies with outdoor temperature - appropriate for heat pumps. Non-electric heat sources (gas boilers, oil burners, district heating) convert input power to heat at a roughly constant efficiency that does not depend on outdoor temperature. For these sources, set the `efficiency` parameter:
+
+* **efficiency**: Optional constant energy-conversion factor (output thermal kW / input kW). When set, EMHASS skips the Carnot COP calculation and uses this flat value for every timestep. `supply_temperature` and `carnot_efficiency` become optional in this mode.
+    * Typical values:
+        * Modern condensing gas boiler: 0.90-0.95
+        * Standard gas boiler: 0.85-0.90
+        * Oil burner: 0.80-0.90
+        * District heating substation: 0.95-0.98
+        * Direct electric heater: 1.0
+    * Example: `0.9` for a condensing gas boiler
+    * When both `efficiency` and heat-pump fields are present, `efficiency` takes precedence.
+
+Configure one `thermal_battery` per source-target pair to model hybrid systems. For example, a gas boiler serving both a heating buffer and a DHW tank is two `def_load_config` entries (one per target), grouped via `deferrable_load_groups` with `mutual_exclusion: true` and `max_power` equal to the boiler's modulating maximum if the two targets share the same physical actuator.
+
 ### Heating demand calculation
 
 EMHASS needs to know how much heat your building requires. There are two methods to calculate this, and EMHASS automatically selects the appropriate method based on which parameters you provide.
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -72,7 +72,7 @@ emhass = { workspace = true }
 [tool.uv]
 default-groups = "all"
 package = true
-override-dependencies = ["yarl!=1.24.1"]
+override-dependencies = ["yarl<1.24"]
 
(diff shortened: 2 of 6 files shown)


snauwaertc:
Thanks @sourcery-ai. Addressed both points in the force-push above:

**Type hint mismatch / `.tolist()` calls**
- Broadened the type hint to `Sequence[float] | np.ndarray | pd.Series | None`.
- Dropped the three `.tolist()` calls at the optimization-side call sites - they were unnecessary; arrays go straight through.

**Length inferred from unused `outdoor_temperature_forecast`**
- In constant-efficiency mode, `outdoor_temperature_forecast` may now be `None`. When it is, `length` must be set explicitly; otherwise we raise with a clear message instead of silently falling back to a meaningless `len()`.
- In heat-pump mode, `outdoor_temperature_forecast` remains mandatory and raises with a clear message when missing.

Added 3 unit tests covering the new behaviour (`test_efficiency_mode_accepts_outdoor_none_when_length_given`, `test_efficiency_mode_outdoor_none_without_length_raises`, `test_heatpump_mode_outdoor_none_raises`).

While in the file I also removed a duplicate `supply_temperature` / `efficiency` presence check that lived in `_add_thermal_battery_constraints`: the helper is now the single source of truth for that validation, matching the same suggestion you raised on PR #861.

davidusb-geek:
Thanks for this contribution and a very complete PR: clear code, docs and unit tests

davidusb-geek (approved):

PR #863, opened 2026-05-16
feat: weather-compensated supply T and buffer floor for thermal_battery

snauwaertc:
## Problem

Two practical gaps left after the graph compiler lands, plus a doc
addition for a setup family we don't yet have an example for:

1. Modern hydronic systems use **weather-compensated supply
   temperature**: a linear law `T_supply = offset - slope * T_outdoor`,
   bounded by a min/max. The `thermal_battery` Carnot path currently
   assumes a fixed supply T, which under-estimates heat pump COP in
   mild weather (the supply T tracks outdoor T down, so the Carnot
   lift shrinks).
2. For radiator-based systems, the buffer **minimum temperature
   needs to track the weather** the same way: cold outdoor -> hot
   buffer, mild outdoor -> can run cooler. A static floor either
   wastes energy in mild weather or starves the radiators in cold
   weather.
3. **Direct heating** (no buffer tank, boiler/HP feeds radiators
   directly) is supported today but undocumented. Users have to
   discover that the building thermal mass plays the role of the
   storage node.

## Change

**1. `heating_curve` on source** - `apply_heating_curve()` returns a
clipped per-slot supply T array.
`resolve_thermal_battery_cop()` picks up the curve when the source
has a `heating_curve` block instead of a fixed `supply_temperature`.
`calculate_cop_heatpump()` accepts a scalar or per-slot array supply
T transparently.

**2. `min_temperature_curve` on storage** -
`resolve_min_temperatures()` returns the elementwise max of static
`min_temperatures` (absolute safety floor) and the curve
(weather-compensated radiator-emission floor).
`_add_shared_thermal_tank_constraints` calls the resolver to build
the per-slot floor.

**3. Soft comfort fields propagated through compiler** - the
existing `desired_temperatures` / `overshoot_temperature` /
`penalty_factor` / `comfort_sense` machinery on the optimizer side is
plumbed through `compile_heat_topology` so users can set them on the
storage block in `heat_topology`. No optimizer changes.

**4. Direct heating doc example** - new Example 6b in
`docs/thermal_battery.md` showing the building-as-storage pattern
(volume = effective thermal mass, no `draw_off_demand` -> defaults to
space-heating mode).

## Compatibility

- All four are opt-in. Defaults are no-op:
  - source without `heating_curve` -> static `supply_temperature` path
  - storage without `min_temperature_curve` -> static
    `min_temperatures` path
  - storage without soft-comfort fields -> hard constraints (existing
    behavior)
- 4-step parameter wiring followed; `OptimizationCacheKey` unchanged
  (the curve fields live inside existing source/storage blocks).

## Test plan

- [x] `pytest tests/test_utils.py::TestApplyHeatingCurve` - 5 unit
      tests (linear law, clipping, scalar/array supply T,
      length-handling)
- [x] `pytest tests/test_utils.py::TestResolveMinTemperatures` - 7
      unit tests (static-only, curve-only, max-of-both,
      length-handling, validation)
- [x] `pytest tests/test_utils.py::TestResolveThermalBatteryCopHeatingCurve` -
      4 unit tests (heating curve picks per-slot supply T, COP varies
      slot by slot)
- [x] Compile-side tests: 5 new cases in `TestCompileHeatTopology`
      covering propagation of `heating_curve`,
      `min_temperature_curve`, and soft-comfort fields, with
      validation of slope+offset / min_supply ordering.
- [x] `pytest tests/test_utils.py tests/test_optimization.py` - 201
      tests pass (1 deselected: pre-existing upstream failure).

## Stack

This PR is the fourth and final in a chain. The base is the
`heat_topology` graph model PR (the `heating_curve` and
`min_temperature_curve` fields live on the source / storage blocks in
`heat_topology`, and the soft-comfort propagation goes through the
graph compiler).

Refs #539.

## Summary by Sourcery

Extend the thermal battery and heat-topology system to support weather-compensated operation, shared thermal tanks, and per-load cost signals, improving modeling of real-world hybrid heating systems while keeping legacy behavior as the default.

New Features:
- Allow heat pump COP calculation to accept time-varying supply temperatures driven by a configurable heating curve and derive per-slot COP for thermal battery sources.
- Introduce weather-compensated minimum temperature curves for storage, combining static safety floors with outdoor-temperature-dependent radiator emission floors.
- Add support for constant-efficiency thermal battery sources (e.g., gas, oil, district heating) that bypass Carnot COP and use a flat conversion factor.
- Model solar gains on exposed thermal masses via surface absorption of GHI, reducing effective heating demand when solar input is present.
- Compile a declarative heat topology (sources, storage, consumers, actuator groups) into optimizer configuration, including shared thermal tanks fed by multiple sources and per-load electric/non-electric flags.
- Support per-deferrable-load cost forecasts so hybrid sources can be priced against different commodity tariffs within a single optimization run.

Enhancements:
- Refine thermal battery and shared-tank constraint construction to share tank dynamics across multiple loads and to apply weather-compensated minimum temperature floors.
- Adjust electric power-balance and objective construction so only electric loads contribute to grid exchange while non-electric loads are still costed correctly via per-load tariffs.
- Improve COP logging for non-physical conditions by handling scalar and array supply temperatures and reporting supply ranges.
- Extend optimization parameter handling to wire in per-load cost parameters and reinitialize them correctly on moving horizons.
- Integrate heat-topology compilation into runtime configuration loading while keeping the optimization cache key minimal but structurally aware of shared tanks and electric flags.

Documentation:
- Update thermal battery documentation with sections on surface solar gain, heating-curve operation, constant-efficiency sources, weather-compensated minimum temperatures, and a direct-heating (no buffer) example.
- Document per-deferrable-load cost override mechanics for hybrid heating scenarios in the forecasts documentation.

Tests:
- Add extensive unit tests for heating curves, minimum-temperature resolution, COP resolution modes, solar gain calculation, and compilation of heat topologies including validation paths.
- Extend optimization tests to cover flat-efficiency thermal batteries, shared tanks with multiple sources, electric vs non-electric load handling, solar gain effects on dispatch, per-load cost overrides, and regression against legacy single-source behavior.


snauwaertc:
Thanks @sourcery-ai. The bug-risk individual comment was real and addressed.

**Scalar `min_temperature` raising `TypeError`**
- Reproduced: `cfg = {"min_temperature": 20.0}` would raise `TypeError: 'float' object is not iterable` in the prior code because of `list(static)`.
- `resolve_min_temperatures` now normalises scalars first: a single `int` or `float` under the singular `min_temperature` key is treated as a one-element list (which the existing padding logic then expands to the horizon).
- Both the plural list form (`min_temperatures: [...]`) and the singular scalar form (`min_temperature: 20.0`) work.
- Added `test_scalar_min_temperature_normalised` (covers both `int` and `float`).

The two overall-comments (`_load_shared_tank_membership` cached once; profile aggregation truncation) were also raised on PR #862 and were addressed in that PR; the fix propagated down to this branch via the rebase.

davidusb-geek (approved):

PR #861, opened 2026-05-16
feat: per-load cost forecasts and surface solar gain on thermal_battery

snauwaertc:
## Problem

Two small but separately useful gaps in `thermal_battery`:

1. Every deferrable load is priced from the same `load_cost_forecast`. In a
   hybrid setup, one load can be a heat pump on a dynamic retail tariff and
   another a gas boiler at a flat per-kWh gas price. Today the optimizer
   prices both at the retail tariff.
2. Outdoor thermal targets (uncovered pools, outdoor buffer tanks,
   solar-thermal collector apertures) receive a non-negligible heat input
   from solar absorption that EMHASS currently ignores.

## Change

Two additive, opt-in changes shipped together because they share
`optim_conf` plumbing in `command_line.py` and one of them needs the
constant-efficiency mode from the previous PR.

**1. Per-load cost forecast** (`cost_forecast_per_deferrable_load`)

- Optional list, one entry per deferrable load. Each entry is either `None`
  (use global `load_cost_forecast`) or a per-timestep list.
- Implemented as a `cp.Parameter` so updating the price track does not
  invalidate the cached problem.
- Objective adds `(per_load_cost - load_cost) * p_deferrable[k]` per load
  when `costfun` is `profit` or `cost`. Zero when override is unset.
- Limitation: only adjusts cost; does not remove a load from the electric
  power balance constraint. A `non_electric` flag is deferred to a later
  PR.

**2. Surface solar gain** on `thermal_battery`

- New optional fields `solar_absorption_area` (m^2) and
  `solar_absorption_factor` (default 0.7).
- When `ghi` is present in `df_input_data_dayahead` (already fetched for
  PV forecast - no extra API call), absorbed solar energy is subtracted
  from `heating_demand`. The existing thermal-battery dynamics already
  treat this as a generalized heating gain, so no new constraints are
  needed.

## Compatibility

- Both fields are opt-in. Defaults are no-op:
  - `cost_forecast_per_deferrable_load = [null, null]` -> baseline dispatch
    unchanged.
  - `solar_absorption_area` unset -> no gain term applied.
- 4-step parameter wiring followed for both fields: `associations.csv`,
  `config_defaults.json`, `param_definitions.json`, and `command_line.py`
  (cache key / runtime keys).

## Test plan

- [x] `pytest tests/test_optimization.py -k "per_load_cost"` (2 tests:
      no-op when unset, dispatch shifts when penalty applied)
[...]

PR #872, opened 2026-05-20
fix(config): guard null defaults and add missing input types in buildParamElement

### LINKED ISSUE #869: v0.17.3 Config page
The EMHASS config page in v0.17.3 is messed up after updating:

Deferrable loads switch not available
Deferrable loads settings almost all gone
Battery settings not possible, only switch available
Button 'Yaml' not working
Save probably not working, I don't get the green confirmation mark

I checked the documentation thinking it might moved somewhere (battery settings), but apparently not

Image
Image
Image

OptimalNothing90:
Fixes #869 — config page crash in v0.17.3.

### Root cause

Two parameters added in v0.17.3 have input types the JS renderer doesn't know about:

| Parameter | Added by | Input type | Default |
|---|---|---|---|
| `cost_forecast_per_deferrable_load` | PR #861 | `array.array.float` | `[null, null]` |
| `heat_topology` | PR #863 | `object` | `null` |

`buildParamElement`'s switch has no cases for either type, so `type` and `placeholder` stay empty and `value` resolves to `null` or `[null, null]`. Both land in the object-rendering `else` branch, where `Object.values(null)` throws:

```
TypeError: Cannot convert undefined or null to object
```

The crash fires at `cost_forecast_per_deferrable_load` (entry #4 in Deferrable Loads). The section loop dies there: entries 5–13 don't render, Battery never builds, and the Yaml/Save listener binding never runs. This matches the symptoms in #869.

### What changed

Three edits in `configuration_script.js`:

1. `case "object"` and `case "array.array.float"` added to the switch, both mapping to `type = "text"` / `placeholder = JSON.stringify(default_value)`. The params now show up as editable text inputs.

2. Early return when `value === null`, before the `Object.values` call. Returns a single empty input; `heat_topology` stops crashing.

3. The nested-object branch condition now also checks `Object.values(value)[0] !== null`. Without it, `[null, null]` passes `typeof ... === "object"` (because `typeof null === "object"`) and crashes on the inner loop.

No Python, JSON, or template changes.

### Verification

Node.js replay of `buildParamElement` against the actual `param_definitions.json` and `config_defaults.json` from v0.17.3:

```
[1/13] OK   number_of_deferrable_loads (input=int)
[2/13] OK   nominal_power_of_deferrable_loads (input=array.float)
[3/13] OK   minimum_power_of_deferrable_loads (input=array.float)
[4/13] FAIL cost_forecast_per_deferrable_load (input=array.array.float)  ← pre-patch
        TypeError: Cannot convert undefined or null to object
```

Docker live test (emhass-i869 image, minimal config, v0.17.3 defaults):

**Before** — 49 rendered param divs, 1 JS error (`Cannot convert undefined or null to object`), Deferrable Loads entries 4–13 absent, Battery absent, Save/Yaml/Defaults buttons unbound:

Before patch

**After** — 59 rendered param divs, 0 JS errors, all Deferrable Loads and Battery sections present, buttons bound:

After patch

### Context

PRs #861 and #863 introduced the two affected parameters. The problem is that there's no contract between the input values in `param_definitions.json` and the cases in the JS switch. A follow-up will look at that.

### Related

- Fixes #869
- Followed contribution conventions per AGENTS.md

## Summary by Sourcery

Handle newly introduced configuration parameter types and prevent null-valued defaults from crashing the config UI renderer.

Bug Fixes:
- Fix config page crash when rendering parameters with null or array-of-null default values.
- Ensure parameters with input types `object` and `array.array.float` render as text inputs instead of breaking the page.


OptimalNothing90:
Fixed in cae1992 — `placeholder` is now quoted in the null-guard branch.

On `Object.values(value)[0]` evaluated twice: agreed, but the double evaluation predates this PR. Pulling it into a variable would touch code outside this diff, so leaving it for a refactor pass.

OptimalNothing90:
The failing jobs (test_forecast on ubuntu and macos) don't look related to this PR. The only change here is in configuration_script.js, no Python touched. Those same tests passed on the first CI run for this PR and nothing has changed in Python since. Looks like a date-sensitive test that started failing today. From my side everything looks good. The stage is yours.

davidusb-geek:
> The failing jobs (test_forecast on ubuntu and macos) don't look related to this PR. The only change here is in configuration_script.js, no Python touched. Those same tests passed on the first CI run for this PR and nothing has changed in Python since. Looks like a date-sensitive test that started failing today. From my side everything looks good. The stage is yours.

Got it, thanks for the hot fix, highly appreciated

davidusb-geek (approved):

snauwaertc:
@OptimalNothing90 thank you for this fix, and apologies for the breakage that made it necessary. The `heat_topology` param I added in #863 was the first `"input": "object"` field with a `null` default, and I didn't account for the config page having no render path for that combination, so it took out the whole section (deferrable loads, battery, Save/Yaml). Thanks for tracking down the `Object.values(null)` cause and getting users back up and running so quickly.

Issue #876, opened 2026-05-21
AttributeError: 'str' object has no attribute 'get'


ThomasCZ:
**Describe the bug**
After updating to v0.17.4, all POST requests to EMHASS fail with an `AttributeError: 'str' object has no attribute 'get'` in `utils.py` during `compile_heat_topology`. 

This is caused by a regression where the add-on configuration has `"heat_topology": "null"` (the default configuration string for unused heat integration), but the new thermal modeling engine treats it as a raw string instead of an object/dictionary, crashing when it tries to run `.get()`.

**To Reproduce**
1. Have `"heat_topology": "null"` in the add-on configuration (or runtime parameters).
2. Upgrade to v0.17.4.
3. Send any optimization request (e.g., POST `/action/naive-mpc-optim`).
4. The web server crashes with a 500 Internal Server Error.

**Home Assistant installation type**
 - Home Assistant OS - 17.3
 - Home Assistant Supervised - 2026.05.0
 - Home Assistant Core - 2026.5.3

**Your hardware**
- OS: HA OS
- Architecture: rpi5-64

**EMHASS installation type**
 - Add-on (v0.17.4)

**Additional context**
The crash happens here in `emhass/utils.py` at line 509:
```python
def compile_heat_topology(topology: dict) -> dict:
    ...
    sources = topology.get("sources", []) or []
```

Since topology contains the string "null", calling .get() throws the AttributeError.

Temporary Workaround / Fix:
Changing the add-on configuration from:

```
"heat_topology": "null"
```

to a valid empty object:

```
"heat_topology": {
  "sources": []
}
```

immediately resolves the issue and optimization works again. The code should probably be guarded with an isinstance(heat_topology, dict) check or properly handle the "null" string case.

**Logs**

```
[2026-05-21 11:39:31,855] INFO in web_server:  >> Obtaining params: 

[...]

PR #879, opened 2026-05-22
fix(optimization): guard string-null cost_forecast_per_deferrable_load

OptimalNothing90:
Refs #876.

## Context

lutorm's follow-up in the #876 thread flagged that `cost_forecast_per_deferrable_load` has the same null-handling problem as `heat_topology` (#878). Both came in with the v0.17.3 thermal-PR wave, both have `null` defaults, and both can get string `"null"` from older HA add-on configs that serialize it that way.

Sibling to #878, disjoint files. Either can merge first.

## Root cause

The cost-override loop character-indexes a string. `"null"[0]` is `"n"`, so `np.asarray("n", dtype=float)` raises `ValueError`. A list like `[None, "null"]` survives the outer check but crashes at the per-element `np.asarray` call.

The shipped default `[null, null]` is fine — `None` hits the existing guard at line 2864 and never reaches `np.asarray`.

## What changed

Two isinstance guards in `src/emhass/optimization.py`:

1. After fetching `cost_per_load_overrides`: if it's not a list/tuple, log a warning and treat as `None`.
2. Inside the per-load loop: if an element is not a list/tuple, log a warning and fall back to the shared tariff for that load.

Three tests in `tests/test_optimization.py`: string outer value, list with string element, valid list-of-lists happy path. All three fail on unpatched code.

No schema, template, or JS changes.

## Verification

Full `test_optimization.py` suite (105 tests) passes locally. Pre-existing failures on upstream master are unrelated.

Followed AGENTS.md conventions (sections 3, 5, 6).

## Summary by Sourcery

Guard per-load cost forecast overrides against invalid string values and ensure safe fallback to shared tariffs.

Bug Fixes:
- Prevent crashes when cost_forecast_per_deferrable_load is provided as a string or contains non-list elements by falling back to shared tariffs.

Tests:
- Add regression tests covering string and mixed-type cost_forecast_per_deferrable_load configurations and a valid list-of-lists override path.

### DIFF
--- a/src/emhass/optimization.py
+++ b/src/emhass/optimization.py
@@ -2855,6 +2855,17 @@ class Optimization:
         # the user provides `cost_forecast_per_deferrable_load`, slot the override
         # array into the corresponding parameter.
         cost_per_load_overrides = self.optim_conf.get("cost_forecast_per_deferrable_load", None)
+        if cost_per_load_overrides is not None and not isinstance(
+            cost_per_load_overrides, (list, tuple)
+        ):
+            self.logger.warning(
+                "cost_forecast_per_deferrable_load is set but is %s, not a list (value: %r). "
+                "Treating as 'no override' for all loads. Use JSON null or an array of "
+                "per-load arrays (not the string \"null\").",
+                type(cost_per_load_overrides).__name__,
+                cost_per_load_overrides,
+            )
+            cost_per_load_overrides = None
         for k, param in enumerate(self.param_cost_per_load):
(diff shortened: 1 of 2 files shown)


davidusb-geek:
All tests checked fine but this is still in draft mode though

OptimalNothing90:
> All tests checked fine but this is still in draft mode though

@davidusb-geek The stage is now yours

davidusb-geek (approved):

PR #885, opened 2026-05-25
fix: LP retry silently skips battery/inverter constraints on cached-problem paths

hossamnagy:
## Problem

When `self.prob` is cached from a previous call, `inv_stress_conf` and `batt_stress_conf` remain `None` — they are only assigned inside the `if self.prob is None:` block. The LP relaxation retry path then guards on these objects:

```python
if inv_stress_conf:   # None on cache hit → False — hybrid inverter constraints SKIPPED
    self._add_hybrid_inverter_constraints(constraints_relaxed, inv_stress_conf)
if batt_stress_conf:  # None on cache hit → False — battery constraints SKIPPED
    self._add_battery_constraints(constraints_relaxed, batt_stress_conf)
```

After the first call, any LP retry runs without SOC limits, charge/discharge direction constraints, or DC-bus balance. The relaxed problem solves trivially with battery = 0 W everywhere, and that result is cached as `self.prob` — so subsequent runs return a flat zero battery schedule indefinitely.

## Reproduction

1. Set `set_use_battery: True` (or `inverter_is_hybrid: True`)
2. Trigger an optimisation that goes infeasible (tight SOC target, low-solar day, missing `inverter_ac_output_max`, etc.)
3. LP relaxation retry runs without battery constraints → battery = 0 W
4. Call optimisation again — `self.prob` is now the bad relaxed problem; returns 0 W every time until container restart

## Fix

Guard on the feature-flag config keys instead of the stress-config objects. Both `_add_battery_constraints` and `_add_hybrid_inverter_constraints` already handle a `None` stress-conf argument safely via their own internal `if batt_stress_conf and batt_stress_conf["active"]` checks.

## Impact

- **Battery users**: any system where an optimisation ever goes infeasible silently loses all battery scheduling on every subsequent run.
- **Hybrid inverter users**: the DC-bus balance constraint is missing from the relaxed problem, making the relaxed solution physically invalid.

## Summary by Sourcery

Bug Fixes:
- Prevent cached LP relaxation retries from silently omitting battery and hybrid inverter constraints by guarding on feature flags instead of stress configuration objects.

### DIFF
--- a/src/emhass/optimization.py
+++ b/src/emhass/optimization.py
@@ -3264,9 +3264,13 @@ class Optimization:
             # Re-apply main constraints
             self._add_main_power_balance_constraints(constraints_relaxed)
             # (Note: We reuse previous stress configs as they don't change with relaxation)
-            if inv_stress_conf:
+            # Guard on feature flags, not on the stress conf objects: stress conf is None
+            # on cached-problem retry paths (self.prob is not None), so guarding on the
+            # object itself would silently skip these constraints on every retry after the
+            # first call, leaving the relaxed problem without battery/inverter constraints.
+            if self.plant_conf.get("inverter_is_hybrid", False):
                 self._add_hybrid_inverter_constraints(constraints_relaxed, inv_stress_conf)
-            if batt_stress_conf:
+            if self.optim_conf.get("set_use_battery", False):
                 self._add_battery_constraints(constraints_relaxed, batt_stress_conf)
 
             if self.plant_conf["compute_curtailment"]:


hossamnagy:
## What triggered this

This was found on a real FoxESS H3 hybrid inverter setup (41.9 kWh battery, 10 kW inverter).

The system had `inverter_is_hybrid: True` in the config but was missing `inverter_ac_output_max`. The fallback in `_add_hybrid_inverter_constraints` sets `p_nom_inverter_output = 0`, which forces both `p_dc_ac_max` and `p_ac_dc_max` to zero. The DC-bus balance constraint then becomes:

```
p_pv + p_sto_pos + p_sto_neg = 0
```

At night (`p_pv = 0`) this locks the battery to 0 W. With a SOC final target that required ~11,740 Wh of net charging but only ~9,017 Wh available from daytime solar, the MILP was presolve-infeasible on every single run.

That infeasibility triggered the LP retry path — which is where this bug bit hard. The LP retry rebuilt the problem without battery constraints (because `batt_stress_conf` was `None` on every cached-problem call after the first). The relaxed problem solved instantly with battery = 0 W everywhere, got cached as `self.prob`, and every subsequent optimisation call returned a flat zero battery schedule without any error or warning.

The symptom from the user side was just: *battery scheduling always shows 0 W, `optim_status = Optimal`*. No errors in the logs. It looked like a working optimisation.

The root config issue is fixed by adding `inverter_ac_output_max: 10000` (the inverter's actual AC capacity in watts). But this bug means any system that hits infeasibility for *any* reason — tight SOC targets, low-solar days, misconfigured constraints — silently loses battery scheduling until the container restarts and `self.prob` is rebuilt from scratch.

davidusb-geek (approved):

Issue #839, opened 2026-05-11
[Feature request] "Optional" loads that are only scheduled below a certain cost threshold


lutorm:
Currently, my impression is that EMHASS considers a deferrable load "mandatory", meaning it must be scheduled during the optimization interval. 

I'm trying to figure out how to more efficiently use my heat pump to heat water. It will normally do its thing and keep the hot water above some threshold using some internal black box model of when it thinks we use hot water. This works pretty well in that we don't run out of water and at the same time we don't need to keep the reservoir at a high temperature all the time with corresponding high heat losses and heat pump inefficiency.

The heat pump has a "SG READY" input whereby you can signal 3 "abnormal" conditions:

- **low price**, in which case it will start using the compressor to heat the water to a higher temperature using a higher power than normal. Both these conditions mean the efficiency is less than its normal operating mode, but if the energy price is _sufficiently_ lower, it will be a net gain. 
- **overproduction**, in which case it will run maximum power using the direct immersion heater up to the maximum allowed temperature. This is intended to be used if you e.g. have negative selling prices and need to avoid feeding energy into the grid. It's of course very inefficient such that if you can do anything else useful with the energy, that's probably a better idea.
- **high price**, in which case it will pause heating.

It's mainly the "low price" mode that I want to make use of. But the current deferrable load model does not work well for this.

What would be needed is to define an "alternative cost" to running the deferrable load. For example, we assume the alternative to using X kWh during one hour is to have the heat pump do its thing, running approximately continuously during 24h. The "optional load" should then only be scheduled if the cost of running the load is some threshold lower (to account for worse efficiency) than the alternative, which is smearing the energy out over 24h using the average energy cost.

davidusb-geek:
Yes heat pumps are a very specific type of deferrables that might need different treatment.
We also might need a better modeling specific for heat pumps.

For now, a "quick win" would be to add a `cost_threshold` parameter to deferrable loads.
i need to think about the complexity of this and the new parameters needed to control this behavior

lutorm:
I've been thinking about how a cost threshold for the loads would work, because it's not entirely obvious to me how to actually count the "cost" of running a load in all situations.

- If we're importing from the grid, then clearly the cost of supplying the load should be lower than the threshold.
- On the other hand, if we can power the load from PV while still exporting to the grid, is the cost zero? No, I'd say the cost is actually the lost revenue from the export we would have gotten without the load.
[...]

PR #886, opened 2026-05-25
Add optional max cost threshold for deferred loads.

lutorm:
As discussed in https://github.com/davidusb-geek/emhass/issues/839, it would be useful to be able to set a max cost threshold such that if the cost of scheduling a load exceeds the specified cost, it should not be scheduled.

I realized that this can be implemented relatively easily through two changes: 

1. Making the scheduling of a load optional for the optimizer.
2. Giving the optimizer a reward equal to the threshold value if a load is scheduled.

The result of this is that as long as the cost of the optimal solution with the load scheduled does not exceed that of the solution without the load by more than the max cost, the optimizer can win by scheduling the load.

With the help of copilot, I came up with the following implementation. It works and has a test.

## Summary by Sourcery

Introduce an optional maximum cost threshold for deferrable loads so the optimizer can skip scheduling loads that are too expensive while rewarding schedules within the configured limit.

New Features:
- Allow configuration of per-deferrable-load maximum cost thresholds that make scheduling optional when the cost exceeds the specified limit.
- Add optimizer reward terms tied to deferrable load scheduling decisions so loads are only run when their cost is below the configured maximum.

Tests:
- Add tests verifying that standard and sequence deferrable loads are only scheduled when their cost is below the configured maximum cost thresholds.


lutorm:
The code quality scan is failing on "Error: Process completed with exit code 1", giving no further info. Is this an infra failure or is it actually complaining about something in the code?

davidusb-geek:
> The code quality scan is failing on "Error: Process completed with exit code 1", giving no further info. Is this an infra failure or is it actually complaining about something in the code?

That's a bit tricky but you need to read the log lines above that final sentence.
So it is failing because there are some Ruff formatting issues and the logs give you what is the expected format.
What I do is that I setup VSCode to auto format my code when saving a file CTR+S.
This way the code quality is always happy no matter what.
If you can update or I can also merge and then I do a final check of these myself before the release

OptimalNothing90:
Following up on the null-handling guards from #878 and #879. I checked whether the same stringly-typed / null-default class reaches `deferrable_load_max_cost`, and it does. The solver logic itself looks solid (Big-M formulation, sequence path, and reward term all check out), but two inputs crash at `optimization.py:~2245`:

```python
max_cost = ["0.6", "0.65"]   # HA add-on UI stringifies float arrays
has_max_cost = max_cost[0] > 0
# TypeError: '>' not supported between instances of 'str' and 'int'

max_cost = [None, 0.5]       # partial set-config payload
has_max_cost = max_cost[0] > 0
# TypeError: '>' not supported between instances of 'NoneType' and 'int'
```

Every adjacent per-load float array (`set_deferrable_startup_penalty` and friends) passes through `check_def_loads` in `treat_runtimeparams` first, which normalises exactly these cases. `deferrable_load_max_cost` is missing from that block. One call covers both:

```python
params["optim_conf"]["deferrable_load_max_cost"] = check_def_loads(
    num_def_loads, params["optim_conf"], 0.0, "deferrable_load_max_cost", logger,
)
```

Two small `param_definitions.json` notes while I was in there: the `unit` reads `¤/kWh` but the reward is added as unscaled currency, so `¤` is more accurate; and the "Does not work for sequence loads" line is stale — the sequence path is implemented and `test_sequence_load_max_cost` passes.

Happy to put up the `check_def_loads` call plus a string-valued test if that helps.

lutorm:
@OptimalNothing90 Thanks for finding those, I was not aware of the check_def_loads function.

But the disable path is setting the cost to 0, not None, and the defaults are 0, so I don't understand where any Nones are entering the path?

OptimalNothing90:
You're right it's not the disable path. The None comes from partial set-config payloads: configure one load, leave the other null, and the array arrives as `[null, 0.5]`. That's the case `check_def_loads` already normalises for the other per-load arrays (its comment even mentions `[null, 0]`); `deferrable_load_max_cost` is just the one that skips it. Strings arrive the same way via HA service calls. Both land on `max_cost[k] > 0` and throw, so the same `check_def_loads` call the siblings use sorts it.

Separately, Sourcery's second point is a real one too. With `max_cost` set and `operating_hours` left at the default 0, the energy constraint forces `total_energy = 0` while `load_is_scheduled = 1`, so the solver banks the reward with `P_deferrable = [0,0,0,0]` (phantom schedule). `param_energy_active[k] * load_is_scheduled` keeps it relaxed when the load has no energy target.

lutorm:
@OptimalNothing90 so if I handle the config with check_def_loads, I can get rid of the current length check
```
        max_cost = self.optim_conf.get("deferrable_load_max_cost",
                                       [0.0] * self.optim_conf["number_of_deferrable_loads"])
        max_cost = max_cost + \
            [0.0] * (self.optim_conf["number_of_deferrable_loads"] - len(max_cost))
```
?

OptimalNothing90:
Yes, check_def_loads already pads and should handle None.

davidusb-geek (approved):

PR #893, opened 2026-05-27
command_line: stop the OptimizationCache invalidating every MPC tick

### LINKED ISSUE #892: OptimizationCache invalidates on every MPC tick when forecast-model actions run
## Symptom

On a steady-state setup where MILP structure does not change tick-to-tick,
`OptimizationCache` misses on every `/action/naive-mpc-optim` call:

    INFO in command_line: OptimizationCache MISS: Config changed - ...
    INFO in optimization: Building CVXPY problem structure...

That forces a full CVXPY canonicalisation + cold MILP solve on every
tick, even though the upstream-claimed warm-cache benefit should drop
that significantly. On a 576-timestep horizon with 5-min cadence and 3
deferrable loads + 1 battery, every tick paid ~1.7 s wall, all of it
spent rebuilding the problem.

### 1. `operating_timesteps_of_each_deferrable_load` is in the cache key

The field is part of `optim_conf_structural_hash` via
`config_hash(optim_conf, exclude_keys=optim_conf_runtime_keys)` in
`command_line.py::OptimizationCache.get`. But the field is
*parameterised* via `param_target_energy` and
`param_required_timesteps` in `optimization.py:2980-3007` — changing
its value at runtime doesn't need a rebuild, just
`param.value = …`.

The exclude-set already covers
`operating_hours_of_each_deferrable_load` and the
`start_/end_timesteps_of_each_deferrable_load` pair, just not
`operating_timesteps_of_each_deferrable_load`. Looks like an oversight.

Result: small tick-to-tick shifts (e.g. hot-water-hours-needed
translating to 37 vs 38 timesteps) silently invalidate the cache.

### 2. ML-forecast actions go through the cache path

`set_input_data_dict` in `command_line.py` currently only excludes
`["publish-data", "export-influxdb-to-csv"]` from the optimization-
construction path. The `forecast-model-{fit,predict,tune}` and
`weather-forecast-cache` actions still build an `Optimization` and call
`OptimizationCache.put`, even though they don't actually need one.

Those actions don't carry `nominal_power_of_deferrable_loads` in their
runtime params, so they fall back to `config_defaults.json`. When a
setup runs `forecast-model-predict` (or any of the others) on a
recurring schedule alongside `naive-mpc-optim`, the predict's
default-config-derived key gets cached, and the subsequent MPC call
sees a mismatch on `nominal_power_of_deferrable_loads` and rebuilds.

## Proposed fix (PR follows)

`src/emhass/command_line.py`:

1. Add `operating_timesteps_of_each_deferrable_load` to
   `optim_conf_runtime_keys`.

2. Extend `actions_without_optimization` from
   `["publish-data", "export-influxdb-to-csv"]` to also include
   `["forecast-model-fit", "forecast-model-predict",
   "forecast-model-tune", "weather-forecast-cache"]`.

## Measured impact

24-core x86_64 Linux, EMHASS v0.17.5 in Docker, 576-timestep MPC at
5-min cadence, 3 deferrable loads + battery:

| state | log line | MPC wall | optim_solve |
| --- | --- | ---: | ---: |
| before, every tick | `Building CVXPY problem structure…` | 1.7–1.9 s | 1.5–1.6 s |
| after, first tick (cold) | `OptimizationCache MISS: Empty cache` | 1.8 s | 1.5 s |
| after, subsequent ticks (warm) | `OptimizationCache HIT: … warm-start enabled` | 0.8–1.3 s | 0.5–1.0 s |

~30–40 % wall-clock reduction across steady-state ticks. No behaviour
change — the cache just respects the existing parameterisation
conventions documented in `optimization.py`.

## Verification

- Regression test for the `operating_timesteps` case included in the PR
  (extends the existing `TestOptimizationCache` class with the same
  pattern as `test_cache_hit_operating_hours_changed`).
- Corpus replay of 36 synthetic MPC payloads (clear day, cloudy,
  evening peak, overnight, heavy-deferrable-window) before and after:
  `opt_res` schema identical, tolerance bands (±1 W power, ±0.001 SOC,
  ±0.5 % objective) hold on every scenario.

BrettLynch123:
## Motivation

See #892 for context and measured impact (~30–40 % MPC wall
reduction on the affected setup; cache transitions from MISS-every-tick
to HIT-every-tick once warm).

## Change

Two cache-key correctness fixes in `src/emhass/command_line.py`:

1. Add `operating_timesteps_of_each_deferrable_load` to
   `optim_conf_runtime_keys`. The field is parameterised in
   `optimization.py` via `param_target_energy` and
   `param_required_timesteps` (~line 2980-3007), so a tick-to-tick
   value change doesn't require a rebuild. The exclude-set already
   covers the related `operating_hours_of_each_deferrable_load`.

2. Extend `actions_without_optimization` to include
   `forecast-model-{fit,predict,tune}` and `weather-forecast-cache`.
   None of these construct or solve an `Optimization`, but they were
   still routing through `OptimizationCache.put` with a stale
   `nominal_power_of_deferrable_loads` from `config_defaults.json`,
   polluting the cache key for the subsequent MPC call.

Includes a regression test in
`tests/test_command_line_utils.py::test_cache_hit_operating_timesteps_changed`
matching the pattern of the existing
`test_cache_hit_operating_hours_changed`.

## Risk + rollback

- Both changes are additive (set-element additions, no removals)
- Public API: unchanged
- Schema impact on `opt_res`: unchanged
- New config keys: none
- Rollback: revert the single commit
- Reasoning preserved in code comments inline with the changes

## References

Closes #892

## Summary by Sourcery

Adjust optimization cache behavior to avoid unnecessary cache invalidations between MPC ticks and ensure non-optimization actions bypass the cache.

Bug Fixes:
- Treat changes in operating timesteps for deferrable loads as runtime-parameterized so they no longer invalidate the optimization cache.
- Exclude forecast and weather cache actions from the optimization path to prevent stale configuration from polluting the cache key and causing cache misses.

Tests:
- Add a regression test to verify that modifying operating timesteps for deferrable loads still results in an optimization cache hit.

### DIFF
--- a/src/emhass/command_line.py
+++ b/src/emhass/command_line.py
@@ -260,6 +260,7 @@ class OptimizationCache:
         optim_conf_runtime_keys = {
             # Parameterized via CVXPY Parameters
             "operating_hours_of_each_deferrable_load",
+            "operating_timesteps_of_each_deferrable_load",
             "start_timesteps_of_each_deferrable_load",
             "end_timesteps_of_each_deferrable_load",
             "def_current_state",
@@ -1115,10 +1116,19 @@ async def set_input_data_dict(
     if isinstance(params, str):
         params = dict(orjson.loads(params))
     costfun = optim_conf.get("costfun", costfun)
-    # Actions that don't require building an Optimization object
-    # publish-data only reads saved results and posts to Home Assistant
-    actions_without_optimization = ["publish-data", "export-influxdb-to-csv"]
-    if set_type in actions_without_optimization:
+    # Two-tier guard:
+    #   - actions_without_fcst_or_opt: read saved results only; build neither.
+    #   - actions_skip_optim_cache: need a Forecast object but no Optimization.
+    #     Keeping these out of the OptimizationCache path stops them poisoning
+    #     the cache key with config-default values that a subsequent
+    #     naive-mpc-optim call would then miss against.
+    actions_without_fcst_or_opt = ["publish-data", "export-influxdb-to-csv"]
+    actions_skip_optim_cache = [
+        "forecast-model-fit",
+        "forecast-model-predict",
+        "forecast-model-tune",
+    ]
+    if set_type in actions_without_fcst_or_opt:
         fcst = None
         opt = None
         logger.debug(f"Skipping Optimization creation for action: {set_type}")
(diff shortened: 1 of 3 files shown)


davidusb-geek:
@BrettLynch123 This one needs attention on those failing tests

BrettLynch123:
@davidusb-geek tests are now passing — force-pushed the fix.

Root cause: the over-eager `actions_without_optimization` guard set `fcst=None` for `forecast-model-{fit,predict,tune}`, which broke `test_get_load_forecast_mlforecaster` (the test accesses `input_data_dict["fcst"]` directly). Split into a two-tier guard so those actions build `fcst` but still skip the OptimizationCache path.

The second commit reformats one line in `optimization.py:2992` (PR #897's log call) that the latest ruff release flags as malformed — same failure is currently red on master too.

The two remaining Ubuntu failures (`test_get_forecasts_with_lists_special_case`, `test_get_weather_forecast_openmeteo_method_mock`) are `api.open-meteo.com` returning 502 / Connection refused — not from this PR. They passed on Windows + macOS this run. A rerun should clear them.

davidusb-geek (approved):

Issue #899, opened 2026-05-27
Deferrable pinning block can produce an infeasible MILP after #897 (single-constant + currently-running + window outside horizon); relates to #873


OptimalNothing90:
#897 correctly deactivates loads whose window is entirely outside the horizon. One sub-case slips through and produces the infeasible MILP / relaxed-LP fallback that #897 set out to avoid.

Trigger: a single-constant deferrable with `def_current_state=True` (running at the end of the previous MPC horizon) and a window outside `[0, n]`. The energy and timestep guards got the `k not in window_empty_loads` exclusion; the pinning block (`optimization.py:~3079`) did not. `constraint_active` is still True locally, so the pinning block fires:

```
param_running_lb[k][:pinned_steps] = 1.0
param_already_running_sc[k] = 1.0
```

while the load-active loop correctly sets `param_load_active[k] = 0.0`. The static constraints then conflict:

```
p_def_bin2[k] <= 0     (load_active)
p_def_bin2[k] >= 1     (running_lb)
sum(p_def_start[k]) == -1
```

Infeasible, optimization fails for that tick. Repro confirmed. Real scenario: a dishwasher scheduled for tonight (start=600, end=800) on a 576-step horizon, still running from the prior cycle.

One-line fix, mirroring the guard already on the energy/timestep blocks:

```python
if (is_single_const and current_state and constraint_active
        and required_timesteps > 0
        and k not in window_empty_loads):   # add this
```

The existing `else` branch already zeros both params, so this routes the case there.

This is also the second of the two paths in #873 (the "starts immediately" report): #897 covers the horizon-overflow path, this covers the currently-running single-constant path. Happy to PR the one-liner plus a test, or leave it to @BrettLynch123 if preferred. Your call, @davidusb-geek.

PR #907, opened 2026-05-27
fix: repair config page rendering for null-default params (#880, #904)

OptimalNothing90:
## Summary

Fixes the config UI crash introduced in v0.17.4/v0.17.5, reported in #880 and #904.

### Root causes

**#880 — config page dead after load**

`window.onload` → `loadConfigurationListView` → `minusElements` crash path:

1. `buildParamElement` for `heat_topology` (input=`"object"`, value=`{"sources":[]}`) entered the nested-object render path and returned `"</br>"` — no `<input>` element.
2. `minusElements("heat_topology")` found `length == 0` and entered the error branch, then referenced `parameter_definition_name` — a variable that exists in `buildParamContainers` but **not** in `minusElements` — causing an uncaught `ReferenceError`.
3. The uncaught error aborted the `async window.onload` promise. All subsequent code (battery toggle, JSON toggle, save button event listeners) was never reached → whole UI dead.

**#904 — heat_topology saved as string `"null"`**

`buildParamElement` for absent `heat_topology`: `placeholder = JSON.stringify(null)` → `"null"` (string). The save path stored this as the string `"null"`, causing a per-request backend warning in `utils.py`.

### Changes — `configuration_script.js`

| Hunk | Change | Fixes |
|------|--------|-------|
| `minusElements` L556, L563 | `parameter_definition_name` → `param` | #880 crash (core fix) |
| `minusElements` L565 | Add `return 1` after the length==0 log | #880 crash (prevents TypeError fallthrough) |
| `buildParamElement` | Render `input === "object"` as a single JSON text box before the nested-object path | #880 UX + #904 display |
| `buildParamElement` | `placeholder = "" ` (not `"null"`) when `default_value === null` | #904 display |
| `saveConfiguration` | JSON.parse object-type inputs; treat `""` / `"null"` → `null` | #904 save |

### Changes — `configuration.html`

Add `?version=1` cache-buster to the `configuration_script.js` `<script>` tag, matching the existing `?version=2` on `style.css`. Forces a cache miss for users who cached the broken file. (A dynamic version via `importlib.metadata` + template context is a worthwhile follow-up but requires touching `web_server.py`.)

### Changes — `tests/test_schema_contract.py`

Four new regression tests (two static, two Node.js replay):

- `test_minus_elements_no_undefined_variable` — asserts `parameter_definition_name` absent from `minusElements` body
- `test_minus_elements_does_not_crash_on_zero_inputs` — Node replay: 0-input param, asserts no `ReferenceError`
[...]

PR #910, opened 2026-05-28
fix: deferrable single-constant pin firing on stringly-typed def_current_state (#873)

### LINKED ISSUE #873: Deferrable load not correct planned v0.17.3/v0.17.4 and v0.17.5
When planning deferrable load it get's planned too soon:

```
 **'operating_timesteps_of_each_deferrable_load': [20, 0, 0, 0]**, 'treat_deferrable_load_as_semi_cont': ['False', 'False', 'False', 'False'], 'set_deferrable_load_single_constant': ['True', 'True', 'True', 'True'], **'start_timesteps_of_each_deferrable_load': [37, 0, 0, 0]**, **'end_timesteps_of_each_deferrable_load': [57, 1, 1, 0]**,
```

Below the EMHASS planning. Deferrable_0 immediately started

Image

OptimalNothing90:
## Problem
Since v0.17.3, single-constant deferrable loads with a future window are scheduled to start at t=0 instead of respecting `start_timesteps_of_each_deferrable_load`. Reported by @RikBast; confirmed across v0.17.3 / v0.17.4 / v0.17.5 with `prediction_horizon=192`, `start_timesteps=37`, `end_timesteps=57` (window fully inside horizon, so horizon-overflow is not the cause).

## Root cause
`def_current_state` arrives as a list of strings (`['False','False','False','False']`). `treat_runtimeparams` in `utils.py` casts with `bool()`, but `bool('False') = True` in Python (non-empty strings are truthy). The v0.17.3 "currently-running single-constant load → pin to start of horizon" path then fires even when the load is not running. The pin sets `window_mask[:pinned_steps] = 1.0`, opening the window at t=0. Under MPC rolling-window the bug self-reinforces (spurious pin → load actually starts → next tick's state IS True → pin re-fires legitimately). Same class as #876 / #878 / #879; `def_current_state` slipped through.

`set_deferrable_load_single_constant` has the same exposure (arrives as strings via the generic `associations.csv` path with no coercion); fixed in the same commit.

A third class of input was also found during review: `def_current_state=[null, null]` (HA sends JSON null when the sensor is unavailable). `ast.literal_eval('None')` returns Python `None` without raising — so `_cast_bool(None)` previously propagated `None` into `def_current_state`. The optimizer's `_update_def_current_state_params` has an `isinstance(state, bool|int|float)` guard that raises `ValueError` on `None`, crashing the optimization. Fixed by adding an explicit `if value is None: return False` guard at the top of `_cast_bool`.

## Fix — three commits
**Commit 1** (`fix:`) extends the existing `_cast_bool` helper (already present in `treat_runtimeparams`) to `def_current_state` and `set_deferrable_load_single_constant`. Replaces `bool()` (wrong for string booleans) with `_cast_bool()` which uses `ast.literal_eval` and correctly maps `'False' → False`. No layer migration, no architectural refactor — the class-level coercion strategy is deliberately deferred to a follow-up after #901 lands.

**Commit 2** (`test:`) is a regression guard for the window-outside-horizon deactivation behaviour fixed in commit `eea5d25` (upstream/master, May 27 by @BrettLynch123). That production fix was already merged before this PR was opened; no existing test covered the `window_empty_loads` path. `test_window_entirely_outside_horizon_zeros_mask` locks it in.

**Commit 3** (`fix:`) adds `if value is None: return False` guard to `_cast_bool`, and adds `test_def_current_state_null_elements_do_not_crash_optimizer` (Test D) covering the `[null, null]` crash path found during review.

## Tests (TDD red-first for Commits 1 and 3; regression guard for Commit 2)
- `test_def_current_state_stringly_false_does_not_pin_window_open` — single-shot regression: load must not run before slot 10 when `def_current_state=['False']` (string). **Was RED on unfixed code** (`12000 W` before window); GREEN after fix.
- `test_def_current_state_stringly_false_no_spurious_pin_across_mpc_ticks` — MPC rolling KEYSTONE: checks per-tick command across 3 ticks. **Was RED on unfixed code** (`def_current_state[0]` was `True` after wrong coercion); GREEN after fix.
- `test_window_entirely_outside_horizon_zeros_mask` — regression guard for Commit 2's already-merged fix. GREEN on current master; added because no prior test covered that path.
- `test_def_current_state_null_elements_do_not_crash_optimizer` — null-path crash guard: `[None, None]` must coerce to `[False, False]` and not raise `ValueError` in the optimizer. **Was RED on tip bb152f7** (`got [None, None]`); GREEN after Commit 3.

## Out of scope / follow-up
- Class-level coercion architecture (centralize in `build_params`): separate follow-up issue to evaluate after #901 lands.
- #899 (deferrable pinning block → infeasible MILP) — same area, different bug.

Fixes #873

## Summary by Sourcery

Fix boolean coercion of deferrable load configuration to prevent spurious pinning and crashes, and add resilience tests around stringly-typed and out-of-horizon parameters.

Bug Fixes:
- Correct def_current_state coercion to use a safe boolean parser so string values like 'False' do not incorrectly evaluate as True and spuriously pin single-constant deferrable loads.
- Apply the same safe boolean coercion to set_deferrable_load_single_constant to avoid misinterpreting string configuration values.
- Treat None values in boolean-like configuration (e.g. def_current_state elements) as False to prevent ValueError crashes during optimization.

Tests:
- Add regression tests ensuring stringly-typed def_current_state values do not cause early pinning for single-constant loads, including across multiple MPC ticks.
- Add a regression test verifying that windows configured entirely outside the optimization horizon leave deferrable load masks zeroed so loads are deactivated.
- Add a regression test confirming def_current_state entries of None coerce to False and do not crash the optimizer.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+LOCATION.md
 # ---> Python
 test.py
 # Legacy config yaml file
--- a/src/emhass/utils.py
+++ b/src/emhass/utils.py
@@ -1686,6 +1686,7 @@ async def treat_runtimeparams(
         # Define Helper Functions
         def _cast_bool(value):
             """Helper to cast string inputs to boolean safely."""
+            # ast.literal_eval('None') returns None without raising — explicit guard needed.
             if value is None:
                 return False
             try:
--- a/src/emhass/utils.py
+++ b/src/emhass/utils.py
@@ -1847,11 +1847,24 @@ async def treat_runtimeparams(
                     dcs = orjson.loads(dcs)
                 except Exception:
                     logger.warning(f"Could not parse def_current_state string: {dcs}")
-            # Check it's iterable before casting to bool
+            # Use _cast_bool (not bool()) so string 'False' → False, not True.
+            # bool('False') = True because non-empty strings are truthy in Python.
             if isinstance(dcs, list):
-                params["optim_conf"]["def_current_state"] = [bool(s) for s in dcs]
+                params["optim_conf"]["def_current_state"] = [_cast_bool(s) for s in dcs]
             else:
-                params["optim_conf"]["def_current_state"] = [bool(dcs)]
+                params["optim_conf"]["def_current_state"] = [_cast_bool(dcs)]
+
+        # set_deferrable_load_single_constant arrives via the generic associations.csv
+        # path as-is (may be a list of strings from runtimeparams JSON).  Apply the
+        # same _cast_bool coercion so 'False' → False, not True (#873, mirrors #876).
+        if "set_deferrable_load_single_constant" in runtimeparams.keys():
+            sdlsc = runtimeparams["set_deferrable_load_single_constant"]
+            if isinstance(sdlsc, list):
+                params["optim_conf"]["set_deferrable_load_single_constant"] = [
+                    _cast_bool(s) for s in sdlsc
+                ]
+            else:
+                params["optim_conf"]["set_deferrable_load_single_constant"] = [_cast_bool(sdlsc)]
 
         # Treat retrieve data from Home Assistant (retrieve_hass_conf) configuration parameters passed at runtime
         # Secrets passed at runtime
--- a/src/emhass/utils.py
+++ b/src/emhass/utils.py
@@ -1855,7 +1855,8 @@ async def treat_runtimeparams(
             if isinstance(dcs, list):
                 params["optim_conf"]["def_current_state"] = [_cast_bool(s) for s in dcs]
             else:
-                params["optim_conf"]["def_current_state"] = [_cast_bool(dcs)]
+                n_loads = len(params["optim_conf"]["nominal_power_of_deferrable_loads"])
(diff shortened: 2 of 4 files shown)


OptimalNothing90:
@davidusb-geek I'm not seeing why codecov blocks. Can you first merge PR #906 #907 #901 & #903?
After that I'm going to rebase and then see what it the real case

davidusb-geek:
> @davidusb-geek I'm not seeing why codecov blocks. Can you first merge PR #906 #907 #901 & #903? After that I'm going to rebase and then see what it the real case

Done! MErged and yes there are some conflicts

OptimalNothing90:
> > @davidusb-geek I'm not seeing why codecov blocks. Can you first merge PR #906 #907 #901 & #903? After that I'm going to rebase and then see what it the real case
> 
> Done! MErged and yes there are some conflicts

thanks for the quick reaktion. Yes, this makes sense, but should be a quick one

OptimalNothing90:
@davidusb-geek resolved, can be merged now

davidusb-geek (approved):

PR #918, opened 2026-05-31
fix: restore delete-on-stale behaviour for Open-Meteo forecast cache

### LINKED ISSUE #917: [BUG] get_cached_forecast_data zeroes irradiance for Open-Meteo (regression v0.17.3)
## Summary

Since **v0.17.3**, when `weather_forecast_data.pkl` no longer covers the requested forecast window, `get_cached_forecast_data` silently returns **zero-filled irradiance data** instead of deleting the stale cache and triggering a fresh Open-Meteo fetch.  
This causes all PV generation forecasts to be **flat/zero** indefinitely — without any error.

---

## Affected versions

- ✅ v0.17.2 — working correctly  
- ❌ v0.17.3 — regression introduced  
- ❌ v0.17.4 — still broken  
- ❌ v0.17.5 — still broken  

---

## How to reproduce

1. Configure EMHASS with `weather_forecast_method: open-meteo`.
2. Let the system run until `weather_forecast_data.pkl` is written.
3. Wait ~24 h (the pkl's index no longer covers the new rolling forecast window).
4. Trigger any optimization or `weather-forecast-cache` action.
5. **Observed**: all `ghi`/`dni`/`dhi` columns are `0.0`, PV forecast is flat.  
   **Expected**: stale pkl is deleted and a fresh request is sent to Open-Meteo.

---

### The breaking change in `get_cached_forecast_data` (forecast.py ~line 1829)

**v0.17.2 (correct for Open-Meteo):**

```python
else:
    self.logger.error("Unable to obtain cached forecast data within the requested timeframe range.")
    self.logger.warning("Removing old forecast cache file. Next optimization will pull data from forecast API...")
    os.remove(w_forecast_cache_path)   # ← deletes stale pkl
    return False                        # ← triggers fresh API fetch
```

**v0.17.3+ (broken for Open-Meteo):**

```python
else:
    # Serve best-effort stale cache: reindex to requested dates,
    # interpolate gaps, and zero-fill edges.  This is far better
    # than deleting the cache and burning Solcast API quota.
    ...
    if irradiance_cols:
        data[irradiance_cols] = data[irradiance_cols].fillna(0.0)  # ← zeroes irradiance forever
    data = data.fillna(0.0)
```

### Why this is correct for Solcast but wrong for Open-Meteo

The v0.17.3 change was motivated by **Solcast rate-limit protection** (8 calls/day quota), which is legitimate.  
However, `get_cached_forecast_data` is a **shared method** called by both `_get_weather_open_meteo` and `_get_weather_solcast` — and the new behaviour is applied unconditionally regardless of `weather_forecast_method`.

For **Open-Meteo** (unlimited, free API):
- The coverage gap is **structural**: Open-Meteo returns calendar-aligned data (00:00 → 23:45 of day N), while EMHASS requests a *rolling* window starting *now*. The gap equals the time elapsed since midnight — e.g. ~10 h at 10:00 local time.
- This gap exists for **any** value of `delta_forecast_daily`.
- Every 24 h the cache will be stale and the zero-fill fallback triggers.
- Because `_get_weather_open_meteo` only hits the API when the pkl does **not** exist (`if not os.path.isfile(w_forecast_cache_path)`), the stale pkl is **never replaced** once the fallback keeps it in place.

### Structural gap visualised

```
Open-Meteo response:  |--day 0--|--day 1--|....|--day N--|
                       00:00                             23:45

EMHASS rolling window:     |----day 1----|....|--day N--|
                            10:00 ↑ gap ↑
                              (time elapsed today)
```

---

## Workaround (manual, until fix is released)

```bash
docker exec $(docker ps --filter "name=emhass" -q) rm /data/weather_forecast_data.pkl
curl -X POST -H 'Content-Type: application/json' -d '{}' http://127.0.0.1:5000/action/weather-forecast-cache
```

---

## Proposed fix

Branch on `self.weather_forecast_method` **inside `get_cached_forecast_data`** — no new parameters, no changes to call sites. Open-Meteo deletes and refetches (v0.17.2 behaviour); Solcast keeps the existing stale-data protection.

```python
async def get_cached_forecast_data(self, w_forecast_cache_path) -> pd.DataFrame:
    ...
    if self.forecast_dates[0] in data.index and self.forecast_dates[-1] in data.index:
        data = data.loc[self.forecast_dates[0] : self.forecast_dates[-1]]
        self.logger.info("Retrieved forecast data from the previously saved cache.")
    else:
        if self.weather_forecast_method == "open-meteo":
            # Open-Meteo has no rate limits: delete the stale cache so
            # the next call fetches fresh data from the API.
            self.logger.warning(
                "Cache does not fully cover the requested timeframe. "
                "Removing stale Open-Meteo cache; fresh data will be fetched from the API."
            )
            os.remove(w_forecast_cache_path)
            return False
        else:
            # Solcast and other rate-limited APIs: serve best-effort
            # stale data to avoid burning daily API quota.
            self.logger.warning(
                "Cache does not fully cover the requested timeframe. "
                "Serving best-effort stale data (reindexed + zero-filled). "
                "Cache preserved to protect API rate limits."
            )
            combined_index = data.index.union(self.forecast_dates).sort_values()
            data = data.reindex(combined_index)
            data.interpolate(method="time", inplace=True)
            data = data.reindex(self.forecast_dates)
            irradiance_cols = [c for c in ["ghi", "dni", "dhi"] if c in data.columns]
            other_cols = [c for c in data.columns if c not in irradiance_cols]
            if other_cols:
                data[other_cols] = data[other_cols].ffill().bfill()
            if irradiance_cols:
                data[irradiance_cols] = data[irradiance_cols].fillna(0.0)
            data = data.fillna(0.0)
    return data
```

`_get_weather_open_meteo` already handles a `False` return correctly — when `get_cached_forecast_data` deletes the pkl and returns `False`, the caller retries and takes the `if not os.path.isfile(...)` branch which fetches from the API.

> **Note:** `_get_weather_solcast` calls `get_cached_forecast_data` as well and will correctly keep the existing stale-data behaviour because `self.weather_forecast_method` will be `"solcast"`.

---

## Impact

| `weather_forecast_method` | v0.17.2 | v0.17.3–v0.17.5 | With proposed fix |
|---|---|---|---|
| `open-meteo` | ✅ deletes stale pkl → refetches | ❌ zero-fills irradiance indefinitely | ✅ deletes stale pkl → refetches |
| `solcast` | ⚠️ deletes pkl → burns quota | ✅ serves stale data, protects quota | ✅ serves stale data, protects quota |

---

## Environment

- EMHASS version: 0.17.5 (Home Assistant add-on)
- `weather_forecast_method: open-meteo`
- `delta_forecast_daily: 8`
- `open_meteo_cache_max_age: 30`
- Home Assistant OS

tomvanacker85:
## Summary

Fixes a regression introduced in v0.17.3 where `get_cached_forecast_data()`
started zero-filling irradiance columns (GHI/DNI/DHI) when the cached
forecast did not cover the full requested timeframe.

This is correct behaviour for **rate-limited APIs** (e.g. Solcast) where
burning an extra API call just to refresh the cache is undesirable.
However, for **Open-Meteo** — which has no rate limits — it causes EMHASS
to optimise against a flat (zero) solar forecast, producing completely
wrong charge/discharge schedules.

## Changes

**`get_cached_forecast_data()`** — branch on `self.weather_forecast_method`:
- `"open-meteo"`: delete the stale pickle and return `False`, preserving
  the pre-v0.17.3 behaviour (fresh fetch on next call).
- all other methods: keep the v0.17.3 reindex + zero-fill path to protect
  rate-limited API quota.

**`_get_weather_open_meteo()`** — handle the new `False` return value from
`get_cached_forecast_data()` by immediately retrying the call, which will
now fetch fresh data from the Open-Meteo API since the stale pickle was
removed.

## How to reproduce

1. Let EMHASS run with `weather_forecast_method: open-meteo` and caching
   enabled until the cached JSON/pickle drifts out of the requested window
   (e.g. after midnight without an explicit cache refresh).
2. Call `/action/weather-forecast-cache`.
3. Observe all GHI / DNI / DHI values are 0.0 in the logged forecast.

With this fix, step 3 shows "Removing stale Open-Meteo cache; fresh data
will be fetched from the API" and the API is called immediately.

Closes #917

## Summary by Sourcery

Adjust forecast cache handling so Open-Meteo uses cache invalidation with refetch while rate-limited providers retain best-effort stale data, and extend tests to cover the distinct behaviours.

Bug Fixes:
- Restore delete-on-stale behaviour for Open-Meteo forecast cache to avoid zero-filled irradiance forecasts.
- Ensure rate-limited providers like Solcast continue to serve reindexed and zero-filled stale cache data instead of deleting it.

Enhancements:
- Allow get_cached_forecast_data to return None when cache files are missing, corrupt, or intentionally deleted and update callers to handle this.

Tests:
- Add regression tests covering stale cache handling for Open-Meteo and Solcast forecast providers.
- Align DST-related forecast date tests with standard import ordering.

### DIFF
--- a/src/emhass/forecast.py
+++ b/src/emhass/forecast.py
@@ -416,6 +416,14 @@ class Forecast:
                 data = await self.set_cached_forecast_data(w_forecast_cache_path, data)
         else:
             data = await self.get_cached_forecast_data(w_forecast_cache_path)
+            if data is None:
+                # Stale Open-Meteo cache was deleted — fetch fresh data from API.
+                self.logger.info(
+                    "Stale Open-Meteo cache removed; fetching fresh forecast from API."
+                )
+                return await self._get_weather_open_meteo(
+                    w_forecast_cache_path, use_legacy_pvlib
+                )
         return data
 
     def _solcast_rate_limit_ok(self, max_calls: int = 8) -> bool:
@@ -1826,14 +1834,16 @@ class Forecast:
         self.logger.debug("get_prod_price_forecast returning:\n%s", df_final)
         return df_final
 
-    async def get_cached_forecast_data(self, w_forecast_cache_path) -> pd.DataFrame:
+    async def get_cached_forecast_data(self, w_forecast_cache_path) -> pd.DataFrame | None:
         r"""
         Get cached weather forecast data from file.
 
         :param w_forecast_cache_path: the path to file.
         :type method: Any
-        :return: The DataFrame containing the forecasted data
-        :rtype: pd.DataFrame
+        :return: The DataFrame containing the forecasted data, or ``None`` when
+            the cache is corrupt/missing or was intentionally deleted (e.g. stale
+            Open-Meteo cache).  Callers must handle the ``None`` case.
+        :rtype: pd.DataFrame | None
 
         """
         async with aiofiles.open(w_forecast_cache_path, "rb") as file:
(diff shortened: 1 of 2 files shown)


tomvanacker85:
@sourcery-ai review

tomvanacker85:
@sourcery-ai review

davidusb-geek (approved):

PR #924, opened 2026-06-01
fix: close weather cache file before unlink (Windows WinError 32)

OptimalNothing90:
## What

`Forecast.get_cached_forecast_data()` called `os.remove(w_forecast_cache_path)` from inside
the `async with aiofiles.open(w_forecast_cache_path, "rb") as file:` block — i.e. while the
file handle was still open — in both the corrupt-cache branch and the stale-Open-Meteo branch.

On POSIX, unlinking an open file is fine. On **Windows** it raises:

```
PermissionError: [WinError 32] The process cannot access the file because it is being
used by another process: '...\weather_forecast_data_stale_test.pkl'
```

which fails `tests/test_forecast.py::TestForecast::test_get_cached_forecast_data_stale_open_meteo_deletes_cache`
on the `windows-latest` (and contributes to the `macos-latest`) CI legs.

## Fix

Narrow the context manager to the **read only**, so the handle is closed before any
`os.remove`. All subsequent work (`pickle.loads`, the staleness checks, the deletions, the
reindex/zero-fill path) operates on in-memory data after the file is closed. No behavioural
change on POSIX; the deletions, log lines and return values are identical.

## Verification

- `test_get_cached_forecast_data_stale_open_meteo_deletes_cache` and the Solcast sibling pass.
- Full `tests/test_forecast.py` green locally (31 passed) on Windows.

## Notes

Closes #922. That issue also flagged a `ruff format` drift in `forecast.py` / `test_forecast.py`;
that half was already resolved in v0.17.6 (both files are format-clean on current `master`), so
this PR only carries the Windows file-handle fix.

## Summary by Sourcery

Fix weather forecast cache deletion to close the file before unlinking, ensuring compatibility with Windows file handling, and extend coverage for corrupt cache scenarios.

Bug Fixes:
- Prevent PermissionError on Windows by ensuring cached forecast files are closed before attempting deletion in corrupt and stale Open-Meteo cases.

Tests:
- Add a test verifying that a corrupt cached forecast pickle is deleted and get_cached_forecast_data returns None.

PR #925, opened 2026-06-02
fix: return HTTP 200 (not 201) from JSON API endpoints

### LINKED ISSUE #923: Question: read-only GET endpoints return HTTP 201 Created — intended?
## What I noticed

While generating an OpenAPI contract for the JSON API (#921), I noticed every JSON handler in `web_server.py` returns **`201 Created`**, including the read-only `GET` endpoints:

- `GET /get-config` → `parameter_get`, `web_server.py:354`: `return await make_response(return_config, 201)`
- `GET /get-config/defaults` → `config_get`, `web_server.py:376`: `return await make_response(return_config, 201)`
- `POST /get-json` → `:407` → `201`
- `POST /set-config` → `:472` → `201`
- `POST /action/<action_name>` and the per-action handlers → `:535, :542, :550, :565, :589, :606, :618, :625, :632, :786` → `201`

## The question

`201 Created` semantically signals that a new resource was created. For the **read-only GETs** (`/get-config`, `/get-config/defaults`) that's surprising — `200 OK` would be the conventional code. The blanket `201` across all handlers looks like it might be a shared default rather than a per-endpoint decision.

**Is the `201`-everywhere behaviour intentional?** If a consumer should treat `/get-config` as "ok, here's the config", `200` would be the expected status.

## Caveat (why this is a question, not a PR)

Changing `201` → `200` is a **breaking change** for any existing consumer that checks for `201` (Node-RED flows, HA automations, etc.), so it shouldn't be changed lightly. Wanted to confirm intent before anyone touches it.

For now the generated contract in #921 documents the **actual** behaviour (`201`), so it stays accurate regardless of the outcome here. If you'd like the GETs (or all non-creating responses) aligned to `200`, happy to follow up — just flagging it for a decision.

_Surfaced during #921 (OpenAPI contract); no code change proposed here._

OptimalNothing90:
## ⚠️ BREAKING CHANGE

The JSON API endpoints now return **`200 OK`** on success instead of **`201 Created`**.
**Any client that checks for HTTP `201` must be updated to accept `200`** — this includes
Node-RED flows, Home Assistant REST commands/sensors, and custom scripts that call the
EMHASS API.

## What

Per the decision in #923, the `201`-everywhere behaviour is corrected. Every JSON handler
previously returned `201 Created` on success, including read-only reads. `201` means *a new
resource was created*; none of these endpoints create a resource addressed by the response,
so they now return `200 OK`:

| Endpoint | Before | After |
|---|---|---|
| `GET /get-config` | 201 | **200** |
| `GET /get-config/defaults` | 201 | **200** |
| `POST /get-json` (YAML→JSON convert) | 201 | **200** |
| `POST /set-config` (save config) | 201 | **200** |
| `POST /action/<name>` (all actions: optim, publish, weather-cache, *-model-fit/predict/tune, regressor-*, export-influxdb) | 201 | **200** |

Error responses (`400`/`500`) are unchanged. The internal success sentinel in `action_call`
(`status == 201`) and its final emit are updated in lockstep so success detection still works.

### Why not keep `201` for the "create-like" actions?

`set-config` persists the config file, and `weather-forecast-cache` / `forecast-model-fit` /
`forecast-model-tune` / `regressor-model-fit` write a cache or model artifact. These have a
write side-effect, but none return a representation (or `Location`) of a newly created,
client-addressable resource — they return a status string. Per #923 ("default to 200 unless
a handler truly creates a resource") they return `200`. Happy to bump any specific one back to
`201` if you'd prefer.

## Docs

Added a prominent breaking-change admonition to `docs/usage_guide.md` next to the API examples.
(The `CHANGELOG.md` is auto-generated from PR titles by `scripts/generate_changelog.py`, so the
`fix:` title carries the changelog entry rather than a manual edit.)

## Tests

`tests/test_web_server.py` assertions updated `201` → `200` (13 sites); full `test_web_server.py`
suite passes (48). `test_command_line_utils.py` + `test_last_run.py` green.

Closes #923.

## Summary by Sourcery

Normalize JSON API success responses to return HTTP 200 instead of 201 across web server endpoints and actions.

Bug Fixes:
- Correct JSON API endpoints that incorrectly returned HTTP 201 on successful reads and actions to now return HTTP 200.

Enhancements:
- Align internal action dispatch and success detection logic with the new HTTP 200 success status semantics.

Documentation:
- Document the breaking change in JSON API HTTP status codes in the usage guide so clients update to expect 200 responses.

Tests:
- Update web server tests to assert HTTP 200 success responses from JSON endpoints and actions.

### DIFF
--- a/docs/usage_guide.md
+++ b/docs/usage_guide.md
@@ -16,6 +16,12 @@ With this web server, you can perform RESTful POST commands on multiple ENDPOINT
 
 A `curl` command can then be used to launch an optimization task like this: `curl -i -H 'Content-Type:application/json' -X POST -d '{}' http://localhost:5000/action/dayahead-optim`.
 
+> **⚠️ Breaking change — HTTP status codes.** On success, the JSON API endpoints
+> (`action/*`, `get-config`, `get-config/defaults`, `set-config`, `get-json`) now return
+> **`200 OK`**. Earlier releases incorrectly returned **`201 Created`** for *every* response,
+> including read-only reads. If you have automations or integrations (Node-RED, Home Assistant
+> REST commands, custom scripts) that check for HTTP `201`, update them to accept `200`.
+
 ## Method 2) Legacy method using a Python virtual environment
 
 To run a command simply use the `emhass` CLI command followed by the needed arguments.
--- a/src/emhass/web_server.py
+++ b/src/emhass/web_server.py
@@ -351,7 +351,7 @@ async def parameter_get():
     # Covert formatted parameters from params back into config.json format
     return_config = param_to_config(params, app.logger)
     # Send config
-    return await make_response(return_config, 201)
+    return await make_response(return_config, 200)
 
 
 # Get default Config
@@ -373,7 +373,7 @@ async def config_get():
     # Covert formatted parameters from params back into config.json format
     return_config = param_to_config(params, app.logger)
     # Send params
-    return await make_response(return_config, 201)
+    return await make_response(return_config, 200)
 
 
 # Get YAML-to-JSON config
@@ -404,7 +404,7 @@ async def json_convert():
     config = orjson.dumps(config).decode()
 
     # Send params
-    return await make_response(config, 201)
+    return await make_response(config, 200)
 
 
 @app.route("/set-config", methods=["POST"])
@@ -469,7 +469,7 @@ async def parameter_set():
         return await make_response(["Unable to save params file, missing data_path"], 500)
 
     app.logger.info("Saved parameters from webserver")
-    return await make_response({}, 201)
+    return await make_response({}, 200)
 
 
 async def _load_params_and_runtime(request, emhass_conf, logger):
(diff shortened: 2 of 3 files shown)


OptimalNothing90:
The red CI here is **not** from this change — `tests/test_web_server.py` (which covers these endpoints) passes on CI. The failing jobs are pre-existing on `master` (the same failures show on `master`'s own latest commit), from two unrelated causes:

1. **aiohttp/aioresponses incompatibility (repo-wide):** the mock-based tests in `test_forecast.py` and `test_retrieve_hass.py` fail with `TypeError: ClientResponse.__init__() missing 1 required keyword-only argument: 'stream_writer'`. aiohttp `3.13` made `stream_writer` required on `ClientResponse`; the installed `aioresponses 0.7.8` doesn't supply it. Both are unpinned in `pyproject.toml`, so CI picks up the breaking aiohttp. This affects every open PR.
2. **Windows-only `test_forecast` file-lock** (`WinError 32`) — the separate issue fixed by #924.

This PR only flips success status codes `201`→`200` and updates the matching `test_web_server.py` assertions; once the dependency issue is resolved on `master` and this branch is rebased, the only remaining concern is #924's Windows fix.

davidusb-geek (approved):

Issue #553, opened 2025-05-27
[Feature Request] Allow Desired Battery SOC to be met inside horizon not just by the end


projectcobalt:
I would like to be able to meet the desired SOC at a predefined time-step opposed to just by the end, Ideally this would be inside the PV window for the day and could also include charging from grid to meet the requirement.

The behavior i experience from EMHASS is it generally forecasts soc to be met by the end of the PV window then again at the end of the horizon, in the real world none of these outcomes are met. Specifically if a new horizon is realized that extends into the follow days solar window battery charging is usually deferred to meet the desired soc the next day and the whole process repeats, Battery desired SOC is never met and external intervention needs to take place.

Open to any thoughts or discussions outside a feature requests.

![Image](https://github.com/user-attachments/assets/3f6679f8-979f-4bbf-8d52-262129e01fd7)

purcell-lab:
This would be a very useful feature.  

Often I would like to achieve 100% SOC before sunset, but unless EMHASS can see a reason to charge it won't.
Similarly I would like to achieve 30% SOC at sunrise to get me through the morning peak.  Often the Forecasts for the morning peak prices are low, so EMHASS doesn't charge, but then the prices change after sunrise.

We have been discussing maybe using another deferrable load to drive the battery behaviour.

TD944:
I'm currently in the process of restructuring my plans to meet these requirements. The current plan is to run the MPC calls in staged windows as follows:

During the day the MPC calls will run until 1700 (about sunset but also when I get a boost in price for export). I target 100% SoC here so that the battery is full charged for the night and any evening export. 

After 1700 it swaps to the next morning. I target the minimum reserve to be met by the time solar generation is expected to be 0.5kW (this is an arbitrary level to give a time sometime after sunrise when I expect to be charging again). 

After 2000 (when the good export window is over) it reverts to the 1700 tomorrow target.

Some issues I've had so far:
1) When it gets too close to 1700 and there are single figure windows it can often become infeasible. My current plan is to set a minimum number of forecasts so it doesn't become infeasible. Even though the strategy would not be ideal after 1700 it won't matter as it would never run.
2) Unexpected export at low prices overnight to meet the morning low. This is why I revert to the new target after the sell window. Previously it would sometimes export to bring the SoC down to target. 

[...]

PR #928, opened 2026-06-02
Add optional intermediate battery SoC target (soc_target / soc_target_timestep) — addresses #553

LesIT1:
## What

Two optional `naive-mpc-optim` runtime parameters to require a minimum battery SoC **inside** the horizon, not just at the end:

- **`soc_target`** — desired minimum SoC (fraction in `[0, 1]`); clamped to `[battery_minimum_state_of_charge, battery_maximum_state_of_charge]`.
- **`soc_target_timestep`** — 0-based horizon step by which it must be met (defaults to the last step).

This implements the request in #553 — e.g. "reach 100% by mid-afternoon, then be free to discharge into the evening peak." Today the only SoC target is `soc_final`, which applies only at the horizon end.

## Design

- A one-sided floor `SOC[k] >= soc_target` implemented as a **single precomputed `cp.Parameter` floor vector** (zeros = no-op), set numerically in `perform_optimization`.
- **DPP / warm-start safe** — the floor is computed in NumPy at parameter-set time, so the cached problem stays DPP and re-solves warm-start. A `self.prob.is_dpp()` assertion in the test locks this in.
- **Opt-in, zero behaviour change when unset** — the floor defaults to all-zeros (constraint becomes `SOC >= 0`, a no-op), and is reset on every call so a target can't leak between runs.
- `>=` only, so it never forces a discharge; `soc_final` still governs the tail (the battery is free to discharge after the target step).
- Out-of-range targets are clamped (with a warning); a best-effort reachability check warns when a target likely can't be met in time (which would otherwise surface as LP infeasibility).

## Usage

```json
{ "prediction_horizon": 48, "soc_init": 0.30, "soc_target": 1.0, "soc_target_timestep": 14 }
```

"Be at 100% by horizon step 14, then discharge after." Documented in `passing_data.md` with a Home Assistant `rest_command` example and a reachability note.

## Included

- Constraint + single floor parameter in `optimization.py` (`perform_optimization` and `perform_naive_mpc_optim` signatures + docstrings), with string/float-tolerant timestep parsing and clamp/reachability warnings.
- Runtime plumbing through `utils.py` (`treat_runtimeparams`) and `command_line.py`.
- `runtime_params.json` schema entries.
- Docs: worked example + reachability warning.
- A test asserting the target is met at the requested step, the plan changes vs an unconstrained baseline, the end-of-horizon SoC still lands on `soc_final`, and `prob.is_dpp()` (warm-start caching preserved). The battery / MPC / schema test sweep passes locally.

## Open to feedback

I kept this a **hard** constraint to match #553. Happy to add a **soft-penalty** variant (to degrade gracefully instead of going infeasible when a target is unreachable), or to generalise to **multiple / per-step** targets, if you'd prefer either direction.

### DIFF
--- a/docs/passing_data.md
+++ b/docs/passing_data.md
@@ -99,6 +99,49 @@ Here is the list of the other additional dictionary keys that can be passed at r
 
 - `publish_prefix` use this key to pass a common prefix to all published data. This will add a prefix to the sensor name but also the forecast attribute keys within the sensor.
 
+### Requiring an intermediate battery SOC target (naive-mpc-optim)
+
+By default the battery SOC is only pinned at the start (`soc_init`) and end (`soc_final`) of the horizon, leaving the optimizer free to choose the trajectory in between. Sometimes you want to *guarantee* the battery reaches a given charge level **partway through** the horizon — for example, "make sure the battery is full by the time the cheap window ends, even though it's fine to discharge it afterwards". This is the request in [issue #553](https://github.com/davidusb-geek/emhass/issues/553).
+
+Two optional runtime keys (only used by `naive-mpc-optim`) make this possible:
+
+- `soc_target`: the desired *minimum* SOC as a fraction in `[0, 1]` (NOT percent). It is clamped to `[battery_minimum_state_of_charge, battery_maximum_state_of_charge]`.
+- `soc_target_timestep`: the 0-based horizon timestep by which `soc_target` must be reached. After this step the battery is free to discharge again. Defaults to the last timestep if `soc_target` is given without it.
+
+Both default to *unset* (`None`), in which case nothing changes — the constraint is a no-op and existing behaviour is preserved. The target is a one-sided constraint (`SOC >= soc_target` at that step), so it never forces the battery to discharge.
+
+For example, the following runtime payload tells EMHASS that the battery must reach 100% SOC by horizon step index 14 (the SoC after that step, ~7.5 h into a 30-min-step horizon), while still letting it discharge in the steps after that:
+
+```json
+{
+  "prediction_horizon": 48,
+  "soc_init": 0.30,
+  "soc_target": 1.0,
+  "soc_target_timestep": 14
+}
+```
+
+As an HA `rest_command`:
+
+```yaml
+rest_command:
+  naive_mpc_optim:
+    url: http://localhost:5000/action/naive-mpc-optim
+    method: post
+    content_type: application/json
+    payload: >
+      {
+        "prediction_horizon": 48,
+        "soc_init": {{ states('sensor.battery_soc') | float / 100 }},
+        "soc_target": 1.0,
+        "soc_target_timestep": 14
+      }
+```
+
+```{warning}
+The target must be *reachable* by its timestep. If the requested SOC cannot be achieved in time given `soc_init`, `battery_charge_power_max` and the optimization time step, EMHASS logs a warning and the LP is likely to be **infeasible** (the target is a hard constraint, so it is not silently relaxed). Size `soc_target` / `soc_target_timestep` to what your charge power actually allows (e.g. don't ask for +0.7 SOC in 3 steps if a full charge takes 10). Note too that with the default `set_nodischarge_to_grid: true` a high target can be infeasible if local load is too low to shed the stored energy back down to `soc_final`. `soc_target_timestep` is clamped to `[0, prediction_horizon - 1]`.
+```
+
 ### Passing forecast data
 
 There is a complete dedicated section in the [Forecast](forecasts) section.
--- a/src/emhass/command_line.py
+++ b/src/emhass/command_line.py
@@ -1657,6 +1657,8 @@ async def naive_mpc_optim(
     )
     soc_init = input_data_dict["params"]["passed_data"]["soc_init"]
     soc_final = input_data_dict["params"]["passed_data"]["soc_final"]
+    soc_target = input_data_dict["params"]["passed_data"].get("soc_target", None)
+    soc_target_timestep = input_data_dict["params"]["passed_data"].get("soc_target_timestep", None)
     def_total_hours = input_data_dict["params"]["optim_conf"].get(
         "operating_hours_of_each_deferrable_load", None
     )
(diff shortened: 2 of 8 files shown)


LesIT1:
Rebased onto current `master` (`cba33af`) so this is up to date and picks up the `aiohttp<3.13` test pin from #927.

The earlier red CI on this branch was that same unrelated breakage (`aioresponses` constructing a `ClientResponse` without the `stream_writer` arg required by `aiohttp>=3.13`) — it was inherited from before #927 landed, not introduced here. The diff is unchanged (the `soc_target` feature + tests). On the rebased branch the battery / MPC / schema test sweep passes locally, the previously-failing `test_forecast`/`test_retrieve_hass` mocks pass again with the pin, and `prob.is_dpp()` stays green.

davidusb-geek:
> Rebased onto current `master` (`cba33af`) so this is up to date and picks up the `aiohttp<3.13` test pin from #927.
> 
> The earlier red CI on this branch was that same unrelated breakage (`aioresponses` constructing a `ClientResponse` without the `stream_writer` arg required by `aiohttp>=3.13`) — it was inherited from before #927 landed, not introduced here. The diff is unchanged (the `soc_target` feature + tests). On the rebased branch the battery / MPC / schema test sweep passes locally, the previously-failing `test_forecast`/`test_retrieve_hass` mocks pass again with the pin, and `prob.is_dpp()` stays green.

Tests are failing for other reasons

LesIT1:
You were right — thanks for pushing back. I dug in and there were genuinely a few separate things, only one of which was really this PR's fault:

**1. A brittle test of my own (the real PR failure).** `test_perform_naive_mpc_optim_intermediate_soc_target` asserted the *unconstrained* baseline sat well below the target at the target step, to prove the floor (not arbitrage/tie-breaking) drove the change. But with that test's price profile the baseline already arbitraged the battery up to ~0.99 SoC by the target step, so the precondition `assertLess(baseline, …)` failed before the feature was even exercised. I've rebuilt it as a deterministic **no-arbitrage** scenario (flat, equal import/export prices + a small `weight_battery_charge/discharge`), so the baseline genuinely stays at `soc_init` and the floor's effect is unambiguous. Added two edge-case tests too: clamp-above-`battery_maximum_state_of_charge` (with the warning) and a non-binding floor (`soc_target` below current SoC) proving it's a no-op.

**2. The commit was never `ruff format`-ed.** Fixed — `ruff check` and `ruff format --check` are now clean on the changed files.

**3. Rebased onto current `master`** (`526ad4b`). While there I also corrected a small bug in the reachability *warning*: it multiplied by `battery_charge_efficiency` an extra time (the charge constraint already caps grid-side power at `max_chg/eff`, so the efficiency cancels), which made it warn "may be unreachable" for perfectly reachable targets whenever efficiency < 1. Warning-only, but it was wrong.

**On the remaining red checks:** the other failures aren't from this PR — they're pre-existing on `master` itself (its latest CI run is currently red). A clean checkout of current `master` here fails the same 4: the three `test_forecast.py` Solcast mock tests and `test_openapi.py::test_committed_matches_generated`. The OpenAPI one is a deterministic stale-artifact issue (the committed `openapi.json` drifted from `generate_openapi.py` after some `param_definitions.json` edits, including my own #939) — I've opened #945 to just regenerate it. Full local `uv run pytest` on this branch is now **532 passed, only those 4 pre-existing failures**, with the `soc_target` feature and its three tests green.

The offer in the description still stands — happy to provide a soft-penalty variant (à la `battery_soc_deficit_cost`) instead of / alongside the hard floor if you'd prefer graceful degradation over infeasibility on an unreachable target.

LesIT1:
Quick follow-up on the remaining red CI here. I traced it down and **the only genuine failure was `test_openapi.py::test_committed_matches_generated`** — the committed `openapi.json` is stale on current `master` (it drifted after the inverter-default, naive-mpc auto-extend, and fractional-`operating_hours` `param_definitions.json` edits, including my own #939). It's unrelated to this PR's code.

I opened #945 for just that regen, but to get **this** PR's CI green against the currently-broken `master` I've also **cherry-picked the same one-file `openapi.json` regen onto this branch** (commit `c799de2`, identical to #945 — `scripts/generate_openapi.py` output, no schema/logic change). Full `uv run pytest` locally is now **536 passed, 0 failed** (Linux); CI should follow.

Your call on dedup: merge #945 separately and I'll drop the regen commit from here on the next rebase, or just take it via this PR and I'll close #945 — whichever you prefer. Happy to strip it back out if you'd rather keep this PR scoped purely to `soc_target`.

(For the record: the three `test_forecast` Solcast mock failures that showed earlier were a local-only artifact — a per-day API-call counter in the temp dir hitting its cap of 8 across repeated same-day runs; they pass on a fresh checkout, which #945's all-green CI confirms.)

davidusb-geek (approved):

LesIT1:
Circling back on the soft-penalty version of soc_target I mentioned when I opened this. Now that the hard one is merged, here's what I had in mind if you're interested.

The problem with the current setup is that soc_target is a hard constraint, so if the target can't be reached the whole solve goes infeasible (the docs flag this now). It isn't hard to hit: a low soc_init late in the day, or a high target combined with soc_final and set_nodischarge_to_grid where there isn't enough local load to drain back down. When that happens you get no plan at all instead of the closest one available.

The handy part is you already have both halves of this. There's the hard timed target from this PR (soc_target / soc_target_timestep), and there's already a soft global floor in battery_soc_deficit_cost / battery_soc_deficit_threshold that penalises SoC below a threshold. So the only missing piece is a soft version of the timed target.

I'd add one optional param, soc_target_penalty, defaulting to 0 so it stays off. When it's set, instead of forcing SoC >= soc_target at that step it adds a slack variable and penalises the shortfall in the objective, same idea as battery_soc_deficit_cost but anchored at soc_target_timestep. The solver then reaches the target when it's worth it and gets as close as the penalty justifies otherwise, with no infeasibility.

No new dependencies, and it's a no-op when the penalty is left at 0 so nothing changes for anyone not using it. Same objective-term shape you already have, wired through param_definitions, utils, command_line and optimization with tests, and I'd match the battery_soc_deficit_cost units so the two read the same.

Worth doing? Happy to put it together if so.

PR #934, opened 2026-06-03
fix: restore hybrid inverter AC power defaults to 5000 (#875)

OptimalNothing90:
## Problem

Hybrid inverters go infeasible on the default config since v0.17.3. @rdeknijf tracked it down in #875 with a clean A/B test.

`inverter_ac_output_max` and `inverter_ac_input_max` defaulted to `5000` up to v0.17.2. The #830 config alignment (commit `1f26b03`, first shipped in v0.17.3) set them to `0` to match `param_definitions.json`. That undid `221453e`, which had deliberately raised the hybrid default to 5000.

For `inverter_is_hybrid: true`, an AC cap of 0 forces the inverter's AC output to 0. PV then can't get off the DC bus, so any PV at all makes the problem infeasible. The relaxed-LP retry either fails too, or (before #885) dropped constraints and produced the -200% SOC plans people reported in that thread. Non-hybrid setups never read this constraint, so they are unaffected.

## Change

Restore both defaults to 5000, `param_definitions.json` first (source of truth per AGENTS.md / develop.md), then `config_defaults.json` to match.

Added a regression test in `test_defaults_smoke.py`: a hybrid inverter with PV present solves on the default AC cap, and is infeasible when the cap is 0 (locks the failure mode so an alignment can't silently zero it again).

## Scan of the rest of 1f26b03

I checked the full `config_defaults.json` diff from `1f26b03`. It touched four values:

| key | before | after | verdict |
|---|---|---|---|
| `historic_days_to_retrieve` | 9 | 2 | intended |
| `load_forecast_method` | naive | typical | intended |
| `inverter_ac_output_max` | 5000 | 0 | regression (this PR) |
| `inverter_ac_input_max` | 5000 | 0 | regression (this PR) |

Only the inverter pair broke. The other two were deliberate.

## Open question (this is why it's a draft)

@davidusb-geek, two things before I mark this ready:

- Is 5000 the value you want restored, or something else? 5000 is what 0.17.2 effectively used.
- Should a guard that rejects a 0 AC cap when `inverter_is_hybrid: true` go in this PR or a follow-up? 0 is never a valid AC rating for a hybrid, so a clear error reads better than a silent infeasible.

refs #875

## Summary by Sourcery

Restore correct AC power defaults for hybrid inverters and add regression coverage to prevent reintroducing infeasible configurations.

Bug Fixes:
- Reset hybrid inverter AC input and output default limits to a nonzero value so hybrid configurations remain solvable with default settings.

Tests:
- Add regression tests ensuring hybrid inverters are feasible with the default AC power cap and infeasible when the cap is set to zero.


davidusb-geek:
> @davidusb-geek, two things before I mark this ready:
> 
> * Is 5000 the value you want restored, or something else? 5000 is what 0.17.2 effectively used.
> * Should a guard that rejects a 0 AC cap when `inverter_is_hybrid: true` go in this PR or a follow-up? 0 is never a valid AC rating for a hybrid, so a clear error reads better than a silent infeasible.

Yes we can use 5000. Its a typical PV install capacity size.
You can add the guard directly on this PR

OptimalNothing90:
@davidusb-geek I can't retrigger from mobile, CI should get Green now if you have the Chance

davidusb-geek:
> @davidusb-geek I can't retrigger from mobile, CI should get Green now if you have the Chance

I've just retriggered the failing job.
You also need to put the branch out of draft mode

PR #939, opened 2026-06-03
docs: correct stale "integer hours only" claim for operating_hours_of_each_deferrable_load + add regression test (#373)

### LINKED ISSUE #373: Feature Request: Support for Fractional Hours in operating_hours_of_each_deferrable_load
**Describe the bug**
Currently, the operating_hours_of_each_deferrable_load parameter in EMHASS only accepts scheduling inputs in full hours. For high-power devices, such as an EV charger, the inability to specify fractional hours limits scheduling precision. This becomes especially important when timesteps are smaller (e.g., 15 minutes), as rounding to a full hour can cause significant inefficiencies in energy use.

**Expected behavior**
Allow operating_hours_of_each_deferrable_load to accept fractional hour inputs, rounded to the nearest timestep. Since EMHASS already recognizes the timestep (e.g., 15 minutes), it could automatically round fractional hours to align with the specified timestep. For example, an input of 1.25 hours would be interpreted as 1 hour and 15 minutes when using 15-minute timesteps, rather than being rounded down to 1 hour.

**Additional context**
This enhancement would provide users with greater control over load scheduling, maximizing energy optimization, particularly for devices with high power draw. For instance, consider an 11 kW EV charger:
	•	Scheduling for 1 hour results in an energy consumption of 11 kWh.
	•	Scheduling for 1 hour and 15 minutes (1.25 hours) results in approximately 13.75 kWh.
	•	In contrast, rounding down to 1 hour when 1.25 hours is needed would mean insufficient charging, while rounding up to 2 hours could nearly double the energy draw.

LesIT1:
## Summary

`operating_hours_of_each_deferrable_load` **already accepts fractional values** in the current code (and in the v0.17.6 release): the optimizer computes `target_energy = nominal_power * operating_hours` and `required_timesteps = ceil(operating_hours / time_step)`, so a fractional value schedules the exact `nominal_power × hours` of energy. This PR does **not** add that — it has been the behaviour since the CVXPY optimizer landed.

What this PR fixes is the **documentation**: several study-case docs (added April 2026) describe the parameter as *integer-only* and cite this issue as an open limitation, which isn't accurate. This corrects the docs, locks the behaviour with a regression test, and aligns the parameter's UI schema hint.

## Evidence (it already works)

Day-ahead optim with a fractional value (default 30-min timestep, nominal 3000 W) schedules the exact energy:

| `operating_hours` | scheduled energy | (integer rounding would give) |
|---|---|---|
| 2.5 | 7500 Wh | 6000 / 9000 |
| 1.25 | 3750 Wh | 3000 / 6000 |
| 0.1 (sub-timestep) | 300 Wh | 0 / 3000 |

## Changes

- **Test** — `tests/test_optimization.py::test_fractional_operating_hours`: asserts a fractional `operating_hours` schedules `nominal_power × hours` of energy (to 1e-3 Wh) for 2.5 h and 1.25 h, with an explicit integer-rounding counterfactual so it cannot false-green.
- **Docs** — correct the "integer hours only" claims in `docs/study_cases/ev.md` and `docs/study_cases/reference_configs.md`, the `list[int]` type hint in `docs/cookbook/transport_nodered_mpc_orchestration.md`, and clarify `docs/config.md`. Each now states fractional values are accepted and points to `operating_timesteps_of_each_deferrable_load` for control finer than the timestep.
- **Schema** — `src/emhass/static/data/param_definitions.json`: change `operating_hours_of_each_deferrable_load` `input` from `array.int` to `array.float`, consistent with the sibling deferrable-load float params (`nominal_power_of_deferrable_loads`, `minimum_power_of_deferrable_loads`, …). This field is UI metadata only — not read by the Python runtime, and the config-page JS already parses the input with `parseFloat` — so it has no runtime effect; it simply lets the config UI accept decimals.

## Behaviour notes (full disclosure)

- No optimizer/source logic is changed — this is docs + a test + a UI schema hint.
- A fraction that isn't a whole number of timesteps (e.g. 1.25 h at a 30-min step, or a `set_deferrable_load_single_constant` load at 2.4 h) makes the strict MILP infeasible, so EMHASS falls back to the relaxed LP (`Optimal (Relaxed)`). **The target energy is still met exactly**; only the strict single-constant/binary structure relaxes. For exact on-block structure, use a timestep-aligned value or the runtime parameter `operating_timesteps_of_each_deferrable_load`.
- Incidental: `ruff` stripped trailing whitespace on 2 pre-existing blank lines in `tests/test_optimization.py` (same file).
- Happy to split the `param_definitions.json` change into its own PR if you'd prefer the schema hint to have a separate review thread.

## Tests

`uv run pytest tests/test_optimization.py` passes locally (108 passed, incl. the new test). `ruff check` / `ruff format --check` clean on the changed file.

Closes #373.

## Summary by Sourcery

Clarify that deferrable load operating hours support fractional values, aligning documentation, tests, and UI metadata with existing optimizer behavior.

Enhancements:
- Add a regression test to verify that fractional operating hours schedule the exact target energy for deferrable loads.
- Update configuration UI parameter metadata so `operating_hours_of_each_deferrable_load` accepts floating-point values instead of integers.

Documentation:
- Correct study case and configuration documentation to state that `operating_hours_of_each_deferrable_load` supports fractional hours, and reference `operating_timesteps_of_each_deferrable_load` for finer control.

Tests:
- Add a fractional operating hours regression test for deferrable-load optimization, including a counterfactual check against integer rounding.

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -61,7 +61,7 @@ These are the parameters needed to properly define the optimization problem.
 - `nominal_power_of_deferrable_loads`: The nominal power for each deferrable load in Watts. This is a list with a number of elements consistent with the number of deferrable loads defined before. For example:
 	- 3000
 	- 750
-- `operating_hours_of_each_deferrable_load`: The total number of hours that each deferrable load should operate. For example:
+- `operating_hours_of_each_deferrable_load`: The total number of hours that each deferrable load should operate. Fractional values are accepted (e.g. `4.5`); the optimizer schedules the exact `nominal_power × hours` of energy. For control finer than the optimization timestep you can also pass `operating_timesteps_of_each_deferrable_load`. For example:
 	- 5
 	- 8
 - `start_timesteps_of_each_deferrable_load`: A list of integers defining the **earliest time step index** from which each deferrable load is allowed to start consuming power.
--- a/docs/cookbook/transport_nodered_mpc_orchestration.md
+++ b/docs/cookbook/transport_nodered_mpc_orchestration.md
@@ -32,7 +32,7 @@ Reference table of the runtime params accepted by `treat_runtimeparams` that thi
 
 | Field | Type | Purpose |
 |---|---|---|
-| `operating_hours_of_each_deferrable_load` | `list[int]` | hours each deferrable should run |
+| `operating_hours_of_each_deferrable_load` | `list[float]` | hours each deferrable should run (fractional allowed) |
 | `start_timesteps_of_each_deferrable_load` | `list[int]` | earliest allowed step per deferrable |
 | `end_timesteps_of_each_deferrable_load` | `list[int]` | latest allowed step per deferrable |
 | `load_cost_forecast` | `list[float]` | per-timestep tariff for load |
--- a/docs/study_cases/ev.md
+++ b/docs/study_cases/ev.md
@@ -55,7 +55,7 @@ Two values you compute fresh per MPC call:
 
 | Variable | Source | Calculation |
 |----------|--------|-------------|
-| `def_total_hours[2]` | a sensor that reports remaining kWh to deliver | `kWh_remaining / charger_kW`, rounded to whole hours (see *Known limits*) |
+| `def_total_hours[2]` | a sensor that reports remaining kWh to deliver | `kWh_remaining / charger_kW` (fractional hours are fine) |
 | `end_timesteps_of_each_deferrable_load[2]` | departure deadline | `(deadline - now) / optimization_time_step_minutes`, integer |
 
 A Home Assistant `rest_command` template, source-agnostic:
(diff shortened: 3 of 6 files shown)


LesIT1:
Thanks @sourcery-ai — both points addressed in 77d410e:

1. **Split the two regimes.** The 2.5 h (timestep-aligned) and 1.25 h (sub-timestep) cases are now wrapped in `self.subTest(fractional_hours=...)`, so each reports pass/fail independently and both run even if one fails — matching the existing `subTest` usage elsewhere in this file.
2. **Scale-relative counterfactual margin.** The hardcoded `1.0` Wh threshold is now `0.1 * nominal * timestep_h` (≈150 Wh at the defaults), so it tracks the optimization timestep and nominal power. It stays comfortably above solver noise (~1e-3 Wh) and well below the ≥750 Wh gap from either test value to its nearest integer-hour schedule, so the discriminating power is unchanged.

`pytest tests/test_optimization.py::TestOptimization::test_fractional_operating_hours` is green (2 subtests), and `ruff check` / `ruff format --check` are clean on the file.

davidusb-geek (approved):

PR #945, opened 2026-06-06
fix: regenerate committed openapi.json (stale vs param_definitions)

LesIT1:
## Problem

`tests/test_openapi.py::TestOpenapiArtifact::test_committed_matches_generated` is currently **failing on `master`** — the committed `src/emhass/static/openapi.json` has drifted from what `scripts/generate_openapi.py` produces.

## Cause

A few recent changes edited `src/emhass/static/data/param_definitions.json` (titles/defaults/descriptions) without re-running the generator, so the committed spec went stale:

- hybrid-inverter `inverter_ac_output_max` / `inverter_ac_input_max` defaults `0 → 5000` (#875 / #934),
- the `delta_forecast_daily` description update from the naive-mpc auto-extend forecast-window work,
- the `operating_hours_of_each_deferrable_load` fractional-value description (#373 / #939).

## Fix

Just re-ran `python scripts/generate_openapi.py` and committed the result. **No schema or code-logic change** — only the committed artifact is brought back in sync (5 lines). `test_committed_matches_generated` passes locally afterward.

```
1 passed in 4.76s
```

Happy to close this if you'd rather fold the regen into another change.

## Summary by Sourcery

Bug Fixes:
- Resolve the mismatch between the committed openapi.json and the version produced by scripts/generate_openapi.py, restoring the passing openapi artifact consistency test.

### DIFF
--- a/src/emhass/static/openapi.json
+++ b/src/emhass/static/openapi.json
@@ -359,7 +359,7 @@
             "default": 1,
             "title": "Number of forecasted days",
             "x-unit": "days",
-            "description": "The number of days for forecasted data. Defaults to 1. (unit: days)"
+            "description": "The number of days for forecasted data; can be overridden at runtime via runtimeparams. Defaults to 1. For naive-mpc, the forecast window is now extended automatically to cover a longer prediction_horizon, so this parameter usually need not be set manually for multi-day MPC runs. (unit: days)"
           },
           "load_forecast_method": {
             "type": "string",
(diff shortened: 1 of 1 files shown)


OptimalNothing90:
Thanks @LesIT1 — this regen is correct; `test_committed_matches_generated` should pass again once it lands.

One heads-up on the current wave: the committed-vs-generated check (added in #921) will fire again whenever a `param_definitions.json` edit forgets to regenerate the spec. Two open PRs do exactly that — #947 (Darts forecaster) and #948 (per-deferrable command-state sensors) edit `param_definitions.json` without regenerating, and #928 edits `openapi.json` directly (possible conflict with this one depending on merge order).

We added the generator + drift-check in #921, so once this merges we will follow up with the systemic fix — an automated regen (a pre-commit/CI step running `scripts/generate_openapi.py`) so the committed spec cannot go stale by hand. Your one-time regen here is the right call to get master green now.

LesIT1:
Makes sense. If #945 lands first I'll rebase #928 and pull the openapi.json commit out of it, so there's no conflict and the committed spec just follows generate_openapi.py. That commit was only ever there to get #928's CI green against the stale master, so it's dead weight the moment this merges.

Auto-regen in CI is the right fix, would've saved me the cherry-pick here. I'll keep an eye on #928 and rebase whenever this or master moves.

davidusb-geek (approved):

Issue #847, opened 2026-05-13
[Feature request] Adding Weather forecast / weather data / outside temperature for ML


Michirieger:
Hey there, 

in the docs you allready noted that you may add other variables like Weather or outside temperature.
In my case the Heatpump is the main Load in my household which is directly dependent to my presence, weekday/time and the weather.
So in my view introducing the Weather data into the ML process could enhance the Load Forecast quiet a bit.

Also not that experienced with Python i am thinking about developing that feature and contibuting it.
As stated in the Docs i open this issue but also to ensure your interest in such a feature or any points that make it useless/obsolete as i dont want to live in a fork the rest of my life :D

Thank you for your great work

BrettLynch123:
I put together an implementation that I think directly addresses this: a native, optional **Darts (LightGBM) time-series load forecaster**, selectable via `load_forecast_method: darts`. PR: BrettLynch123/emhass branch `feat/darts-load-forecaster-pr` (opened against `master`).

It is built specifically around the weather-aware use case described here (a heatpump-dominated load that depends on weather, time and day-of-week):

- **Weather as future covariates.** The model consumes weather columns that are *known* for the forecast window (temperature, cloud cover, humidity, etc.) as Darts future covariates, on top of always-included calendar features. Columns are configured with a new `darts_covariate_columns` option, so it is fully generic — you choose which retrieved weather columns to feed it.
- **Single-shot, not recursive.** Unlike the existing `mlforecaster` (recursive scikit-learn), it emits the whole horizon in one pass, so it does not accumulate recursive error over a long horizon.
- **Optional uncertainty bands.** A new `darts_quantiles` option produces a probabilistic forecast (e.g. P10/P50/P90). The P50 is what `get_load_forecast` returns; the other quantiles are exposed for callers that want uncertainty. This also covers the load-forecast half of #841.
- **Zero-config-friendly.** It works on the load sensor alone (calendar-only) if you pass no covariates, so it is not gated on having weather data.

It mirrors the existing `MLForecaster` fit/predict/tune interface and is wired through the documented add-a-parameter workflow (`associations.csv` + `config_defaults.json` + `param_definitions.json` + `command_line.py` + `forecast.py`). Models are saved as `{model_type}_darts.pkl` alongside the existing `_mlf.pkl`.

[...]

Issue #841, opened 2026-05-11
[Feature request] Include the effect of forecast uncertainty


lutorm:
Has anyone considered how to include the _uncertainty_ in the PV forecast? Solcast also provides a forecast of the 10% and 90% percentiles, and it seems like it would be good to try to take this into account.

An optimal plan for the nominal forecast that incurs an energy deficit if the actual PV comes in below the forecast, because it sold off a bunch of battery energy earlier, is pretty brittle and probably not best in the long run. If  the effect of forecast uncertainty were symmetric, it would average out in the long run, but it's not.
I'm pretty sure that loss from the realization coming in below the forecast usually significantly outweigh the gains if it comes in above, since the cost of purchasing extra energy to make up for a shortfall later usually far outweighs what you get for selling off energy you think will be in excess. Thus, a somewhat "defensive" strategy that holds some energy back for adverse outcomes is probably better.

A possible way to include this in the optimization might be to actually simultaneously optimize all 3 cases, nominal, 10%, and 90%, simultaneously, and pick the path that maximizes some weighting of the merit functions in the three scenarios.

A similar situation exists wrt the load forecast. I dunno if the mlforecaster provides an estimate of the forecast uncertainty, but if it does it might make sense to use it as well.

purcell-lab:
+1 — this is a real gap and the asymmetric-loss argument is the right framing.

## Where EMHASS sits today

EMHASS currently consumes only the **nominal** Solcast estimate. The `_get_weather_solcast` helper in `forecast.py` reads `forecasts[].pv_estimate` and discards `pv_estimate10` / `pv_estimate90`, even though Solcast returns all three in the same API response — so the P10/P90 data is already in flight, just not retained. That's a useful implementation hook for whichever direction this lands.

On the load side, the `mlforecaster` (skforecast-based) supports prediction intervals via the `predict_interval()` path, so quantile-style load forecasts are achievable too — they're just not currently surfaced as inputs to the LP.

Related but distinct: [#473](https://github.com/davidusb-geek/emhass/issues/473) (closed) added historical-bias correction for PV forecasts, and [#564](https://github.com/davidusb-geek/emhass/issues/564) covers the `adjust_pv_forecast` path. Both shift the *mean* of the forecast; this FR is about acting on the *spread* around the mean — orthogonal and complementary.

## On the approach

[...]

PR #947, opened 2026-06-07
feat: add optional Darts (LightGBM) time-series forecaster for load, cost and price

BrettLynch123:
## What

Adds a native, **optional** time-series forecaster built on [Darts](https://unit8co.github.io/darts/) (a single-shot LightGBM model) as an alternative to the existing recursive scikit-learn `mlforecaster`. The forecaster is **target-agnostic**, so the same model serves three EMHASS forecasts, each selectable independently:

- the **load** forecast — `load_forecast_method: darts`
- the **load cost** forecast — `load_cost_forecast_method: darts`
- the **PV production price** forecast — `production_price_forecast_method: darts`

It mirrors the `MLForecaster` fit/predict/tune interface and adds:

- **Weather future covariates** — the model can use weather that is *known* for the forecast window (temperature, cloud cover, humidity, …) via the `darts_covariate_columns` option, on top of always-included calendar features.
- **Single-shot horizon** — emits the whole horizon in one pass, so it does not accumulate recursive error over a long horizon (verified non-degenerate out to 7 days).
- **Optional uncertainty bands** — the `darts_quantiles` option produces a probabilistic forecast (e.g. P10/P50/P90); the P50 is returned, the rest are exposed for callers that want uncertainty.
- **Negative-target support** — `darts_non_negative` (default `true`, a zero floor for a load in watts) can be set `false` for a price target, which can legitimately go negative (negative spot prices, e.g. Amber).
- **Zero-config-friendly** — works on the target sensor alone (calendar-only) if no covariates are passed.

Load models are saved as `{model_type}_darts.pkl`; price/cost models as `{model_type}_load_cost_darts.pkl` / `{model_type}_prod_price_darts.pkl`, alongside the existing `_mlf.pkl`.

## Why

Directly addresses #847 (weather-aware ML load forecasting — the heatpump/weather-dependent-load use case) and the load-forecast half of #841 (forecast uncertainty). The price/cost extension was requested in this PR's discussion by @purcell-lab, who uses the existing ML forecaster for load, price **and** cost out to 7 days, and wanted a weather covariate for price/cost.

**Load** is where the accuracy win is, and it's real: on ~8 months of 5-minute history, with identical folds, Darts beat a production Lasso baseline by **~7% MAE**, and the 7-day single-shot horizon stays non-degenerate (~6 s) where a recursive model flattens toward the mean.

**Price/cost: honest framing — this is an *option*, not an accuracy win.** An earlier, narrower 12-fold window suggested Darts was ~15% better than same-time-yesterday persistence on Amber import price. That figure did **not** hold up under a deeper look: a 20-fold rolling-origin backtest plus a dedicated price-forecast study show Darts essentially **ties a seasonal (same-time, day-of-week-aware) naive** on price, and the realised **bill impact is ≈$0** — price is dominated by the time-of-day / weekly cycle, which a good seasonal-naive already captures, so a weather covariate adds little here. I'm keeping the price/cost path in the PR not because it forecasts price better, but because it (a) gives a single, unified forecaster across load, cost and price instead of separate mechanisms, and (b) carries the weather-covariate capability through to cost/price for anyone who wants it — most usefully for users **without a dedicated satellite PV/forecast source** (e.g. Solcast), where a weather-aware ML model has more to offer. If you'd prefer to land the load forecaster alone and drop the price/cost extension, that's an easy split — happy to do it.

## How the price/cost wiring works

The `DartsForecaster` already fits/predicts an arbitrary `var_model` target, so:

- **Training** reuses the existing `forecast-model-fit` action unchanged — point `var_model` at the price sensor and use `model_type` `<base>_load_cost` / `<base>_prod_price`; fit routing now also triggers when a cost/price method is `darts`.
- **Prediction** adds a `darts` branch to `Forecast.get_load_cost_forecast` / `get_prod_price_forecast`: it loads the matching pickle, fetches a fresh recent window of the price sensor from Home Assistant to anchor the lags, and returns the horizon-aligned series. Calendar covariates are exact; any weather covariates are reused where the window supplies them and forward-filled otherwise. The price window is taken raw (not zero-clamped) so negative prices survive.

The two `get_*_forecast` methods stay synchronous (they have many synchronous call sites and test mocks); the async fetch + async `predict` run via a small private-event-loop bridge, leaving all signatures and callers untouched.

## Design / scope

- **Optional dependency isolation.** Darts and LightGBM are an `emhass[darts]` optional extra, imported **lazily** with a clear, actionable install error when absent. Default installs are completely unaffected.
- **Zero-config default preserved.** `load_forecast_method` / `load_cost_forecast_method` / `production_price_forecast_method` all keep their existing defaults; nothing changes unless a user opts into `darts`.
- **Add-a-parameter workflow** followed: `associations.csv` + `config_defaults.json` + `param_definitions.json` (the three method enums gain `darts`; the `darts_covariate_columns` / `darts_quantiles` / `darts_non_negative` params) + `utils.py` runtime params + regenerated `openapi.json`. The new params affect forecasting only, not LP problem structure, so no `OptimizationCacheKey` field is needed.
- **Out of scope (kept focused):** the standalone `forecast-model-predict` / `forecast-model-tune` actions still operate on the load model; price/cost prediction flows through the `get_*_forecast` methods at optimization time, which is all the optimizer needs.

## Testing

- `tests/test_darts_forecaster.py`: **17 tests** — **8 that always run** (clear dependency-error path + load/cost/price config-schema wiring + new-param defaults + the sync→async bridge surfacing real exceptions + the list-param normalizer + the **missing-extra dispatch path for load-cost and production-price** asserting it's logged and returns `False`) and **9 functional tests** (deterministic & quantile fit/predict, tune, sanity-check, the load dispatch, the **load-cost** and **production-price** dispatch, a missing-price-model guard, and a **negative-price** assertion that the zero floor is correctly skipped) that **skip cleanly when the optional extra is not installed**. All 17 pass with the extra installed; 8 pass / 9 skip without it.
- `ruff check` + `ruff format --check` clean.
- `openapi.json` regenerated; the drift test (`test_committed_matches_generated`) passes.
- Regression-checked: the existing load/cost/price/fit tests (`test_get_load_cost_forecast`, `test_get_prod_price_forecast`, `test_get_load_forecast_mlforecaster`, `test_forecast_model_fit_predict_tune`, …) pass unchanged.

## Code-quality

While in the forecaster: made it genuinely target-agnostic (the zero-floor is now opt-out via `non_negative`), de-duplicated the fit/predict window-preparation into a shared helper, and hardened `sanity_check` so an empty or all-NaN forecast can't raise from `np.nanmax`.

**Sourcery review addressed:**
- **Real bug fixed** in the sync→async bridge (`_run_coro_blocking`): if the bridged coroutine raised before setting its result, the worker thread exited and the caller hit a misleading `KeyError` that masked the true error. The worker now captures the exception and re-raises it on the calling thread, so failures surface properly (regression test added).
- The missing-extra (`DartsDependencyError`) case is now caught in the forecast **and** command-line routing for load, cost **and** price — logged with a clear, actionable message and surfaced as `False`/return (mirroring the missing-model-file handling) instead of a raw stack trace.
- `darts_quantiles` / `darts_covariate_columns` are normalized when they arrive as strings (JSON/YAML scalar or UI text) into proper typed lists, so a string like `"[0.1, 0.5, 0.9]"` isn't exploded into characters.
- The Darts fit error message is target-agnostic (it serves load/cost/price, not only load).
- `sanity_check` already guards the empty/all-NaN `np.nanmax` edge case flagged in the review.

Happy to adjust the covariate handling, the price-window retrieval, the optional-extra packaging, or defaults to match how you'd prefer to land this — and, as noted above, to split the price/cost extension out if you'd rather land the load forecaster alone.

## Summary by Sourcery

Introduce an optional Darts/LightGBM-based time-series forecaster for load, load cost, and production price, wired into existing fit/predict/tune and forecast flows, with optional weather covariates and quantile outputs, plus documentation and tests.

New Features:
- Add a Darts-based single-shot LightGBM time-series forecaster as an alternative to the existing scikit-learn ML forecaster for load, load cost, and production price.
- Expose new configuration options for Darts covariate columns, quantile outputs, and non-negative targets, including OpenAPI and schema updates.
- Support Darts-based forecasting methods for load cost and PV production price, using shared model infrastructure and separate saved models.

Enhancements:
- Extend forecast command-line fit, predict, and tune paths to load and handle either MLForecaster or DartsForecaster models with clear error handling when optional dependencies are missing.
- Add a sync-to-async bridge utility to run async coroutines from synchronous forecast methods without changing existing call sites.
- Improve configuration handling by normalizing list-typed Darts parameters from strings into typed lists and ensuring robust forecast sanity checks.

Build:
- Declare an optional 'darts' extra dependency in the project configuration for Darts and LightGBM.

Documentation:
- Document the new Darts time-series forecaster, its configuration, usage for load, cost, and price, and when to choose it over the existing ML forecaster.

Tests:
- Add an extensive Darts forecaster test suite covering dependency handling, configuration wiring, deterministic and probabilistic forecasts, tuning, dispatch for load, load cost, and production price, and negative-price behaviour.


BrettLynch123:
CI status: the full Python test matrix now passes — all `build` jobs (macOS / Ubuntu / Windows / Debian amd64+arm64) are green. The only remaining red is the `run` job, and within it solely the *Upload coverage reports to Codecov* step — a tokenless Codecov upload from a fork with `fail_ci_if_error: true`. The pytest run itself (the "Generate Report" step) passes, so this is environmental and clears on a maintainer re-run with repo secrets, not a code/test failure. Happy to address any further review feedback — thanks!

purcell-lab:
@BrettLynch123 Is this implementation limited to load forecasting only? 

I currently use the existing EMHASS ML forecaster for both load, price & cost forecasting, out to 7 days. 

Weather is also a big factor in price & cost forecasting so being able to use a weather covariate would be excellent.

BrettLynch123:
Hi @purcell-lab — good questions!

Yes, this PR scopes Darts to **load** forecasting (it plugs in as `load_forecast_method: darts`, mirroring where the existing `mlforecaster` sits) to keep the change focused and reviewable.

The `DartsForecaster` class itself is generic, though — it fits/predicts a target series with optional weather + calendar covariates and quantiles — so reusing it for the price/cost forecast methods (the same way the ML forecaster is reused there) would be a natural follow-up rather than a rewrite. Weather as a covariate for price/cost is a great point; Darts takes arbitrary future covariates, so the plumbing carries straight over.

On the 7-day horizon: it's a single-shot (direct multi-horizon) LightGBM, so it avoids the recursive error-compounding that tends to hurt long autoregressive forecasts — `output_chunk_length` just needs to span your horizon.

Keen to see Darts useful beyond load — happy to explore a price/cost extension as a follow-up once this one lands, if there's interest. Thanks for taking a look!



Just updating this. I am doing some checking of what needs to happen for it to do price & cost forecasting. if I find it works and if any code changes are needed I will update this PR with its findings.

BrettLynch123:
@purcell-lab — ran some backtests for real numbers:

**Price:** yes, Darts is target-agnostic — on my Amber import-price history (5-min, ~8 months) it gets ~2.5 c/kWh MAE vs ~2.9 for same-time-yesterday persistence (~15% better) on the cleaner Nov–Jun window, fully non-degenerate.

**Weather → price:** only marginal — ~3% (≈0.1 c/kWh) at the larger sample, near the noise floor. Price is dominated by the time-of-day/weekly cycle, so weather helps it far less than it helps load. (My bench uses Open-Meteo *archive* weather, so the live-forecast benefit would be a touch lower still.)

**7-day horizon:** works cleanly — single-shot `output_chunk_length=2016` produces a non-degenerate forecast in ~6s, no recursive flatten, and essentially no accuracy decay from 48h to days 5-7 (it leans on the stable diurnal/weekly cycle).

For load (this PR's focus), on my system Darts beats the prod Lasso baseline by ~7% MAE on identical folds.

Given that, I'm extending this PR to wire Darts into the cost/price forecast methods too (with the weather-covariate option), so it'll serve load, price & cost — I'll update here once it's pushed. Thanks for the nudge!

purcell-lab:
Excellent. 

I didn't realize you are on Amber too.

You might be interested in this which is a complete recalibration of the AEMO 7 day pre dispatch forecasts.

https://github.com/purcell-lab/nem_pd7day

BrettLynch123:
@purcell-lab — done, it's pushed. Darts now serves **load, load cost and production price**, each selectable independently:

```json
"load_forecast_method": "darts",
"load_cost_forecast_method": "darts",
"production_price_forecast_method": "darts"
```

How it works for price/cost:

- **Train** with the existing `forecast-model-fit` action — just point `var_model` at your price sensor and use `model_type` `<base>_load_cost` / `<base>_prod_price` (the fit routing now also fires when a cost/price method is `darts`). Set `darts_non_negative: false` so negative spot prices aren't clamped.
- **Predict** loads the matching pickle, pulls a fresh window of the price sensor from HA to anchor the lags, and returns the horizon-aligned price series. Weather covariates carry straight over via `darts_covariate_columns` — calendar features are always exact, weather is reused wherever the window supplies it.

Matches what I measured on my Amber price history (the numbers above): ~15% better than persistence, non-degenerate single-shot out to 7 days, and a smaller but real weather gain — so `darts_covariate_columns: ["temperature_2m", ...]` is worth a try on your price/cost models, just expect a more modest lift than on a weather-driven load.

New docs section: [Using it for the load cost and production price](https://github.com/davidusb-geek/emhass/blob/feat/darts-load-forecaster-pr/docs/darts_forecaster.md#using-it-for-the-load-cost-and-production-price). I've taken the PR out of draft. Thanks again for the nudge — keen to hear how it does on your 7-day price/cost.

BrettLynch:
> Excellent. 
> 
> I didn't realize you are on Amber too.
> 
> You might be interested in this which is a complete recalibration of the AEMO 7 day pre dispatch forecasts.
> 
> https://github.com/purcell-lab/nem_pd7day

Yes I'm with Amber as a customer great company. I am heading to Melbourne office in a few weeks. And that would be interesting will take a look tomorrow.

BrettLynch123:
## Correction on the price/cost numbers — being upfront

I want to correct the record on the **price/cost accuracy claim** I made earlier in this PR, because it doesn't hold up and I'd rather be honest than oversell it.

An earlier comment (and the original PR body) cited Darts as **~15% better than persistence** on Amber import price. That figure came from a **narrow 12-fold window** and was not robust. A deeper look — a **20-fold rolling-origin backtest** plus a dedicated price-forecast study — shows Darts essentially **ties a seasonal (same-time, day-of-week-aware) naive** on price, and the **realised bill impact is ≈$0**. Price here is dominated by the time-of-day / weekly cycle, which a good seasonal-naive already captures, so the weather covariate adds little on the price target specifically.

So, to be clear about what this PR does and doesn't give you:

- **Load: the ~7% MAE gain is real** (identical folds vs a production Lasso baseline), and the single-shot horizon stays non-degenerate where a recursive model flattens. That's the accuracy win.
- **Price/cost: this is an *option*, not an accuracy win.** I'm keeping it because (a) it gives one unified forecaster across load, cost and price instead of separate mechanisms, and (b) it carries the weather-covariate capability through to cost/price for anyone who wants it — most useful for users **without a dedicated satellite PV/forecast source** (e.g. Solcast). It is *not* in here because it forecasts price better — it doesn't.

I've updated the PR body to reflect this honest framing rather than silently editing the earlier figures, so the history of the discussion stays transparent. If you'd prefer to land the **load forecaster alone** and drop the price/cost extension, that's a clean split and I'm happy to do it.

davidusb-geek:
@BrettLynch123 Please see my comment in https://github.com/davidusb-geek/emhass/issues/847

BrettLynch123:
Closing this in favour of #955.

Thanks for the thoughtful feedback @davidusb-geek — you're right that Darts + LightGBM is too heavyweight to bring into core, especially for add-on users who can't easily add an optional extra. I've taken your "build on what's already there" suggestion and opened **#955**, which adds opt-in weather-covariate support to the existing skforecast `mlforecaster` instead — no new dependency, backward-compatible (byte-identical when the option is off), and it directly addresses the weather-aware load-forecasting use case in #847. On a real heat-pump-influenced home it gave a ~12% load-MAE improvement in a paired 20-fold backtest.

Appreciate the steer toward the lighter approach — that's the better direction for core. I'll keep the Darts forecaster running externally on my own setup, feeding EMHASS via the `load_power_forecast` input (the external-forecaster pattern @purcell-lab also suggested), so nothing's lost.

PR #948, opened 2026-06-07
feat: add optional per-deferrable command-state sensors

### LINKED ISSUE #946: [Feature request] Optional interpretable command-state sensors per deferrable load
## Feature request

Today each deferrable load is published only as a **power** sensor (`sensor.p_deferrableN`). To act on the optimizer's decision, every automation has to re-derive intent by thresholding that raw power value itself (e.g. "if `P_deferrable0` > some W then switch on"). Each user reinvents the same thresholding logic, and it is easy to get wrong for loads that modulate.

It would be useful for EMHASS to optionally publish a ready-made, interpretable **command** per deferrable load, so automations can act on a clear label instead of the raw power forecast.

## Proposal

An **opt-in** option (default off, so the zero-config behaviour and all existing sensors are unchanged) that additionally publishes one command sensor per deferrable load:

- `sensor.p_deferrableN_state`, whose state is `on`, `off` or `variable` for the current timestep;
- the full optimized plan over the horizon attached as a `schedule` attribute (a list of `{date, value}` entries), so an automation can also look ahead.

The labels are derived from the optimized `P_deferrableN` power and the load's `nominal_power_of_deferrable_loads`, so they stay consistent with the existing power forecast:

- `off` when the load is essentially idle;
- `on` when it runs at (near) its nominal power;
- `variable` for any modulated level in between.

The label is deliberately generic (`on`/`off`/`variable`) so it stays **glue-layer agnostic** — the user maps it onto their own switch, number or script (Node-RED, MQTT, HA, generic automations), rather than any HA-specific control living in core.

## Notes

- Strictly additive and opt-in: nothing new is published unless the option is enabled; no existing sensor/column changes.
- The entity ids would be overridable at runtime (mirroring `custom_deferrable_forecast_id`).

I have a small implementation ready (a new `categorical` publish type + a pure power→label mapping + the opt-in publisher, with unit tests and docs) and can open a PR if this direction is welcome.

BrettLynch123:
## What

Adds an **opt-in**, glue-agnostic command surface for deferrable loads. When the new `publish_deferrable_load_states` option is enabled, EMHASS additionally publishes one interpretable command sensor per deferrable load:

- `sensor.p_deferrableN_state`, whose state is `on`, `off` or `variable` for the current timestep;
- the full optimized plan over the horizon attached as a `schedule` attribute (a list of `{date, value}` entries).

## Why

Today each deferrable load is published only as a **power** sensor (`sensor.p_deferrableN`), so every automation has to re-derive intent by thresholding the raw power forecast itself. This lets automations act on a clear, ready-made label instead — and because the label is generic (`on`/`off`/`variable`), it stays **glue-layer agnostic** (Node-RED, MQTT, HA, generic automations) rather than baking any HA-specific control into core.

## Issue

Closes #946.

## Design / scope

- **Opt-in, zero-config default preserved.** Defaults to `false` — nothing new is published and all existing sensors are unchanged unless the user opts in.
- **Consistent with the power forecast.** Labels are derived from the optimized `P_deferrableN` power and `nominal_power_of_deferrable_loads`: `off` when idle, `on` at (near) nominal power, `variable` for any modulated level in between (non-finite power is treated as `off`, defensively).
- **Strictly additive.** No existing `opt_res_*` columns, units or sensors change (respects the additive-only invariants for `param_definitions.json` and the result DataFrame).
- **Runtime-overridable** entity ids via `custom_deferrable_state_id`, mirroring `custom_deferrable_forecast_id`.
- Implementation: a new `categorical` `post_data` type in `retrieve_hass.py` (string state + `schedule` attribute, no numeric rounding, reusing existing plumbing); a pure `_deferrable_power_to_state` mapping + an opt-in `_publish_deferrable_states` publisher wired into `publish_data` in `command_line.py`; a `custom_deferrable_state_id` default in `utils.py`; and the new option in `associations.csv` + `config_defaults.json` + `param_definitions.json`.

## Testing

- New `tests/test_deferrable_states.py`: **2 tests** — the power→label mapping (`_deferrable_power_to_state`) and the `categorical` `post_data` path (asserts the string state + the `schedule` attribute shape). Both pass; neither needs any optional extra.
- `ruff check` + `ruff format --check` clean.
- Regression-checked: the existing `tests/test_command_line_utils.py` + `tests/test_forecast.py` failure set is identical with and without this PR's changes against `master` — no new failures introduced.
- Docs added (`docs/publish_data.md`).

## Summary by Sourcery

Add optional publication of categorical command-state sensors for deferrable loads and support for a new categorical post_data type, along with documentation and tests.

New Features:
- Introduce opt-in per-deferrable command-state sensors derived from optimized power and nominal load power.
- Add runtime-configurable defaults for deferrable load command-state entity identifiers.

Enhancements:
- Extend the publish pipeline with a categorical data path that sends string states plus full-horizon schedules as attributes for command-like sensors.

Documentation:
- Document the new publish_deferrable_load_states option and the resulting deferrable load command-state sensors in publish_data.md.

Tests:
- Add tests covering the deferrable power-to-state mapping and the categorical post_data path including schedule attributes.


BrettLynch123:
CI status: the full Python test matrix now passes — all `build` jobs (macOS / Ubuntu / Windows / Debian amd64+arm64) are green. The only remaining red is the `run` job, and within it solely the *Upload coverage reports to Codecov* step — a tokenless Codecov upload from a fork with `fail_ci_if_error: true`. The pytest run itself (the "Generate Report" step) passes, so this is environmental and clears on a maintainer re-run with repo secrets, not a code/test failure. Happy to address any further review feedback — thanks!

davidusb-geek (approved):

PR #950, opened 2026-06-07
forecast: add timeout + cold-start retry to the Open-Meteo fetch

### LINKED ISSUE #949: Open-Meteo fetch has no request timeout and no cold-start retry (recurring CI 502 flake)
### Summary

`Forecast.get_cached_open_meteo_forecast_json()` fetches the Open-Meteo weather forecast with a **single** `aiohttp` request and **no explicit per-request timeout**. On any `aiohttp.ClientError` it falls back to the pre-loaded cache if one exists, otherwise it returns `None`. Two robustness gaps follow:

1. **No request timeout.** The `aiohttp.ClientSession()` is created without a `ClientTimeout`, so the default (long) timeout applies. A slow or hanging Open-Meteo response can therefore stall the whole optimisation cycle.

2. **No retry on a cold start.** When there is no cache to fall back on — the first run after install, and importantly the CI environment (a fresh runner has no cached JSON) — a single transient error (e.g. a `502`/`504` from Open-Meteo rate-limiting the runner IP) makes the fetch give up and return `None` rather than retrying.

Gap (2) is the cause of the recurring `api.open-meteo.com` flake in CI. The Python-test workflow makes a real network call to Open-Meteo in:

- `tests/test_forecast.py::TestForecast::test_get_forecasts_with_lists_special_case`
- `tests/test_forecast.py::TestForecast::test_get_weather_forecast_openmeteo_method_mock`

and they intermittently fail with `aiohttp.client_exceptions.ClientResponseError: 502, message='Bad Gateway'` / `ClientConnectionError: Connection refused`, moving between the Ubuntu and Windows matrices run to run. A single transient error in a cache-less environment has no retry to recover from.

The same single-attempt behaviour appears to be what users hit at runtime, e.g. #794 (a `client_error` at the start of a run that succeeds on an immediate re-run).

### Proposal

Keep the steady-state behaviour byte-for-byte and fast, and only make the cache-less path resilient:

1. Wrap the GET in an `aiohttp.ClientTimeout(total=...)` (a small constant, ~12s) so a hanging request cannot stall the cycle.
2. Retry with a short backoff (e.g. 3 attempts, 1s/2s/4s) **only when there is no usable cache** (`data is None`). When a cache is present, keep exactly one attempt and fall back to the cache **immediately** on failure — no retry, no added delay (this is the common path, run every cycle).
3. Preserve the existing semantics: never overwrite the cache on a failed fetch, and keep returning a present cache on failure.

Net effect: the steady-state (cache present) path is unchanged and adds zero delay; only the cold-start (no cache) path retries — which also lets a transient Open-Meteo `502` self-recover in CI.

This touches `forecast.py` beyond ~3 lines, so I'm opening this issue first per `AGENTS.md`. Happy to send the PR (with tests for the retry-then-succeed, immediate cache fall-back, and timeout paths) if this approach looks good.

BrettLynch123:
Closes #949

### What

Makes the Open-Meteo weather fetch in `Forecast.get_cached_open_meteo_forecast_json()` robust to a slow/hanging endpoint and to transient errors on a cold start, **without adding any delay to the steady-state path**.

Today the method does a single `aiohttp` GET with no explicit timeout; on any `aiohttp.ClientError` it falls back to the pre-loaded cache if one exists, else returns `None`. There is no request timeout and no retry — so a transient `502`/`504` (or a hang) on the first run / in CI cannot recover.

### Changes

- **Per-request timeout.** The GET is now wrapped in `aiohttp.ClientTimeout(total=12)` so a slow or hanging Open-Meteo response can no longer stall the optimisation cycle.
- **Cold-start retry only.** When there is **no usable cache** (`data is None`), the fetch retries up to 3 times with a `1s/2s/4s` backoff. When a cache **is** present, behaviour is unchanged: a single attempt, and on failure an **immediate** fall-back to the cache — no retry, no added delay. This is the key guarantee: the common, every-cycle path is byte-for-byte as before.
- **Preserved semantics.** The cache is still never overwritten on a failed fetch, and a present cache is still returned on failure.

The constants live at module scope (`open_meteo_request_timeout`, `open_meteo_max_attempts`, `open_meteo_backoff_seconds`) so they're easy to tune/patch.

### Why it also fixes the recurring CI flake

The `api.open-meteo.com` flake (`502 Bad Gateway` / `Connection refused`, moving between the Ubuntu/Windows matrices) hits the two tests that make a real Open-Meteo call. A CI runner has **no cached JSON**, so it always takes the cache-less branch — which previously gave up on the first transient error. With the cold-start retry, a transient `502` self-recovers instead of failing the job.

### Tests

Added to `tests/test_forecast.py` (mocked with `aioresponses`, no real network):

- `test_open_meteo_cold_start_retries_then_succeeds` — no cache: a `504` then a connection error then success; asserts the payload is returned and exactly 3 attempts were made.
- `test_open_meteo_cold_start_all_attempts_fail_returns_none` — no cache, all attempts `502`: returns `None` and writes no cache file.
- `test_open_meteo_cache_present_falls_back_immediately_no_retry` — cache present + forced refresh that fails: returns the cached payload, asserts **exactly one** attempt (no retry) and that the cache file is left untouched.
- `test_open_meteo_request_timeout_is_set` — asserts a finite `ClientTimeout` is supplied to the session and that a `TimeoutError` is handled like any other transient error.

Backoff is patched to zero in the retry tests so the suite doesn't sleep.

### Validation

- Full `tests/test_forecast.py` passes locally from a clean data dir (`38 passed`).
- `ruff format --check` and `ruff check` are clean on the changed files.

Note: the network-dependent Open-Meteo tests may still flake on this PR's CI run until this lands, since the fix targets exactly that path — please re-run if you see a `502`/`Connection refused` from `api.open-meteo.com`.

## Summary by Sourcery

Add bounded-timeout and cold-start-only retry logic to the Open-Meteo forecast fetch while preserving existing cache semantics.

Bug Fixes:
- Prevent slow or hanging Open-Meteo responses from stalling the optimisation cycle by enforcing a per-request timeout.
- Allow transient Open-Meteo errors on a cold start (including in CI) to self-recover via limited retries instead of failing permanently.

Enhancements:
- Introduce configurable timeout and retry/backoff parameters for the Open-Meteo fetch at module scope to ease tuning and patching.

Tests:
- Add async tests covering cold-start retries, all-attempts-fail behaviour, cache-present no-retry fallback, and enforcement of the Open-Meteo request timeout.


davidusb-geek:
This branch has failing tests

BrettLynch123:
Quick note on the red CI here: it's not caused by this PR. This branch only touches `forecast.py` + `tests/test_forecast.py`, and the failure is entirely a stale-base artifact.

Across all the red jobs (`build` on ubuntu/macos/windows + the `CodeCov` run), the **only** failing test is:

- `tests/test_openapi.py::TestOpenapiArtifact::test_committed_matches_generated` — `committed openapi.json != freshly generated`, on the `delta_forecast_daily` description ("…Defaults to 1. (unit: days)" vs "…can be overridden at runtime via runtimeparams…").

Everything else is green (the build jobs report `536 passed` alongside that one failure).

The reason is just that this branch is **behind `master`**. Its merge-base is `526ad4b`, which predates #945 ("regenerate committed openapi.json — stale vs param_definitions"). #945 already regenerated `openapi.json` on `master` to match the reworded `delta_forecast_daily`, so the staleness this PR's CI trips has since been fixed upstream — it just isn't in this branch's base yet.

I merged current `master` into this branch's tree locally and re-ran the suite: `test_committed_matches_generated` passes, the four `test_optimization.py::...::test_thermal_battery_*` tests pass, and the full `tests/test_forecast.py` (38 tests, including the four new Open-Meteo retry/timeout/cache-fallback ones) is green. So updating this branch to `master` should turn CI fully green — happy to push that rebase whenever it's convenient to review.

One caveat that's expected for this particular change: the network-dependent Open-Meteo tests (`test_get_weather_forecast_openmeteo_method_mock`, `test_get_forecasts_with_lists_special_case`) can still hit the recurring `api.open-meteo.com` 502 / Connection-refused flake on CI — the cold-start retry this PR adds is exactly what lets that path self-recover.

BrettLynch123:
Two CI reds, neither caused by this PR:

1. **Codecov GPG flake** (both `run` and `build (ubuntu-latest)` / CodeCov workflow): same `gpg: Can't check signature: No public key` failure that is also hitting upstream `master` (run [#27099793490](https://github.com/davidusb-geek/emhass/actions/runs/27099793490)) — unrelated to the code change.

2. **Time-of-day-sensitive tests** (`test_get_weather_forecast_csv`, `test_get_power_from_weather`, `test_get_power_from_weather_single_system`, `test_get_prod_price_forecast`) — `ValueError: Length mismatch: Expected axis has 25 elements, new values have 48 elements`. These tests load `data_weather_forecast.csv`, a single-day fixture, and fail when CI runs at a time of day where fewer than 48 slots remain in that day. They are pre-existing and pass on upstream `master` at other hours — not introduced by this PR. A re-run at a different time of day would clear them.

Could you trigger a re-run when convenient? Happy to address any other feedback.

davidusb-geek (approved):

Issue #900, opened 2026-05-27
Prevention tests: extend the #882 contract/smoke suite to cover stringly-typed and non-default config paths


OptimalNothing90:
The schema-contract and defaults-smoke tests from #882 catch two classes: input-type drift between `param_definitions.json` and the renderer, and crashes when loading the shipped JSON defaults. Reviewing the recent thermal/cost PRs (#886, #888, #897) surfaced a class they don't cover: bugs triggered by stringly-typed or non-default config values.

None of these are caught by #882 as it stands:

- `cost_forecast_per_deferrable_load` / `deferrable_load_max_cost` as string `"0.0"` or with `None` elements raises TypeError. The defaults are `[0.0, 0.0]`, so the smoke never reaches the string path.
- `heat_topology` / `sense` as the string `"null"` or YAML null crashes. These live under `def_load_config`, not in `config_defaults.json`, so the contract test never sees them.

The common thread is the HA add-on UI persisting values as strings, plus per-load arrays with null elements. One parametrized "stringly-typed resilience" test over every per-load-array and config-dict parameter would cover the class and catch the next instance automatically, rather than one regression test per bug.

Proposed extension to the #882 suite:

1. a parametrized test feeding `"0.0"` / `None` / `"null"` through `treat_runtimeparams` plus a minimal `perform_optimization` for each per-load-array param, asserting no crash and a sane fallback;
2. a thermal fixture so the smoke exercises `def_load_config` sub-keys (`sense` and friends) with null values.

Follow-on to #882. I'm happy to put it up. Tracking it here so the coverage gap stays visible.

PR #953, opened 2026-06-07
fix(utils): stop spurious deferrable_load_max_cost warning on startup (#929)

### LINKED ISSUE #929: v0.17.6: deferrable_load_max_cost warning fires even when list length matches num_def_loads
## Bug description

After upgrading to v0.17.6, this WARNING appears on every EMHASS startup even with a correctly-sized `deferrable_load_max_cost` list:

```
WARNING in utils: deferrable_load_max_cost does not match number in num_def_loads, adding default values (0.0) to parameter
```

The warning fires regardless of value content — observed with both `[null, null, null]` and `[0.0, 0.0, 0.0]`.

## Environment

- EMHASS version: **0.17.6**
- Python: 3.12.13
- Installation: HA add-on (`5b918bf2_emhass`)
- `number_of_deferrable_loads`: **3**

## Config snippet (relevant)

```json
{
  "number_of_deferrable_loads": 3,
  "nominal_power_of_deferrable_loads": [11000, 5000, 25000],
  "deferrable_load_max_cost": [0.0, 0.0, 0.0]
}
```

## Expected behavior

With a list of length 3 matching `number_of_deferrable_loads: 3`, no warning should fire.

## Actual behavior

WARNING printed on every startup. Optimization runs normally with status `Optimal` (defaults are applied), so functionality is not affected — but the false positive pollutes logs.

## Background

Prior to adding the key entirely, EMHASS raised:
```
KeyError: 'deferrable_load_max_cost'
  File "/app/src/emhass/utils.py", line 3040, in check_def_loads
    result = parameter[parameter_name]
```
Adding the key with a correctly-sized list resolves the `KeyError` but produces the false-positive WARNING shown above.

## Related

Introduced by #886 (Add optional max cost threshold for deferred loads).

LesIT1:
Fixes #929.

## What's happening

`check_def_loads` logs a WARNING whenever it pads a per-load array up to `number_of_deferrable_loads`. The defaults in `config_defaults.json` are all sized for the default count of 2, so as soon as someone sets `number_of_deferrable_loads: 3` without also restating every per-load array, the padding kicks in and the warning fires on every startup. `deferrable_load_max_cost` is the one people actually report because it was added recently (#886) and isn't always written back into existing configs, so it falls back to the length-2 default while the other arrays don't.

Padding short lists up to the load count is exactly what the function is documented to do ("if they do not match, enlarge to fit"), and the optimisation runs fine afterwards, so the WARNING is a false alarm. This drops it to debug.

## Changes

- `check_def_loads`: log the padding at debug instead of warning.
- `check_def_loads`: a missing key or an explicit `null` now returns a full default-filled list instead of raising `KeyError`. That's the other half of the same report, the original `KeyError: 'deferrable_load_max_cost'` people saw before adding the key by hand.
- Removed a duplicate `check_def_loads` call for `deferrable_load_max_cost` in `build_params`. It ran right after the canonical one and passed an int `0` default instead of the float `0.0` the canonical call uses. It was a no-op (the first call already padded the list), so dropping it doesn't change the result.

The stringly-typed path from the #900 resilience work is untouched: a present-but-non-list value is still returned as-is.

## Behaviour changes worth calling out

- `check_def_loads` with a missing key was a `KeyError`, now returns `[default] * num_def_loads`. The existing unit test that asserted the `KeyError` is updated to the new contract.
- The padding message moves from WARNING to DEBUG, so it no longer shows at the add-on's default INFO level.

## Tests

Expanded `test_check_def_loads` in `tests/test_utils.py`:
- missing key returns the default-filled list (was asserting `KeyError`)
- explicit `None` returns the default-filled list
- a length-2 list with count 3 pads to length 3, logs at DEBUG and never at WARNING (the #929 regression)
- the existing pad / no-pad cases still pass

Full `uv run pytest` is green locally apart from the three `test_forecast.py` solcast mock tests, which fail the same way on a clean `master` (a Solcast API safety cap in my environment), so they're unrelated to this change.

Happy to add a CHANGELOG line if you'd like one, I just didn't see an unreleased section to slot it into.

## Summary by Sourcery

Relax deferrable load configuration handling by treating missing or shorter per-load arrays as defaults rather than errors and silencing noisy startup warnings.

Bug Fixes:
- Prevent KeyError when deferrable load per-load arrays such as deferrable_load_max_cost are missing or set to null in the configuration.
- Stop spurious startup warnings by downgrading padding messages for undersized deferrable load lists from warning to debug and removing a redundant deferrable_load_max_cost check.

Tests:
- Extend check_def_loads unit tests to cover missing keys, explicit null values, and debug-only logging when padding shorter per-load arrays.

### DIFF
--- a/src/emhass/utils.py
+++ b/src/emhass/utils.py
@@ -2907,13 +2907,6 @@ async def build_params(
             "nominal_power_of_deferrable_loads",
             logger,
         )
-        params["optim_conf"]["deferrable_load_max_cost"] = check_def_loads(
-            num_def_loads,
-            params["optim_conf"],
-            0,
-            "deferrable_load_max_cost",
-            logger,
-        )
         # Validate deferrable_load_groups
         groups = params["optim_conf"].get("deferrable_load_groups", [])
         if groups:
@@ -3073,19 +3066,26 @@ def check_def_loads(
     :return: parameter list
     :rtype: list[dict]
     """
-    if (
-        parameter.get(parameter_name, None) is not None
-        and type(parameter[parameter_name]) is list
-        and num_def_loads > len(parameter[parameter_name])
-    ):
-        logger.warning(
+    current = parameter.get(parameter_name, None)
+    # Missing key or explicit JSON null: apply the default for every load. This is
+    # expected when a per-load array is absent from the user config (e.g. a parameter
+    # added in a later release), so do it silently rather than raising KeyError.
+    if current is None:
+        parameter[parameter_name] = [default] * num_def_loads
+        return parameter[parameter_name]
+    if type(current) is list and num_def_loads > len(current):
+        # Enlarging a short list to match number_of_deferrable_loads is this function's
+        # documented job, not an error: the shipped defaults are sized for the default
+        # load count, so any user that raises the count without restating every per-load
+        # array lands here. Log at debug to avoid spurious startup warnings (#929).
+        logger.debug(
             parameter_name
-            + " does not match number in num_def_loads, adding default values ("
+            + " has fewer entries than number_of_deferrable_loads, padding with default ("
             + str(default)
-            + ") to parameter"
+            + ")"
         )
-        for _x in range(len(parameter[parameter_name]), num_def_loads):
-            parameter[parameter_name].append(default)
+        for _x in range(len(current), num_def_loads):
+            current.append(default)
     result = parameter[parameter_name]
     # Replace any None elements with the default (can occur when set-config
     # receives a partial config, e.g. [null, 0] instead of [0, 0]).
--- a/src/emhass/utils.py
+++ b/src/emhass/utils.py
@@ -3053,6 +3053,10 @@ def check_def_loads(
     """
     Check parameter lists with deferrable loads number, if they do not match, enlarge to fit.
 
+    A missing key or ``None`` value is filled with ``default`` for every load. ``parameter``
+    is updated in place (and the same list returned), matching how every call site reassigns
+    ``params["optim_conf"][name] = check_def_loads(...)``.
+
     :param num_def_loads: Total number deferrable loads
     :type num_def_loads: int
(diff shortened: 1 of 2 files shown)


LesIT1:
Thanks for the review. Took the first point in c757d33: switched to `isinstance(current, list)`, which also lines up with the `isinstance` check already used further down in the same function.

On the second: I kept the in-place mutation rather than returning a fresh list. It's the function's existing contract already, the pad and the None-replacement paths both mutate `parameter`, and every call site does `params["optim_conf"][name] = check_def_loads(...)` and then reads back the updated dict, so a non-mutating branch would leave the function half-mutating. I documented the in-place behaviour in the docstring instead.

LesIT1:
Heads up on CI: the `run` (CodeCov) check is failing here, but it's also failing on the current `master` HEAD (400c595), so it looks pre-existing rather than something this PR introduced. The `build` (pytest) jobs on this PR pass. Locally the full suite is green under both `pytest` and `coverage run -m unittest` apart from three `test_forecast.py` solcast mock tests, which fail the same way with my changes stashed out (a Solcast API safety cap in my environment), so they're unrelated to this change.

LesIT1:
Correction to my CI note above. The `run` (CodeCov) check isn't failing on the tests, it fails at the Codecov upload step, which has `fail_ci_if_error: true` and needs `secrets.CODECOV_TOKEN`. GitHub doesn't pass secrets to PRs from forks, so that step can't succeed on any fork PR regardless of the code. The tests themselves pass under both `pytest` and `coverage run -m unittest`. (The solcast detail I mentioned earlier was a local rate-limit counter on my own machine, not relevant to CI.)

I opened #954 to gate that upload step on forks the same way the SonarQube step already is, which should let this check pass on external PRs. The `build` (pytest) jobs here are green.

davidusb-geek (approved):

PR #956, opened 2026-06-08
fix: recover from schema-incompatible weather forecast cache (issue #932)

### LINKED ISSUE #932: [BUG] Solcast stale-cache fallback crashes with KeyError: 'yhat' in get_power_from_weather (v0.17.3+, not fixed by PR #918)
## Summary

Since v0.17.3, when `get_cached_forecast_data()` serves stale reindexed+zero-filled Solcast data, the downstream `get_power_from_weather()` crashes with `KeyError: 'yhat'` at `forecast.py:902`. This renders Solcast-based installations non-functional for MPC whenever the stale-cache path is triggered.

PR #918 (merged v0.17.6) intentionally preserved the stale-cache-reuse path for Solcast to protect API rate limits — but the preserved path is broken: instead of returning zeroed (wrong) forecasts as it does for Open-Meteo users before the fix, it returns a DataFrame that is missing the `yhat` column entirely, causing an unhandled crash.

---

## Environment

- **EMHASS version:** 0.17.6 (Home Assistant add-on)
- **Home Assistant version:** Home Assistant OS (add-on environment)
- **`weather_forecast_method`:** `solcast`
- **Failing endpoint:** `/action/naive-mpc-optim`

---

## Symptoms

The following sequence reproduces on every MPC call once the stale-cache path is triggered:

1. `/action/weather-forecast-cache` completes successfully at 05:00 (HTTP 200).
2. `/action/naive-mpc-optim` is called later (08:00 onward) via HA automation with a standard payload:
   ```json
   {"soc_init": 0.02, "prediction_horizon": 6}
   ```
3. EMHASS log shows:
   ```
   WARNING in forecast: Cache does not fully cover the requested timeframe.
     Serving best-effort stale data (reindexed + zero-filled). Cache preserved to protect API rate limits.
   ```
4. EMHASS crashes immediately after:
   ```
   ERROR in app: Exception on request POST /action/naive-mpc-optim
     File "/app/src/emhass/forecast.py", line 902, in get_power_from_weather
       p_pv_forecast = df_weather["yhat"]
   KeyError: 'yhat'
   ```
5. HTTP 500 returned. MPC aborts. Downstream integrations do not update.

Confirmed HA-triggered observations (2026-06-03, AEST):

| Time | HA payload | Result |
|---|---|---|
| 09:30:00 | `soc_init=0.02, prediction_horizon=6` | HTTP 500 — KeyError: 'yhat' |
| 10:00:00 | `soc_init=0.025, prediction_horizon=6` | HTTP 500 — KeyError: 'yhat' |

---

## Why this is Solcast-specific

The `yhat` column is the output of a Prophet model fitted to Solcast irradiance data at runtime — it is **not** stored in the raw Solcast pkl. The pkl contains raw API columns (e.g. `pv_estimate`, `pv_estimate10`, `pv_estimate90`).

When the stale-data fallback reindexes and zero-fills the pkl:
- The resulting DataFrame contains only raw API columns.
- `yhat` is absent.
- `get_power_from_weather()` at `forecast.py:902` performs an unconditional `df_weather["yhat"]` lookup → `KeyError`.

---

## Relationship to PR #918

PR #918 branches on `weather_forecast_method`:
- `"open-meteo"`: delete stale pkl and return `False` → triggers a fresh fetch → `yhat` is recomputed cleanly.
- **all other methods (including Solcast)**: keep the reindex+zero-fill path to protect API quota.

The Solcast preservation is intentional and correct in principle — but the reindexed DataFrame does not satisfy the `yhat` precondition that `get_power_from_weather()` assumes.

---

## Comparison with closed issues

| Dimension | Open-Meteo (#916/#917, fixed in v0.17.6) | Solcast (this report) |
|---|---|---|
| Stale-cache fallback triggers | YES | YES |
| "Cache does not fully cover" warning | YES | YES |
| Result | Zero-filled irradiance (wrong but no crash) | `KeyError: 'yhat'` crash, HTTP 500 |
| Fixed in v0.17.6 (PR #918) | YES | **NO** |
| Active in v0.17.6 | NO | **YES** |
| Upstream issue open | NO (closed) | **This report** |

---

## Why the coverage check may trigger despite recent cache refresh

Exact cause not confirmed without source access. Most likely explanation: the Solcast pkl stores timestamps in UTC (naive or UTC-aware), while `forecast_dates` computed at MPC call time may be timezone-aware (local time). The coverage check performs an `in` test (`self.forecast_dates[0] in data.index`) — if the timezone representations differ, the check fails as an exact-match failure even though the underlying instants are identical.

This would mean the stale-cache path triggers on every MPC call, regardless of how recently the Solcast cache was refreshed.

---

## Expected behaviour

The Solcast stale-cache fallback should do one of:

1. Recompute / derive a `yhat`-equivalent column from the raw Solcast API columns before passing the DataFrame to `get_power_from_weather()`, **or**
2. Return a clear recoverable cache-miss condition (similar to the Open-Meteo path) when `yhat` cannot be guaranteed, **or**
3. Guard `forecast.py:902` against the absence of `yhat` and raise an informative, catchable error rather than an unhandled `KeyError`.

---

## Actual behaviour

Unhandled `KeyError: 'yhat'` at `forecast.py:902` → HTTP 500. MPC is completely non-functional for Solcast users whenever the stale-cache path is triggered (which appears to be on every call).

---

## Safety note

This crash was observed and diagnosed in a shadow-only environment:
- Remote EMS was not enabled.
- No hardware writes were performed.
- Both write gates remained `off` throughout.

---

## Questions for maintainers

1. Should the raw Solcast pkl produced by `/action/weather-forecast-cache` contain a `yhat` column? If so, at what point is it computed and written?
2. Is the coverage check timezone-sensitive? Could a UTC pkl index vs. a local-timezone-aware `forecast_dates` cause a false coverage failure even after a successful pre-cache?
3. Is there a recommended workaround for Solcast users on v0.17.6 until a fix is available (e.g. a config flag, a manual cache format, or a fallback order)?

LesIT1:
Fixes #932.

## What's going on

The weather forecast cache lives in one file, `data/weather_forecast_data.pkl`, and the name is the same for every `weather_forecast_method`. So when the method in use changes (open-meteo and solcast running at different times, say), the cache on disk can belong to a different method than the one running now.

The stale-cache fallback in `get_cached_forecast_data` then serves that old frame as-is, and solcast/solar.forecast hit an unconditional `df_weather["yhat"]` in `get_power_from_weather`. An open-meteo cache has ghi/dni/dhi, not yhat, so it throws `KeyError: 'yhat'` and the optim returns a 500. A valid solcast cache is fine here (it does carry a yhat column), so the crash specifically needs a schema-mismatched cache, which the shared file produces once methods differ.

## The fix

Make the rate-limited methods self-heal like open-meteo already does (it deletes a stale cache and refetches):

1. `get_cached_forecast_data`: before returning a served cache, if the active method needs `yhat` and it isn't there, warn, drop the incompatible cache, and return `None` (already the documented recoverable-miss signal).
2. `_get_weather_solcast` and `_get_weather_solar_forecast`: on `None`, fall through to a fresh fetch instead of returning `None`, mirroring `_get_weather_open_meteo`.

Quota stays protected by the existing `_solcast_rate_limit_ok` cap, so worst case is one refetch of a genuinely unusable cache. The guard skips `solar_forecast_kwp == 0` (where yhat is never read) and leaves valid caches alone, so #918's rate-limit-preserving behaviour holds for the normal stale-but-valid case. The mirror case (a yhat-only cache read as open-meteo over a covered window) is pre-existing and method-gated out of this change; happy to do a follow-up if you want it.

## Tests

Added `test_get_weather_forecast_solcast_incompatible_cache_recovers`: writes an open-meteo-schema cache (no yhat) over a stale window, runs solcast against a mocked API, and asserts the result has `yhat` and `get_power_from_weather` doesn't raise. Fails on master, passes with the fix; full suite stays green (539 passed).

### DIFF
--- a/src/emhass/forecast.py
+++ b/src/emhass/forecast.py
@@ -463,7 +463,11 @@ class Forecast:
     async def _get_weather_solcast(self, w_forecast_cache_path: str) -> pd.DataFrame:
         """Helper to retrieve weather data from Solcast or cache."""
         if os.path.isfile(w_forecast_cache_path):
-            return await self.get_cached_forecast_data(w_forecast_cache_path)
+            cached_data = await self.get_cached_forecast_data(w_forecast_cache_path)
+            if cached_data is not None:
+                return cached_data
+            # Incompatible/stale cache was discarded (issue #932); fall through
+            # to fetch fresh data from the Solcast API (quota-guarded below).
         if self.params["passed_data"].get("weather_forecast_cache_only", False):
             self.logger.warning("Solcast cache file missing or deleted due to being out of date.")
             self.logger.warning(
@@ -545,7 +549,11 @@ class Forecast:
     async def _get_weather_solar_forecast(self, w_forecast_cache_path: str) -> pd.DataFrame:
         """Helper to retrieve weather data from solar.forecast or cache."""
         if os.path.isfile(w_forecast_cache_path):
-            return await self.get_cached_forecast_data(w_forecast_cache_path)
+            cached_data = await self.get_cached_forecast_data(w_forecast_cache_path)
+            if cached_data is not None:
+                return cached_data
+            # Incompatible/stale cache was discarded (issue #932); fall through
+            # to fetch fresh data from the forecast.solar API.
         # Validation and Default Setup
         if "solar_forecast_kwp" not in self.retrieve_hass_conf:
             self.logger.warning(
@@ -1898,6 +1906,31 @@ class Forecast:
                     data[irradiance_cols] = data[irradiance_cols].fillna(0.0)
 
                 data = data.fillna(0.0)
+
+        # The weather cache file is shared across every weather_forecast_method,
+        # so a cache written by a different method can lack the column the active
+        # method needs. Serving it would crash later in get_power_from_weather
+        # with an opaque KeyError: 'yhat' (issue #932). Treat a schema-incompatible
+        # cache as a recoverable miss: drop it and return None so the rate-limited
+        # fetchers fall through to a fresh fetch (the open-meteo path already does
+        # this for its own stale cache).
+        if (
+            self.weather_forecast_method in ("solcast", "solar.forecast")
+            and self.retrieve_hass_conf.get("solar_forecast_kwp") != 0
+            and "yhat" not in data.columns
+        ):
+            self.logger.warning(
+                "Cached forecast is missing the 'yhat' column required by "
+                "weather_forecast_method='%s' (cache likely written by a "
+                "different method). Discarding the incompatible cache and "
+                "fetching fresh data.",
+                self.weather_forecast_method,
+            )
+            try:
+                os.remove(w_forecast_cache_path)
+            except FileNotFoundError:
+                pass
+            return None
         return data
 
     async def set_cached_forecast_data(self, w_forecast_cache_path, data) -> pd.DataFrame:
--- a/src/emhass/forecast.py
+++ b/src/emhass/forecast.py
@@ -460,14 +460,25 @@ class Forecast:
             self.logger.error(f"Failed to check or increment Solcast rate limit: {e}")
             return False
 
+    async def _get_cached_forecast_or_none(self, w_forecast_cache_path: str) -> pd.DataFrame | None:
+        """Return a usable cached forecast, or None when there is no cache file
+        or it was discarded as stale/schema-incompatible (issue #932).
+
+        Lets the rate-limited fetchers (solcast, solar.forecast) share one
+        cache-recovery path: a non-None result is served directly, None means
+        fall through to a fresh API fetch.
+        """
+        if not os.path.isfile(w_forecast_cache_path):
+            return None
+        return await self.get_cached_forecast_data(w_forecast_cache_path)
+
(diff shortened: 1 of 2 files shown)


LesIT1:
Thanks @sourcery-ai, took both in e495066:

- **Decoupled the schema check from `solar_forecast_kwp`.** You're right that solcast doesn't use that key, and it turned out to be a latent bug rather than just a smell: `solar_forecast_kwp` defaults to 5 but a real solcast user can leave it at 0, in which case the old `solar_forecast_kwp != 0` guard switched the `yhat` check off and served the incompatible cache anyway (the exact crash this PR targets). The check now keys only on the active method needing `yhat`. I extended the regression test to pin `solar_forecast_kwp = 0`, and confirmed it fails on the old guard and passes with the fix.
- **Extracted the shared recovery path** into a small `_get_cached_forecast_or_none` helper (the `os.path.isfile` + `get_cached_forecast_data` dance), now used by both the solcast and solar.forecast fetchers so they stay consistent.

Full solcast/solar suite + ruff green.

davidusb-geek (approved):

PR #957, opened 2026-06-08
feat: add set_battery_first_priority to drain battery before grid import (closes #834)

### LINKED ISSUE #834: [Feature request] `set_battery_first_priority` — enforce battery-before-grid for flat tariffs
# Feature request: `set_battery_first_priority` — enforce battery-before-grid for flat tariffs

## Summary

Add an optional MILP constraint that forbids grid import while the battery
is above its minimum SOC. With a flat (non-time-of-use) electricity
tariff, the current cost function leaves the battery-vs-grid sourcing
order under-determined and the solver can plan grid imports while the
battery is still well above SOC_min. This is a real-world problem for
users on fixed tariffs who would (rationally) prefer to drain the battery
first before importing.

Closely related to #795/#796 (which forbids simultaneous discharge **and**
import). The proposal here is the next logical step: forbid imports
**at any time** when the battery has stored energy left to give.

## Use-case / motivation

A typical user setup with:

- flat tariff (e.g. Germany, ~0,28 €/kWh import / ~0,082 €/kWh export)
- 24h load > usable battery capacity (overnight load drains the battery
  before the next sunrise)
- inverter that does not allow battery export to grid
  (`set_nodischarge_to_grid: True`)
- read-only mode (no inverter setpoint write-back yet) **or** write-back
  active

In this configuration the optimizer regularly produces plans like:

```
21:00  load=1320W  p_grid=+1199W  p_batt=+127W   soc=85.0  ← split import while battery has 75 % usable energy left
22:00  load= 581W  p_grid= +581W  p_batt=   0W   soc=80.6  ← idle battery, full grid import
22:30  load= 669W  p_grid=    0W  p_batt=+704W   soc=75.8  ← battery resumes
…
07:30  load= 532W  p_grid=    0W  p_batt=+360W   soc=11.3
08:00  load= 673W  p_grid=    0W  p_batt= +75W   soc=10.0  ← reaches SOC_min
08:30  load= 618W  p_grid=    0W  p_batt= +98W   soc=10.0
09:00  load= 566W  p_grid=    0W  p_batt=   0W   soc=10.0
```

Total grid import is the unavoidable shortfall (load minus battery
capacity), but the temporal distribution is arbitrary because the
solver finds many cost-equivalent solutions. With a write-back
controller this is **actually executed** — and the user ends up with
ca. 75 % SOC at 22:00 when there is no rational reason to keep that
reserve at a flat tariff: losing capacity available to absorb the next
day's PV peak is the dominant risk.

## Proposed solution

Add a new Boolean configuration parameter
`set_battery_first_priority` (default `false`, opt-in).

When enabled, add a Big-M constraint per timestep:

```python
# Pseudo-code in Optimization._add_battery_constraints
if self.optim_conf.get("set_battery_first_priority", False):
    M = self.plant_conf["battery_nominal_energy_capacity"]   # safe upper bound
    eps = 0.01                                               # 1 % SOC tolerance
    g = cp.Variable(self.opt_horiz, boolean=True)            # g_t = 1 ↔ grid import allowed
    constraints += [
        P_grid_pos <= self.plant_conf["maximum_power_from_grid"] * g,
        SOC[t] - self.plant_conf["battery_minimum_state_of_charge"] - eps <= M * (1 - g[t])
        for t in range(self.opt_horiz)
    ]
```

Semantics: `g_t = 1` (import allowed in slot *t*) requires SOC ≤
SOC_min + ε; otherwise `g_t = 0` and `P_grid_pos[t] = 0`. The battery
is therefore fully drained to the lower bound before grid import is
allowed.

This is consistent with the existing `E ≤ D` pattern (PR #796) and adds
~96 binaries per 48-h horizon — negligible against the existing battery
mode and grid direction binaries.

## Alternatives considered

1. **Cost-function tweaks (`weight_battery_discharge`, `weight_p_grid_pos`,
   `costfun: cost`).** Tested; in flat-tariff with active SOC_min
   constraint the total battery discharge is constant across all optimal
   solutions, so weighted Discharge-bonus has no leverage on the
   temporal split. Confirmed empirically on a real install
   (full discussion in #XXX-this-issue).
2. **Pre-solve deterministic simulation.** Maintainable as user-side
   pyscript/AppDaemon, but is by-design out of EMHASS scope per
   AGENTS.md §4 (and #789).
3. **Different solver / `lp_solver_mip_rel_gap`.** Does not break ties
   in this specific structural ambiguity.

## Backward compatibility

Default `false` — current behaviour preserved. Existing users see no
change. Users on flat tariffs opt in.

## Tests

Reproducer setup (will provide as `tests/test_battery_first_priority.py`):

- Mock 48-slot horizon with constant load > available battery energy
- `costfun: profit`, flat HP/HC tariff
- Run dayahead-optim once with `set_battery_first_priority: false`,
  once with `true`
- Assert: in the first slots where `SOC > SOC_min + ε`,
  `P_grid_pos[t] == 0` in the second run

Will follow the test/coverage standard set by PR #796.

## Implementation plan / questions

I'd like to take the implementation. Open questions for the maintainer:

1. Is `optim_conf` the right place for this flag, or do you prefer
   `plant_conf` (it's strategy, not plant)? My read of `associations.csv`
   suggests `optim_conf`.
2. Should this also be exposed as a runtime parameter override
   (`runtimeparams`)? Useful to A/B test from a Home Assistant automation
   without restarting the addon.
3. Acceptable to add the schema entries (`config_defaults.json`,
   `param_definitions.json`, `associations.csv`,
   `OptimizationCacheKey`) following the standard 4-step routine in
   `docs/develop.md`?

Happy to iterate on naming (`set_battery_first_priority` /
`set_grid_import_only_at_socmin` / something else) — mostly looking for
agreement on the constraint semantics before opening the PR.

## Related

- #795 / #796 (simultaneous discharge+import already addressed) — exact
  same problem class
- #553 (SOC inside horizon)
- #585 (SOC starting point)
- #751 (opportunistic charging)
- Discussion #381 (forecast-driven charge/discharge)

LesIT1:
Closes #834.

This adds the `set_battery_first_priority` flag that @twobeass proposed in #834. It's off by default, so nothing changes unless you turn it on.

Credit for the idea and the analysis is all @twobeass's. The issue had gone quiet for a few weeks waiting on a design decision, and we run almost exactly the setup it targets (AC-coupled battery, set_nodischarge_to_grid, flat tariff), so we picked it up and built it. Happy to hand it back or adjust if @twobeass wants to take it from here.

## The problem

On a flat (non time-of-use) tariff, "drain the battery before importing" and "interleave grid import with discharge" cost exactly the same, so the solver is free to plan grid imports while the battery is still nearly full. With a writeback controller that actually gets executed, that's not what a flat-tariff user wants: you'd rather use stored energy first and keep the battery low so it can absorb the next PV peak. soc_final doesn't fix it, since it constrains the endpoint of the horizon, not the path.

## The constraint

When the flag is on, grid import is forbidden in any timestep where the battery is still above its minimum SoC. It's a Big-M gate on a dedicated per-timestep binary, in the same spirit as the discharge/import coupling from #796:

- a new binary `battery_first_import_gate` (1 = import allowed in that slot),
- `current_stored_energy - min_energy - 1%*cap <= cap*(1 - gate)` so the gate can only open once the battery has reached its floor,
- `p_grid_pos <= max_from_grid * gate` so no import flows while the gate is shut.

The gate is deliberately separate from the existing grid-direction binary D. Reusing D would collide with the `E <= D` constraint from #796 (set_nodischarge_to_grid) and force the battery to stop discharging, which is the opposite of what we want. With a separate gate the battery still discharges to cover the load while import is held at zero.

## Heads-up: it's a hard constraint

It can make the optimization infeasible in a timestep where load minus PV exceeds the battery's maximum discharge power, since import would then be the only way to balance power and the flag forbids it. So it's really meant for setups where the battery discharge power can cover the load. That's called out in the param help and the docs, and there's a test pinning the infeasible behaviour so it's explicit.

If you'd prefer it never go infeasible, I'm happy to switch to a soft-penalty form (penalise import-while-charged in the objective instead of forbidding it). I went with the hard constraint to match the issue and the #796 precedent, but the soft version is a small change if you'd rather have that.

## Plumbing and tests

Added the flag through the standard places (associations.csv, config_defaults.json, param_definitions.json, openapi.json) plus the docs, default false. Runtime override via runtimeparams works (it's keyed off associations.csv like the sibling flags), so you can A/B it from an automation without restarting.

Two tests in test_optimization.py:
- `test_battery_first_priority_drains_before_import`: a night-time flat-tariff scenario where the unconstrained baseline provably imports while the battery is full, and the flag moves all that import to after the battery hits min. The total import energy is identical, only the timing changes.
- `test_battery_first_priority_infeasible_when_load_exceeds_discharge`: pins the documented infeasible edge.

Full suite stays green.

Suggested changelog line:
- feat: add set_battery_first_priority to drain the battery before grid import (closes #834)

## Summary by Sourcery

Add an optional battery-first priority mode that forces the optimizer to drain the battery before importing from the grid on flat tariffs.

New Features:
- Introduce the set_battery_first_priority configuration flag to prevent grid import while the battery state of charge is above its minimum.
- Add a per-timestep binary gate in the optimization model to control grid import based on available battery energy.

Enhancements:
- Wire the new battery-first priority flag through configuration defaults, parameter definitions, OpenAPI schema, and documentation, with the constraint documented as potentially making problems infeasible in some setups.

Tests:
- Add regression tests to verify that battery-first priority shifts grid import to after the battery reaches its minimum SoC without changing total import energy.
- Add a test that pins the expected infeasibility when load exceeds the battery's maximum discharge power under battery-first priority.


twobeass:
I tested the constraint behaviour against a real install (rather than installing
the PR branch in the add-on, which would have meant rebuilding the container — I
went with a code-equivalent shadow simulator instead, see below).

### Setup

- Kostal Plenticore Plus, 7.7 kWh usable battery, `set_nodischarge_to_grid: true`
- KSEM smart meter at the grid connection point as the truthful import/export
  counter (`sensor.ksem_ac_meter_total_watt_hours_imported / _exported`)
- Solcast PV forecast (`prognose_heute` + `prognose_morgen` + `prognose_tag_3`
  → 96 half-hour slots, 48 h horizon)
- Flat tariff 0,28 €/kWh import, 0,082 €/kWh export
- EMHASS master running in the add-on with current config:
  `costfun: profit`, `set_use_battery: true`, `set_nocharge_from_grid: true`,
  `set_nodischarge_to_grid: true`, `weight_battery_*: 0`, `delta_forecast_daily: 2`

### Shadow simulator

For ~10 days I've been running an in-house pyscript that simulates the
**exact strategy this PR implements as a hard constraint**: walk the 96 forecast
slots, charge from PV surplus first, discharge into the load while SOC > SOCmin,
import from the grid only after SOC has hit SOCmin. It exposes
`sensor.af_p_grid_forecast` / `_p_batt_forecast` / `_soc_batt_forecast` with the
same attribute schema as EMHASS so it can be plotted side-by-side. Math:

```python
if pv_w >= load_w:
    surplus = pv_w - load_w
    charge_kwh = min(slot_surplus, capacity_left / eta_ch, P_charge_max * Δt)
    grid_w = -(surplus - charge_w)   # negative = export
else:
    deficit = load_w - pv_w
    avail = (soc - SOC_min) * E_nom
    disch_kwh = min(slot_deficit, avail * eta_dis, P_discharge_max * Δt)
    grid_w = deficit - disch_w        # positive ⇒ soc has reached SOC_min
```

By construction `grid_w > 0  ⟺  soc ≤ SOC_min` — which is exactly the
semantics of the PR's `current_stored_energy - min_energy - 1%·cap ≤ cap·(1 - gate)`
gate. The 1 % tolerance in the PR maps to the floating-point floor we hit
in practice; I've not seen chatter at the SOC floor in the shadow
simulator, so the tolerance feels right.

### Live diff — EMHASS mainline vs. PR-equivalent (now, 2026-06-08T12:00 +0200, 100% SOC)

A sunny-day forecast, so absolute differences are small. The structural
behaviour is still visible:

| Metric over next 48 h         | EMHASS mainline | AF (≡ PR #957)  |
| ----------------------------- | --------------- | --------------- |
| Total grid import             | 0,07 kWh        | **0,00 kWh**    |
| Total grid export             | 20,98 kWh       | 24,00 kWh       |
| First import slot             | 06-08T17:30     | none in horizon |
| Phantom-import slots*         | 1               | 0               |

\* slots where SOC is still above SOCmin **and** `p_grid_pos > 50 W` in the plan.

Side-by-side excerpt around the only phantom slot:

```
slot          EMHASS mainline    AF (≡ PR #957)
17:00         -513 W (export)    -587 W (export)
17:30         +102 W (import) ← phantom; battery is at ~99 %
18:00            0 W (idle)         -46 W (idle)
18:30          +39 W (idle)            0 W (idle)
```

On a sunny day with 100 %-charged battery, EMHASS planning +102 W of grid
import at all is the structural artefact #834 describes. The flag from
this PR removes it. On overcast / winter / high-load days the same
mechanism produces multi-kWh phantom splits — I posted one of those traces
in #834.

### Edge cases I've hit empirically that this PR's tests pin

1. **`test_battery_first_priority_infeasible_when_load_exceeds_discharge`** —
   I trigger this same infeasibility a couple of times per week when an
   appliance pulls > `Pd_max` on the AC side. Documenting it as expected
   behaviour rather than a latent crash is right.

2. **`naive-mpc-optim` + `soc_final=SOCmin` + `set_nodischarge_to_grid: true`**
   also goes infeasible in our setup, but for a different reason: the
   forced discharge has nowhere to go. That's pre-existing, not introduced
   by this PR; just flagging it for anyone reaching for `soc_final` as an
   alternative — the issue thread has the trace.

### Test methodology call-out

The rising-cost trick in `test_battery_first_priority_drains_before_import`
(`unit_load_cost = 0.20 + 0.001·t`) is the right way to make the baseline's
"import early" optimum **unique** — without that, the no-flag run could
pick either order at zero cost difference and the test would be a tie. The
`assertGreater` on baseline `import-while-charged` guards future refactors
from silently making the test pass for the wrong reason. Both things I
wouldn't have thought to do in my own first sketch.

### On the soft-penalty alternative offered in the PR body

I'd argue against switching to it for two reasons:

1. The hard constraint matches the actual hardware semantics: with
   `set_nodischarge_to_grid: true` and an AC-coupled battery, "battery
   covers it or grid does" is the physical truth — there is no
   intermediate state that a penalty term would model better.
2. The infeasibility edge is detectable and clearly attributable
   (covered by the pinned test). A soft penalty hides this case behind
   plausible-looking but physically inconsistent plans (small import +
   small discharge mixed) at solver-tunable rates, which is exactly the
   behaviour the original issue complains about.

But it's @davidusb-geek's call on style — both are defensible.

### Summary

From the spec author / issue reporter: ✅ this implements what #834
asked for, with several details done better than my own sketch
(separate gate vs. reusing D, the rising-cost test, conditional binary
creation). Ready to merge from my side. Happy to test the merged
build against the production system once a release is cut.

LesIT1:
Thanks for putting it through a real install, that's a much better test than anything I could do from my side, and good to see the phantom-import slot drop out in your shadow run.

Agreed on keeping it a hard constraint. Your point that a soft penalty would hide the infeasibility behind mixed small-import/small-discharge plans is exactly why I left it hard. The soft option was only in the PR body in case David preferred it, but the hardware-truth argument is the stronger one, so I'll leave it as is unless he says otherwise.

I pushed one thing from the Sourcery pass: the infeasibility test now runs the same setup with the flag off and asserts it stays feasible, so the infeasible case is provably down to the constraint and not the scenario. I left the gate variable created only when the flag is on (Sourcery flagged that too) since it's only ever referenced under the same flag and after the no-battery early return, so there's no KeyError path. Thanks again, and I'll take you up on testing the merged build once there's a release.

davidusb-geek (approved):

PR #958, opened 2026-06-08
fix: keep sequence (list-power) deferrable loads active when operating_hours is 0 (closes #887)

### LINKED ISSUE #887: Sequence loads don't work if operating_hours is set to zero
**Describe the bug**
If you define a sequence load, it appears the optimizer is prevented from scheduling it unless its "set_operating_hours_of_deferred_loads" is nonzero, even though this parameter (as I understand it) is meaningless for sequence loads since their operating hours are defined by the length of the sequence. This results in an infeasible optimization.

**To Reproduce**

Create a deferrable load with "nominal_power_of_deferrable_loads" = [1000,1000]. Set its "operating_hours_of_each_deferrable_load" = 0. The optimizer will return an infeasible optimization. Set the operating hours to a positive value, and the optimization will succeed.

**Expected behavior**

Since operating hours does not apply to a sequence load, I expect its value to not have an effect on the outcome. (If, on the other hand, setting its value to zero is meant as a way to quickly disable the load, that functionality is broken.)

LesIT1:
Closes #887.

This is the sequence-load case from #887. To clear up the question on the issue about what "sequence" means here: it's a deferrable load whose `nominal_power_of_deferrable_loads[k]` is a list, i.e. a fixed power profile that runs once contiguously (only the start time is chosen). That's different from having several scalar loads. `[[1000, 1000]]` with `number_of_deferrable_loads = 1` is one sequence load; `[1000, 1000]` with `number_of_deferrable_loads = 2` is two normal loads, and there the `[4, 0]` form works fine, load1 just gets disabled.

The bug is specific to the sequence form. For a sequence load the runtime is fixed by the length of the sequence, so `operating_hours_of_each_deferrable_load` doesn't apply. The optimizer already knows this in one place: the energy/operating-hours constraint skips sequence loads (the `is_sequence_load` exemption in optimization.py). But the "load active" parameter pass didn't have the same exemption. A non-thermal load with operating hours 0 gets `param_load_active = 0`, which forces `p_deferrable <= 0`, while the sequence's own `sum(y) == 1` forces it to run with non-zero power. Those two can't both hold, so the whole optimization comes back infeasible.

## The fix

Treat a sequence load as active in that pass too, the same way thermal loads are always active, so `param_load_active` lines up with the energy constraint that already exempts sequence loads:

```python
is_sequence = isinstance(nominal_powers[k], list)
...
elif (has_operating_requirement or is_sequence) and not window_outside_horizon:
    param_load_active = 1
```

It stays window-aware (a load whose configured window is entirely outside the horizon is still deactivated), so nothing changes there.

The only behaviour change is for a sequence load with operating hours 0: it now runs its sequence instead of making the optimization infeasible. Scalar loads are untouched, so disabling a scalar load with operating_hours 0 (the `[4, 0]` example) still works exactly as before.

## Test

`test_sequence_load_runs_with_zero_operating_hours`: one sequence load `[[1000, 1000]]` with operating hours 0. It's infeasible on master and Optimal with the fix, and the sequence runs once (total 2000 W). I also checked the scalar-disable case, mixed scalar plus sequence, max-cost sequences (can still opt out), and sequences longer than the horizon, all fine. Full suite stays green.

Suggested changelog line:
- fix: sequence (list-power) deferrable loads no longer go infeasible when operating_hours is 0 (closes #887)

## Summary by Sourcery

Ensure sequence deferrable loads remain active even when operating_hours is set to 0 to avoid infeasible optimizations.

Bug Fixes:
- Prevent sequence (list-valued power) deferrable loads from being deactivated when operating_hours is 0, aligning load-activation logic with existing energy constraints to keep problems feasible.

Tests:
- Add a regression test verifying that a sequence deferrable load with zero operating_hours runs its full sequence once and yields an optimal solution.


LesIT1:
Rebased onto current master. The only conflict was in `test_optimization.py`, where today's #957 merge added the battery-first-priority tests right next to this PR's sequence-load test, so I kept both. No behaviour change, the sequence-load fix and the docs are unchanged, and tests pass. Mergeable again.

davidusb-geek (approved):

Issue #943, opened 2026-06-05
EMHASS 0.17.6: thermal_config causes deferrable load to remain at 0 W


renaatdb:
# Thermal deferrable load not scheduled in MPC (EMHASS 0.17.6)

## Describe the bug

I may have found an issue in EMHASS 0.17.6 related to thermal deferrable loads used with naive MPC.

### Environment

* EMHASS 0.17.6
* Python 3.12.13
* CVXPY 1.7.5
* Home Assistant OS
* EMHASS Add-on

### Setup

Deferrable loads:

* Deferrable load 0 = Geothermal heat pump
* Deferrable load 1 = Dishwasher
* Deferrable load 2 = DHW Boost
* Deferrable load 3 = Passive cooling (optional)

### Observed behaviour

When deferrable load 0 is configured as a normal deferrable load, EMHASS schedules it correctly.

Example:

* P_deferrable0 = 2200 W during periods with high PV production or low electricity prices.

However, when I add a thermal_config section to the exact same deferrable load, the optimizer no longer schedules it.

The optimization runs successfully and:

* predicted_temp_heater0 is published
* thermal model outputs are generated

But:

* P_deferrable0 remains 0 W for the entire horizon

### Runtime parameters

The following parameters are passed correctly:

* def_total_hours[0] = 3.0
* entity_id = sensor.emhass_p_deferrable0
* sense = "heat"
* method = "optim"
* start_temperature < target_temperature

Example:

```json
{
  "name": "Geothermische Warmtepomp",
  "entity_id": "sensor.emhass_p_deferrable0",
  "def_min_on_time": 2,
  "def_min_off_time": 4,
  "thermal_config": {
    "heating_rate": 0.25,
    "cooling_constant": 0.0008,
    "start_temperature": 22.0,
    "target_temperature": 23.0,
    "min_temperature": 22.0,
    "max_temperature": 25.5,
    "cop": 4.0,
    "sense": "heat",
    "method": "optim"
  }
}
```

### What I expected

I expected EMHASS to schedule P_deferrable0 similarly to the non-thermal version while respecting the thermal constraints.

### What actually happens

* Thermal outputs are generated.
* predicted_temp_heater0 is published.
* P_deferrable0 stays at 0 W for the entire optimization horizon.

### Additional observation

Removing thermal_config immediately restores normal scheduling of P_deferrable0.

This suggests the optimizer receives the deferrable load correctly, but the thermal model prevents actual power scheduling.

[...]

PR #959, opened 2026-06-08
feat: warn on unrecognized thermal_config keys (issue #943)

LesIT1:
Follow-up to #943.

That issue turned out not to be a bug. The reporter's thermal load scheduled 0 W because their `thermal_config` used singular keys (`min_temperature`, `max_temperature`, `target_temperature`) that the model doesn't read, so the load had no temperature bound and, with a tiny `cooling_constant`, never needed to heat. The model only reads the plural list keys `min_temperatures` / `max_temperatures` (the hard comfort band) and `desired_temperatures` (soft, used with `overshoot_temperature`).

What made it hard to spot is that EMHASS silently ignores unknown keys in `thermal_config`, so a typo gives you a flat 0 W schedule with no feedback. This adds a warning for any unrecognized key, with a hint pointing at the right list key for the common singular typos.

The model behaviour is unchanged, this is log-only. The warning fires once per load at build time, for example:

```
Deferrable load 0 thermal_config: unknown key 'min_temperature' is ignored; did you mean 'min_temperatures'? Temperature bounds use the list keys min_temperatures/max_temperatures.
```

Recognized keys: heating_rate, cooling_constant, start_temperature, min_temperatures, max_temperatures, desired_temperatures, overshoot_temperature, penalty_factor, sense.

Test `test_thermal_config_unknown_key_warns` checks the singular-typo case warns and suggests the plural key. Full suite stays green.

Suggested changelog line:
- feat: warn when a thermal_config contains an unrecognized key (issue #943)

## Summary by Sourcery

Warn on unrecognized thermal_config keys for deferrable loads to make configuration typos visible without changing optimization behavior.

New Features:
- Log warnings when a thermal_config contains unknown keys, with hints for common singular-to-plural key typos.

Tests:
- Add a regression test ensuring unknown thermal_config keys emit warnings and suggest the correct list-based keys.

### DIFF
--- a/src/emhass/optimization.py
+++ b/src/emhass/optimization.py
@@ -12,6 +12,36 @@ import pandas as pd
 
 from emhass import utils
 
+# Keys the thermal model actually reads from a load's thermal_config (issue #943).
+# Any other key is silently ignored, so a typo such as the singular
+# `min_temperature` (the model reads the list key `min_temperatures`) yields a
+# load that never schedules, with no feedback; we warn on unrecognized keys.
+THERMAL_CONFIG_KNOWN_KEYS = frozenset(
+    {
+        "heating_rate",
+        "cooling_constant",
+        "start_temperature",
+        "min_temperatures",
+        "max_temperatures",
+        "desired_temperatures",
+        "overshoot_temperature",
+        "penalty_factor",
+        "sense",
+    }
+)
+# Common singular typo -> (correct list key, what that key controls). The role
+# tailors the guidance so the hint for `desired_temperature` talks about the soft
+# target rather than the hard min/max comfort band.
+THERMAL_CONFIG_KEY_HINTS = {
+    "min_temperature": ("min_temperatures", "the hard comfort band"),
+    "max_temperature": ("max_temperatures", "the hard comfort band"),
+    "target_temperature": ("min_temperatures/max_temperatures", "the hard comfort band"),
+    "desired_temperature": (
+        "desired_temperatures",
+        "the soft target used with overshoot_temperature",
+    ),
+}
+
 
 class Optimization:
     r"""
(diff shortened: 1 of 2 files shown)


LesIT1:
Thanks @sourcery-ai, took both suggestions in 93ebc69:

- Hoisted the two key sets to module-level constants (`THERMAL_CONFIG_KNOWN_KEYS`, now a `frozenset`, and `THERMAL_CONFIG_KEY_HINTS`) so they're not rebuilt on every `_init_deferrable_load_params` call.
- Tailored the trailing guidance per hint. Each hint now carries the role of the key it points at, so the message reads `did you mean 'min_temperatures' (the hard comfort band)?` for the band keys and `did you mean 'desired_temperatures' (the soft target used with overshoot_temperature)?` for the soft one, instead of always mentioning the min/max band.

Behaviour is still log-only, 46 thermal tests + ruff stay green.

LesIT1:
Rebased onto current master to clear a conflict from today's #957 merge (its battery-first-priority tests sat next to this PR's thermal-config test in `test_optimization.py`, so I kept both). No behaviour change, tests green. Mergeable again.

davidusb-geek:
> Rebased onto current master to clear a conflict from today's #957 merge (its battery-first-priority tests sat next to this PR's thermal-config test in `test_optimization.py`, so I kept both). No behaviour change, tests green. Mergeable again.

This one has a conflict needing attention

LesIT1:
Resolved. Rebased onto current master, the conflict was just this PR's thermal-config test sitting next to the #958 sequence-load test in `test_optimization.py` after that merged, so I kept both. No behaviour change, full `test_optimization.py` passes (115). Mergeable again.

davidusb-geek (approved):

PR #961, opened 2026-06-08
feat(forecast): blend Solcast P50/P10 via weather_forecast_pv_quantile_bias

LesIT1:
## What this does

Solcast returns three PV estimates per period (P50 central, P10 low, P90 high), but EMHASS only reads P50 and discards P10/P90. This adds an opt-in `weather_forecast_pv_quantile_bias` parameter (`optim_conf`, float, default `0.0`) that blends the conservative P10 into the forecast:

```
estimate = bias * pv_estimate10 + (1 - bias) * pv_estimate
```

`0.0` (default) keeps the central P50, so existing behaviour is unchanged. `1.0` uses P10, intermediate values blend linearly. A higher bias makes the optimizer plan against a more cautious solar outlook, which helps when a solar shortfall bought back at peak grid rates costs more than holding a little extra battery reserve.

This is Phase 1 of #841. The idea is @lutorm's and the design is @purcell-lab's. It makes no change to the optimization or solver, the blend happens forecast-side on the values fed in. Phase 2 (scenario-based stochastic optimisation that consumes the quantiles) is the larger follow-up.

## Behaviour notes

- Solcast-only (the only method returning probabilistic estimates); no effect on `open-meteo`/`solar.forecast`/`csv`/`list`.
- Falls back to `pv_estimate` for any period missing `pv_estimate10`.
- Non-numeric or out-of-range values fall back to `0.0` (or clamp to `[0, 1]`) with a warning, so a quoted/templated config value doesn't silently misbehave.
- With the Solcast cache enabled the blended forecast is what's cached, so a changed bias applies on the next cache refresh (documented).

## Plumbing + docs

Added to `associations.csv`, `config_defaults.json`, `param_definitions.json`, `openapi.json` (regenerated via `scripts/generate_openapi.py`), and documented in `docs/config.md` and `docs/forecasts.md` with the formula, a worked example, and the rationale.

## Tests

- `test_get_weather_forecast_solcast_pv_quantile_bias`: asserts `bias=1.0` yields P10, `bias=0.5` the midpoint, and that unset == `bias=0.0` == the existing P50 output (backward-compat). Fails on master, passes here.
- `test_get_weather_forecast_solcast_pv_quantile_bias_invalid_inputs`: bool/string/NaN/out-of-range handling.

Full suite passes locally.

## Summary by Sourcery

Introduce an optional Solcast-specific bias parameter to blend PV P50 and P10 forecasts and make solar planning more conservative when desired.

New Features:
- Add weather_forecast_pv_quantile_bias configuration parameter to control blending between Solcast P50 and P10 PV estimates in the weather forecast pipeline.

Enhancements:
- Update Solcast forecast handling to safely coerce, validate, and clamp the bias value, falling back to P50 when invalid or missing.
- Extend configuration metadata and OpenAPI schema to expose the new bias parameter through associations, defaults, and API docs.

Documentation:
- Document the new Solcast PV quantile bias parameter, its formula, examples, and behavioural notes in the configuration and forecast documentation.

Tests:
- Add Solcast forecast tests to verify correct blending for various bias values and fallback behaviour when pv_estimate10 is missing.
- Add tests to ensure invalid or edge-case weather_forecast_pv_quantile_bias inputs are handled safely without crashes and with expected forecast ratios.


LesIT1:
Thanks @sourcery-ai, both addressed in b1ed584:

- The bias coerce/validate/clamp logic now lives in a small `_parse_pv_quantile_bias()` helper rather than inline in `_get_weather_solcast`, so the fetch method reads cleaner and the validation policy can be tested on its own.
- The two bias tests now share `_build_solcast_bias_payload`, `_setup_solcast_bias_env` and `_fetch_solcast_with_bias`, with the cache restore and param cleanup moved to `addCleanup` instead of a per-test try/finally.

No behaviour change. `test_forecast.py` is green and ruff is clean.

LesIT1:
Rebased onto current master to clear the conflict introduced by today's #956 merge (it touched the same Solcast helper area in `forecast.py`). The conflict was additive: #956's `_get_cached_forecast_or_none` and this PR's `_parse_pv_quantile_bias` are separate helper methods, so I kept both. No behaviour change, and the bias blend plus #956's cache recovery test green together. This also now includes the solcast-method guard from the #841 discussion (warns and ignores the knob if it's set for a non-solcast method). Should be mergeable again.

davidusb-geek (approved):

PR #955, opened 2026-06-08
feat: add optional weather covariates to the mlforecaster (#847)

BrettLynch123:
## Summary

This implements the **second option** from #847 — adding weather-covariate support to the **existing skforecast / sklearn `mlforecaster`**, rather than introducing a new heavyweight forecasting engine.

@davidusb-geek — following your steer on #847:

> I really think that the way to go is to build upon what's already there. Adding support for more input covariates to the current skforecast/sklearn combo is not a huge amount of work. [...] Would you be able to put this second option to work? So that we don't make a huge step towards another direction in the code architecture?

This is that. It is in-core, opt-in, adds **no new dependency**, and is available to add-on users (it only uses Open-Meteo, which EMHASS already fetches). It does **not** touch the existing PV/weather path, and with no covariates configured the `mlforecaster` behaves byte-for-byte as before. (This supersedes the Darts direction in #947 — happy to close that out separately.)

## What it does

The `mlforecaster` already feeds calendar features (hour/day/week/…) to `skforecast`'s `ForecasterRecursive` via `exog`. This PR lets you additionally feed **weather** columns as exog. Because the weather forecast spans the optimization horizon, they are valid *future-known covariates* for the recursive model — exactly like the calendar features.

- **New option `mlforecaster_weather_features`** — a list of columns to use. Empty (the default) ⇒ unchanged behaviour. Supported values come from Open-Meteo (`temp_air`, `relative_humidity`, `cloud_cover`, `wind_speed`, `ghi`, `direct_radiation`, `diffuse_radiation`, `precipitation`) plus two derived thermal-demand signals `heating_degree`/`cooling_degree` (computed from temperature, base 18 °C).
- **`Forecast.get_weather_covariates(index, features)`** — fetches Open-Meteo over a window spanning both the **training history** (via `past_days`, so no HA weather sensor is required) and the **forecast horizon**, and aligns the requested columns onto any index. It uses its own cache file (`cached-open-meteo-covariates.json`) so the existing PV/weather cache is untouched.
- **`MLForecaster`** gains a `weather_features` argument; the columns flow straight through `exog` in fit/tune, and `predict()` takes the horizon weather via a new `weather_future` argument. Models pickled before this change still load and predict (getattr fallback).

The orchestration (`forecast_model_fit` / `forecast_model_predict` / `_get_load_forecast_ml`) attaches the historical weather to the training data and supplies the future weather at predict time.

## Why — does it actually help?

[...]

PR #962, opened 2026-06-08
feat(mlforecaster): add backtest goodness-of-fit metrics and extract _build_weather_future helper

BrettLynch123:
## Summary

Follow-up to #955, implementing the two improvements @davidusb-geek suggested in review:

> "add the quality of fit to the attributes of the trained objects"
> (and a Sourcery code-quality flag on the duplicate horizon-construction block)

### 1. `backtest_metrics_` attribute on `MLForecaster`

When `perform_backtest=True`, the fitted object now exposes a `backtest_metrics_` dict with four standard goodness-of-fit metrics computed over the out-of-sample backtest folds:

| Key | Metric |
| --- | --- |
| `mae` | Mean Absolute Error (same unit as the target sensor, e.g. W) |
| `rmse` | Root Mean Squared Error (same unit) |
| `r2` | Coefficient of determination R² |
| `mape` | Mean Absolute Percentage Error (%; zero-actual rows excluded) |
| `n_samples` | Number of backtest steps used |

All four are also logged at INFO level after the backtest, making it easy to compare model variants or measure the effect of adding weather covariates without inspecting the object. `backtest_metrics_` is `None` before `fit()` and when `perform_backtest=False`, so the default behaviour is unchanged.

### 2. `_build_weather_future()` shared helper on `Forecast`

The future-weather-index construction block was duplicated verbatim in both `Forecast._get_load_forecast_ml` (forecast.py) and `command_line.forecast_model_predict` (command_line.py). Both call-sites now delegate to a new `Forecast._build_weather_future(data_last_window, mlf)` async method, eliminating the drift risk. Returns `None` when no weather features are configured or when `data_last_window` is `None`.

### Tests

5 new tests in `test_machine_learning_forecaster.py` for `backtest_metrics_` (keys, types, non-negative values, stays-None-by-default). 4 new tests in `test_forecast.py` for `_build_weather_future` (None-no-features, None-no-window, correct-horizon, uses-lags_opt-when-tuned). All targeted tests pass locally, ruff format + lint clean.

### Docs

`docs/mlforecaster.md` gains a *Backtest goodness-of-fit metrics* section documenting the attribute keys and units, and a *Future directions* section proposing Gaussian Process Regression as the path toward per-step confidence intervals and stochastic MPC — explicitly deferred as a large architecture change, per your suggestion.

### What this does NOT include

The GPR / stochastic-MPC direction is intentionally left as a future direction. Wiring a probabilistic regressor into the MILP and propagating uncertainty through the objective function is a large architecture change that warrants its own issue + design discussion when the time is right.

## Summary by Sourcery

Expose backtest goodness-of-fit metrics on MLForecaster and centralise future weather covariate construction for ML forecasts.

New Features:
- Add a backtest_metrics_ attribute on MLForecaster populated with MAE, RMSE, R2, MAPE and sample count when perform_backtest=True.

Enhancements:
- Factor out shared future weather covariate index construction into a Forecast._build_weather_future helper used by both forecast and command-line prediction paths.

Documentation:
- Document the MLForecaster backtest_metrics_ attribute and its metrics in the mlforecaster guide, and add a future directions section on probabilistic forecasting.

Tests:
- Add unit tests covering MLForecaster.backtest_metrics_ behaviour and Forecast._build_weather_future edge cases and horizon sizing.

### DIFF
--- a/docs/mlforecaster.md
+++ b/docs/mlforecaster.md
@@ -79,12 +79,29 @@ The results of the backtest will be shown in the logs:
 
     2023-02-20 22:05:36,825 - __main__ - INFO - Simple backtesting
     2023-02-20 22:06:32,162 - __main__ - INFO - Backtest R2 score: 0.5851552394233677
+    2023-02-20 22:06:32,163 - __main__ - INFO - Backtest metrics — MAE: 142.3456, RMSE: 198.2345, R2: 0.5852, MAPE: 12.34%
 
 So the mean backtest metric of our model is $R^2=0.59$. 
 
 Here is the graphic result of the backtesting routine:
 
 ![](./images/load_forecast_knn_bare_backtest.svg)
+
+### Backtest goodness-of-fit metrics
+
+When `perform_backtest=True` the fitted `MLForecaster` object exposes a `backtest_metrics_` attribute — a dict with the following keys computed over the out-of-sample backtest folds:
+
+| Key | Metric |
+| --- | --- |
+| `mae` | Mean Absolute Error (same unit as the target sensor, e.g. W) |
+| `rmse` | Root Mean Squared Error (same unit) |
+| `r2` | Coefficient of determination $R^2$ (dimensionless, higher is better) |
+| `mape` | Mean Absolute Percentage Error (%; rows where the actual value is zero are excluded) |
+| `n_samples` | Number of backtest steps used to compute the metrics |
+
+All four metrics are also printed to the log at INFO level after the backtest completes, making it easy to compare model variants or assess the effect of adding weather covariates.
+
+`backtest_metrics_` is `None` before `fit()` is called and when `perform_backtest=False` (the default).
 
 ## Adding weather covariates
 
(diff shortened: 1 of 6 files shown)


BrettLynch123:
Thanks for the review @sourcery-ai — addressed in commit 0f83954:

**1. Degenerate sample guard** (`machine_learning_forecaster.py` fit):
- Added `n_valid_samples = int(valid_mask.sum())` check before calling sklearn metrics.
- `n_valid == 0` → all metrics set to `float('nan')` + `WARNING` log (prevents `mean_absolute_error` / `r2_score` raising on empty arrays).
- `n_valid == 1` → `r2` set to `float('nan')` (undefined for a single sample), MAE/RMSE still computed.
- Also added `self.backtest_metrics_ = None` reset at the top of `fit()` so a reused instance never exposes stale metrics from a previous call.
- New test: `test_backtest_metrics_reset_on_refit_without_backtest`.

**2. Horizon index assertions** (`test_forecast.py` `test_build_weather_future_builds_correct_horizon`):
- Added `weather_future.index[0] == data_last_window.index[-1] + data_last_window.index.freq` assertion.
- Added `weather_future.index.freq == data_last_window.index.freq` assertion.
- Also hardened `_build_weather_future` itself to handle `index.freq is None` (falls back to `pd.infer_freq`; raises `ValueError` with a clear message if still unresolvable after inference).
- New test: `test_build_weather_future_raises_on_non_uniform_index`.

All targeted tests pass in `emhass-fast:test-highs-native`; ruff format + lint clean.

davidusb-geek (approved):

PR #964, opened 2026-06-09
feat: add battery SOC surplus cost (high-SOC dwell penalty)

LesIT1:
## What

Adds an optional high-SOC dwell penalty for the battery, the mirror of the existing `battery_soc_deficit_cost`. It penalizes the battery sitting above a configurable SOC threshold.

## Why

Closes the gap raised in #881. On a flat tariff nothing stops the optimizer charging to 100% as soon as energy is cheap or surplus, then holding the battery full for hours. That accelerates calendar aging (the HTW Berlin Energy Storage Inspection 2025 measured around 8 fewer hours at 100% SOC on sunny days with smarter charging) and exports the midday peak instead of soaking it up gradually. A soft cost on high SOC makes the battery fill more gradually into the solar peak and spend less time near full, and it works without a price spread.

## How

Exact mirror of the deficit cost:

- New optim_conf params `battery_soc_surplus_threshold` (default 0.90) and `battery_soc_surplus_cost` (default 0.0 = disabled).
- A nonneg slack var `soc_surplus_cost`, constrained to `>= (stored_energy - threshold_energy) * rate * timestep`, subtracted from the objective.
- New `soc_surplus_cost` result column.

## Behaviour change

None when off. `battery_soc_surplus_cost` defaults to 0.0, which gates out the constraint entirely, so existing configs are unchanged. The new test asserts the cost column is zero and SOC is unaffected when the cost is 0.

## Tests

New `test_battery_soc_surplus_cost`: with the cost off the optimizer dwells above the threshold to exploit a price spread; with the cost on it never dwells above the threshold. Full local suite passes.

## Docs

- advanced_math_model.md: new section with the formula and a runtimeparams example.
- config.md: documented both the new surplus params and the existing deficit pair (the deficit ones weren't listed there).
- openapi.json regenerated from param_definitions.json.

## Alternatives

I kept it as a hard mirror of the deficit cost for consistency. If you'd prefer a different default threshold or a different framing, happy to adjust.

## Summary by Sourcery

Add an optional high-SOC dwell penalty to the battery optimization model, mirrored from the existing SOC deficit cost, and expose its results in the optimization outputs.

New Features:
- Introduce configurable battery SOC surplus threshold and cost parameters to penalize time spent above a high SOC level in the optimization model.

Enhancements:
- Include SOC surplus cost as a decision variable and objective term, and return its per-timestep values in optimization results.

Documentation:
- Document the battery SOC surplus cost behaviour and configuration, and add missing documentation for the existing SOC deficit parameters, updating the API schema accordingly.

Tests:
- Add a regression test verifying that enabling the SOC surplus cost prevents the optimizer from dwelling above the configured SOC threshold and that the reported surplus penalty behaves as expected.


LesIT1:
Thanks. I used the non-negative slack plus a lower bound here on purpose, to keep this an exact mirror of the existing `battery_soc_deficit_cost` term (#859), which is built the same way. The solver drives `soc_surplus_cost` down to its bound, so it already evaluates to `max(0, (stored - threshold) * rate * dt)`: the non-negative variable supplies the `max(0, .)` and the constraint supplies the inner expression. That's the standard epigraph form and it keeps the constraint linear, whereas `cp.pos`/`cp.maximum` would add a nonlinearity node for no change in result and would diverge from the deficit term sitting right next to it. I'd rather keep the two mirrored for maintainability, but happy to switch both in a follow-up if you'd prefer that style across the board.

LesIT1:
Two good questions, on the inline comments.

On the threshold: gating off when `battery_soc_surplus_threshold` is 0 is deliberate, and it mirrors the existing `battery_soc_deficit_cost` (#859), which gates the same way. The on/off here is the cost (`battery_soc_surplus_cost`, default 0), not the threshold; the threshold just sets where the high-SoC band starts. A threshold of 0 would mean penalising any stored energy at all, which isn't what a high-SoC dwell penalty is for, so treating 0 as "disabled" is the sensible reading. I kept it identical to the deficit term on purpose. Happy to add a docs line spelling out that the cost is the switch and the threshold should be above 0.

On the fixed timestamp: fair in principle, but `pd.Timestamp.now(tz=...)` is the existing convention in this file (six tests build their index this way), and this test's outcome is driven entirely by the explicit price arrays (`[0.05, 0.05, 0.30, 0.30]`), not by the start time, so it isn't actually time- or DST-sensitive. I'd rather not make this one test diverge from the other five. If you'd like the whole file pinned to a fixed start for determinism, I'm happy to do that as a separate cleanup so it stays consistent.

davidusb-geek (approved):

Issue #623, opened 2025-11-24
[Feature Request] Multi‑Objective Optimization with Peak Power Consideration


matti-oss:
EMHASS provides a flexible framework for integrating forecasting and optimization into Home Assistant energy management workflows.
This proposal is a future topic of interest that could further expand EMHASS’s capabilities.

Motivation
The current EMHASS optimization framework focuses primarily on minimizing energy costs and maximizing self‑consumption. However, in many electricity markets, distribution companies apply additional charges based on peak power demand during a billing period. This means that even if total energy costs are minimized, a single high‑power spike can significantly increase consumer bills.

As highlighted in [Multi‑objective optimization of energy systems (ScienceDirect, 2020)](https://www.sciencedirect.com/science/article/pii/S2352484720312890), multi‑objective optimization approaches can balance competing goals such as cost minimization and peak load reduction. Incorporating this methodology into EMHASS would make the optimizer more grid‑friendly and economically robust.

Proposed Enhancement:
- Introduce multi‑objective optimization into EMHASS, where the solver considers both:
- Energy cost minimization (existing objective).
- Peak power reduction / load equalization (new objective).

A Pareto‑optimal solution set could be generated, allowing users to select trade‑offs between minimizing costs and reducing peak demand. This would:
- Equalize power loads across the optimization horizon.
- Reduce stress on the distribution grid.
- Lower peak‑related charges for consumers.

The user could control the operation using a relatively simple weighting rule, such as:

optimization_objectives:

- cost_weight: 0.7
- peak_weight: 0.3

Future impact
The support for grid‑friendly power management could become a positive differentiator for the EMHASS community.  EMHASS could proactively balance loads on the consumer side and directly address the arguments often used to justify forced peak power tariffs.

scruysberghs:
not sure how that would/could work. emhass optimisation window is typically 24 or 48 hrs while a billing period is monthly in most countries

max power from grid is already a constraint you can pass to emhass. You would have to calculate/predics the optimal valua for this once at the start of a billing period for your specific usage profile

Whatsonyourmind:
The cost-vs-peak multi-objective optimization has a clean mathematical formulation.

**Weighted-sum LP (simplest, when users set weights):**
```
minimize: w_cost * sum(price_t * power_t) + w_peak * max_power

subject to:
  power_t >= 0                          // non-negative consumption
  sum(power_t * dt) >= energy_required   // meet daily energy need
[...]

PR #966, opened 2026-06-09
feat: opt-in capacity / demand charge with a billing-period peak floor (#623)

LesIT1:
## What

Adds the opt-in capacity (demand) charge from #623, in two parts:

- `capacity_cost_per_kw` (config, default 0): a cost on the single highest grid-import power over the horizon, so the optimizer flattens import peaks. This is the single-pass MILP term we discussed in #623, not a Pareto / multi-objective approach.
- `current_period_peak` (runtime, naive-mpc only, default unset): the peak grid import already incurred this billing period. The planned peak is floored at this value, so the optimizer doesn't waste battery or deferrable flexibility shaving below a peak that's already locked in for the month.

## Why

Lots of tariffs bill a demand charge on the monthly peak kW, so flattening peaks has real value beyond energy arbitrage. The `current_period_peak` floor is the point @scruysberghs raised in the thread: optimizing for a daily peak is wasteful if a higher peak is already set for the month.

## How

Phase 1 (`capacity_cost_per_kw`): when the cost is > 0 the build adds one non-negative scalar `peak_import`, constrained `>= P_grid_pos` (the epigraph of the horizon peak), and prices it in the objective at `capacity_cost_per_kw * peak_import / 1000` (¤/kW; `peak_import` is in Watts). It's a power charge, so it isn't scaled by the timestep the way energy terms are. When the cost is 0 (default) no variable, constraint or term is created, so the build is byte-identical to before. The gate is part of the optimisation-cache structural hash, so toggling it 0↔>0 correctly forces a rebuild.

Phase 2 (`current_period_peak`): one extra affine constraint `peak_import >= current_period_peak` against a scalar `cp.Parameter`. No new variable, no objective change. The parameter is set per call, so changing the monthly peak between MPC ticks updates the value only, never forces a rebuild, and is DPP / warm-start safe. It lives in `passed_data`, not config, so it's cache-key neutral. Default unset / 0 makes the constraint redundant with the variable's non-negativity and the epigraph above, so the plan is identical to Phase 1.

It's a one-sided floor on a free epigraph variable, so it can never force extra import or make the problem infeasible. The capacity offset only ever surfaces in the "Total value of the Cost function" log line, never in a reported cost column (those are recomputed from the solved `P_grid_pos` / `P_grid_neg`).

## Behaviour change

None when off. `capacity_cost_per_kw` defaults to 0 (feature off, byte-identical build) and `current_period_peak` defaults to unset. A non-numeric, non-finite (NaN/inf) or negative value for either is ignored with a warning rather than crashing.

## Plumbing / docs

- Phase 1: `config_defaults.json`, `param_definitions.json`, `associations.csv`, regenerated `openapi.json`, `docs/config.md` (kept distinct from the TOU `load_peak_hours_cost`).
- Phase 2: `runtime_params.json` + `treat_runtimeparams` + the naive-mpc caller, documented in `passing_data.md` with a worked `rest_command` example. It's runtime-only and disjoint from config (enforced by `test_keys_disjoint_from_config`), so `openapi.json` doesn't change.

## Tests

- `capacity_cost_per_kw` shaves the peak with a positive cost; 0/unset is byte-identical; a numeric string is coerced; a non-finite/invalid value is ignored; the problem stays DPP.
- `current_period_peak` above the achievable peak leaves the plan equal to the no-capacity baseline (nothing left to shave); 0/unset reproduces Phase 1; a numeric string is coerced; a non-finite/invalid value falls back to 0.
- Full local suite passes.

## One thing to confirm (units)

The cost param is `capacity_cost_per_kw` (priced per kW), but the runtime floor `current_period_peak` is passed in Watts, to match the internal `p_grid_pos` / `peak_import` units (the runtime schema's unit enum has no kW, and keeping it in W avoids an extra in-code scaling). The docs disclose this and the HA example multiplies a kW sensor by 1000. If you'd rather `current_period_peak` accept kW for consistency with the `_per_kw` cost param, I can move the conversion to the boundary instead. Happy either way.

## Summary by Sourcery

Add an opt-in capacity/demand charge based on peak grid import, with optional runtime capping by the billing-period peak, and wire it through optimization, runtime plumbing, and documentation.

New Features:
- Introduce the configurable capacity_cost_per_kw parameter to apply a cost to the optimization horizon peak grid import and incentivize peak shaving.
- Add the runtime current_period_peak parameter so the capacity charge only applies above the already incurred billing-period peak when using naive MPC.

Enhancements:
- Ensure capacity_cost_per_kw and current_period_peak are defensively coerced and validated, treating invalid, negative, or non-finite values as disabled with warnings.
- Preserve DPP/warm-start behavior while integrating the new capacity charge and peak-flooring into the optimization problem.

Documentation:
- Document the new capacity_cost_per_kw configuration option and its relationship to existing TOU peak cost settings.
- Extend runtime data passing docs with guidance and examples for using current_period_peak in naive-mpc-optim.

Tests:
- Add optimization tests covering peak shaving behavior, zero vs unset semantics, string coercion and invalid-value handling for capacity_cost_per_kw.
- Add tests verifying current_period_peak semantics, including no-op behavior when the feature is off, coercion, invalid handling, and interaction with the capacity charge.


LesIT1:
Thanks. The first point is already handled, and I think the split you're describing is the one that's there. The structural gate and the per-call value are separated, and the gate is cache-keyed: `capacity_cost_per_kw` isn't in `optim_conf_runtime_keys`, so it's part of `optim_conf_structural_hash` (command_line.py:368, the catch-all that hashes every optim_conf key except the explicit runtime-safe allowlist). A 0 -> >0 change therefore misses the cache and rebuilds, which is exactly what creates `peak_import`. The only value that varies per call, `current_period_peak`, is the `cp.Parameter`, which deliberately doesn't affect structure. So changing the cost rebuilds; changing the incurred-peak floor doesn't. I double-checked the cache key to confirm.

On the tests, fair point that the battery + spike fixture repeats. I kept each one self-contained to match the existing `capacity_cost_per_kw` tests already in the file, which build the scenario inline too, so a reader sees the whole setup in one place. Happy to pull a shared builder out across all of them in a follow-up if you'd rather.

davidusb-geek:
@LesIT1 This one needs rebase and conflict solving

LesIT1:
@davidusb-geek  Rebased onto master and resolved the conflicts. They were all additive. In optimization.py the #964 surplus-cost term now sits just above this PR's capacity term in the objective, and the new config keys live alongside the surplus ones in config_defaults, associations, param_definitions and openapi. The one spot worth a note was utils.py: master's no-horizon branch now reads soc_target from the runtime params, so I kept that and added the current_period_peak line next to it. I regenerated openapi.json from source afterwards and it came back identical to the merge, so the committed spec still matches the generator. test_optimization.py is green here (121 passed) along with the capacity and current_period_peak tests. Should be clean to look at again.

davidusb-geek (approved):

PR #968, opened 2026-06-09
feat: schedule PV curtailment as late as possible (closes #342)

### LINKED ISSUE #342: [Feature Request] Schedule PV_curtailment later in the day
**Describe the bug**
PV_curtailment is often scheduled early in the day.

**To Reproduce**
Schedule data with a high level of PV_curtailment

**Expected behavior**
It would be better allocation of energy resources to schedule the curtailment as the last event in the day, after battery charging and allocation to all deferrable loads has been completed.

**Screenshots**
EMHASS-–-Home-Assistant (4)

image

 - Home Assistant OS

**Your hardware**
- OS: HA OS
- Architecture: amd64

**EMHASS installation type**
 - Add-on

LesIT1:
When `compute_curtailment` is enabled, `p_pv_curtailment` carries no cost of its own, so among equal-cost plans the solver places curtailment anywhere in the horizon, and it often lands early in the day. That skews the SoC forecast and feeds the plan flip-flopping between MPC runs that @TD944 described. This implements the option 1 tie-break you greenlit in #342.

The change adds a tiny time-decreasing penalty on curtailed power, active only when `compute_curtailment` is true. The weight is `EPS * (T - t) / T`, so early curtailment costs slightly more than late curtailment and the solver pushes it as far right as the other constraints allow. One small departure from the issue sketch: the weight is normalized by the horizon length, so the worst-case weight is exactly EPS per watt no matter how long the window is. Without that, a 288-step run would see weights 288 times larger than a short one.

EPS is 1e-7 per W. Typical tariff coefficients in the objective are 1e-5 to 1e-4 per W per step, so the tie-break sits two to three orders of magnitude below any real price signal. I probed near-tie scenarios where a tiny export price favours early curtailment: the real signal wins for any price difference above 0.0001 $/kWh, far below real tariffs. The cost-invariance test pins that real cost and total curtailed energy are unchanged with the term on or off. Long horizons were checked too (n up to 288, curtailment still lands in the latest feasible steps).

No new configuration parameter, per your note in the issue.

One related fix that validation surfaced: a pure continuous deferrable load (not semi-continuous, no min power, no startup penalty) skips the binary logic, and the existing `param_load_active` deactivation only binds the binaries. So an inactive continuous load (0 operating hours) is left as a free energy sink bounded only by nominal power. Today that is a latent degeneracy; with the tie-break it becomes deterministic, because dumping surplus into the disabled load is cheaper than booking curtailment. The fix is one constraint in the continuous branch, `p_deferrable[k] <= M * param_load_active[k]`, exact parity with what the binary branch already does. Thermal loads are unaffected since `param_load_active` is pinned to 1 for them, and loads whose window falls outside the horizon stay feasible because the energy constraint is already relaxed in the same cases. The relaxed LP fallback inherits the fix as well since it rebuilds through the same code path. Happy to split this into its own PR if you would rather keep them separate, but the tie-break should not ship without it.

Tests (there were no curtailment behaviour tests before):
- `test_curtailment_scheduled_late`: forced-curtailment scenario with temporal freedom (battery absorbs half the surplus, the rest must be curtailed somewhere); asserts the curtailment mass lands in the late half, and that negating the weight flips it early, so the test cannot pass by solver luck. Fails on master.
- `test_curtailment_tiebreak_cost_invariant`: same scenario with the term on vs off; real grid cost and total curtailed energy are identical.
- `test_curtailment_tiebreak_absent_when_disabled`: with `compute_curtailment` false the objective does not reference the curtailment variable at all.
- `test_continuous_deferrable_inactive_no_phantom_consumption`: regression test for the free-sink fix; fails without the new constraint (each disabled load was absorbing 250 W per step).

Docs: one sentence on the new behaviour in docs/config.md and in the param_definitions.json description, with openapi.json regenerated from it.

Full test_optimization.py passes (119), whole suite green locally (578 passed), ruff check and format clean.

The diagnosis and the option 1 design are @purcell-lab's (#342), handed over earlier this week. @TD944 and @tbrasser corroborated the behaviour from their setups. I run `compute_curtailment` daily on a hard-capped export setup, so I will be validating this on live data as soon as it lands.

## Summary by Sourcery

Prioritize scheduling PV curtailment as late as possible in the optimization horizon while fixing a deferrable-load degeneracy revealed by the new tie-break logic.

New Features:
- Introduce a small time-weighted tie-break term so that, when curtailment is enabled and economically neutral, PV curtailment is pushed to later timesteps in the horizon.

Bug Fixes:
- Prevent inactive pure-continuous deferrable loads from acting as free energy sinks by bounding their power with the load-activation parameter.

Enhancements:
- Ensure the PV curtailment tie-break term is only present in the objective when curtailment computation is enabled.

Documentation:
- Document the curtailment timing behaviour when compute_curtailment is enabled and propagate the change into the API parameter definitions.

Tests:
- Add scenario-based tests that verify late scheduling of curtailment, cost invariance of the tie-break term, absence of the tie-break when curtailment is disabled, and correct behaviour of inactive continuous deferrable loads.

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -201,7 +201,7 @@ Then the additional technical parameters:
 - `modules_per_string`: The number of modules per string. Defaults to 16. This parameter can be a list of items to enable the simulation of mixed orientation systems, for example, one east-facing array (azimuth=90) and one west-facing array (azimuth=270). 
 - `strings_per_inverter`: The number of used strings per inverter. Defaults to 1. This parameter can be a list of items to enable the simulation of mixed orientation systems, for example one east-facing array (azimuth=90) and one west-facing array (azimuth=270).
 - `inverter_is_hybrid`: Set to True to consider that the installation inverter is hybrid for PV and batteries (Default False).
-- `compute_curtailment`: Set to True to compute a special PV curtailment variable (Default False).
+- `compute_curtailment`: Set to True to compute a special PV curtailment variable (Default False). When enabled, curtailment that is cost-equivalent is scheduled as late as possible in the optimization horizon (issue #342).
 - `inverter_stress_cost`: The virtual penalty cost (in currency/kWh) applied if the inverter runs at its maximum nominal power (Recommended: 0.05 - 0.20).
 - `inverter_stress_segments`: The number of linear segments used to approximate the quadratic curve. Higher values are more accurate but increase computation slightly (Recommended: 10).
 
--- a/src/emhass/optimization.py
+++ b/src/emhass/optimization.py
@@ -42,6 +42,12 @@ THERMAL_CONFIG_KEY_HINTS = {
     ),
 }
 
+# Tie-break weight for PV curtailment timing (issue #342): must be far below
+# real tariff coefficients (~1e-4 $/W per step) yet large enough for the LP to
+# act on (above HiGHS dual feasibility tolerance once multiplied by realistic
+# curtailment powers in W).
+CURTAILMENT_TIEBREAK_EPS = 1e-7
+
 
 class Optimization:
     r"""
@@ -1286,6 +1292,20 @@ class Optimization:
         capacity_cost_per_kw = self._get_capacity_cost_per_kw()
         if capacity_cost_per_kw > 0 and "peak_import" in self.vars:
             objective_terms.append(-capacity_cost_per_kw * (self.vars["peak_import"] / 1000.0))
+
+        # Curtailment timing tie-break (issue #342). p_pv_curtailment carries no cost
+        # of its own, so among equal-cost optima the solver may curtail early in the
+        # horizon even when storing now and curtailing later is equally cheap. Add a
+        # tiny time-decreasing penalty so the latest feasible timesteps are preferred.
+        # The weight is normalized by the horizon length and the epsilon is orders of
+        # magnitude below any real tariff coefficient, so it breaks ties without ever
+        # flipping a real economic decision.
+        if self.plant_conf["compute_curtailment"]:
+            p_pv_curtailment = self.vars["p_pv_curtailment"]
+            tiebreak_weights = np.arange(self.num_timesteps, 0, -1) / self.num_timesteps
+            objective_terms.append(
+                -CURTAILMENT_TIEBREAK_EPS * cp.sum(cp.multiply(tiebreak_weights, p_pv_curtailment))
+            )
 
         # Sum all terms to create the final objective expression
         return cp.Maximize(cp.sum(objective_terms))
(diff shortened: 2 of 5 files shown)


LesIT1:
@sourcery-ai

Answering the Sourcery review:

**Tests mutating the global eps: partially taken.** I've switched the tests to `mock.patch.object` (e330662), so the mutation is scoped to a context manager and restoration is guaranteed, which is the standard idiom for exactly this. I didn't route the epsilon through the `Optimization` constructor though. It's a fixed tie-break constant, not a tuning knob, and the issue thread settled on no new setup parameters or configuration surface for this. Adding a constructor argument would create a public API for something users should never touch. The tests are the only place it's ever varied.

**Length-mismatch guard (both the overall comment and the inline one): declined, with reasoning.** `param_load_active` and `p_deferrable` are both sized from `number_of_deferrable_loads` in the same `__init__`, so a mismatch can't occur by construction. The `if k < len(self.param_load_active)` guard is copied verbatim from the binary branch's deactivation constraint a few lines up, where it already guards the same parameter the same way. Asserting in just this one branch while the binary branch keeps the defensive guard would leave the file inconsistent. If a stricter alignment check is wanted, it belongs in a file-wide pass over all the per-load parameter guards, and I'm happy to do that as a follow-up.

davidusb-geek:
@LesIT1 This one too, conflict on optimization.py

scruysberghs:
Also, make sure to (add a) test if a system is upgraded to this and the new parameter is not specificly set in settings or passed at runtime the behaviour remains as is.

LesIT1:
@davidusb-geek Rebased onto master. The only conflict was in optimization.py, in the objective function where #964's surplus-cost term landed right next to this PR's curtailment tie-break. I kept both. The module constant and the continuous-branch load-active fix weren't touched by the merge. test_optimization.py passes (120) including the four curtailment tests.

@scruysberghs on the upgrade-safety point: this PR doesn't add a new parameter. The tie-break only switches on when `compute_curtailment` is true, and that ships false by default (config_defaults.json and param_definitions.json), so a system that upgrades and changes nothing gets the same plan as before. `test_curtailment_tiebreak_absent_when_disabled` builds the optimization at that default and asserts `p_pv_curtailment` never even enters the objective, so the term is structurally gone rather than just numerically tiny. For anyone already running `compute_curtailment=true`, the only thing that moves is that equally-cheap curtailment shifts later in the day, and `test_curtailment_tiebreak_cost_invariant` pins that the real grid cost and the total curtailed energy are identical with the term on or off. If you were thinking of the capacity charge in #966, that one has the same guarantee: `capacity_cost_per_kw` defaults to 0 and `current_period_peak` defaults to unset, both with tests that the plan is byte-identical when they aren't set.

davidusb-geek (approved):

davidusb-geek:
> @davidusb-geek Rebased onto master. The only conflict was in optimization.py, in the objective function where #964's surplus-cost term landed right next to this PR's curtailment tie-break. I kept both. The module constant and the continuous-branch load-active fix weren't touched by the merge. test_optimization.py passes (120) including the four curtailment tests.

I think that when merging #966 a new conflict appeared in optimization.py

LesIT1:
@davidusb-geek i can investigate that and propose a fix if you'd like?

LesIT1:
@davidusb-geek Rebased again onto master now that #966 is in. Same story as last time: the only conflict was in optimization.py, in the objective function where #966's capacity charge term landed right next to this PR's curtailment tie-break. I kept both, capacity first then the tie-break. The module constant and the continuous-branch load-active fix were untouched by the merge.

Full test_optimization.py passes (125), with the four curtailment tests and the five capacity tests green side by side, so the two features coexist. ruff check and format are clean. Should be conflict-free now.

davidusb-geek (approved):

PR #973, opened 2026-06-11
feat: support arithmetic expressions in InfluxDB var_list entries

LesIT1:
## Support arithmetic expressions in InfluxDB `var_list` entries

This picks up #803 (idea by @dewi-ny-je). It lets an entry in the InfluxDB sensor list be a small arithmetic expression over several time series instead of a single sensor, for example:

```text
{{'sensor.power_a' - 'sensor.power_b' * 1000}}
```

Each quoted entity is queried separately and the operation is applied element-wise on the values for the matching timestamps. Supported operators are `+ - * / ** %`, unary `+`/`-` and numeric constants. This is handy when the quantity EMHASS needs is not stored as one sensor: a net power that is the difference of two meters, a unit conversion, and so on.

It carries forward @dewi-ny-je's proposal in #803 (which was opened as a not-yet-tested draft from his fork's `master`, so it could not be pushed to). He kindly blessed the handoff. This branch keeps his expression parser and safe evaluator, adds the time alignment he and David discussed, hardens the evaluator a little, and adds tests plus docs.

### Alignment (the main review point on #803)

I reproduced the behaviour against a real InfluxDB 1.8 container before deciding the fix. Two findings shaped it:

- `GROUP BY time() FILL(previous)` already returns every entity on an identical bucket grid over the query window, so the element-wise arithmetic is aligned by construction. A reindex to a shared grid is a no-op.
- The real gap is leading nulls when a sensor's most recent reading sits just outside the queried window (`WHERE time >= start` hides it, so `FILL(previous)` has nothing to carry). For two series that update at different phases this shows up as spurious leading-edge NaN in the result.

So expression entities are queried over a window padded one day earlier than the requested start, then sliced back to `[start, end)`. That lets `FILL(previous)` seed the first in-window buckets from the most recent prior reading. This follows @dewi-ny-je's leading-bucket-padding suggestion. A prior reading older than the lookback still leaves a short leading gap, which the downstream regression absorbs (documented).

### Behaviour and scope

- No new config parameter. `var_list` already exists; this only changes how an entry is interpreted.
- True no-op for existing setups: a `var_list` with no `{{ }}` entry behaves exactly as before. There is a test pinning that a plain missing sensor is still skipped (not a hard failure).
- One deliberate behaviour difference, confined to the new feature: if an entity referenced by an expression returns no data, the InfluxDB retrieval fails (an expression cannot be partially computed). Plain sensors are unchanged. The padding only applies to expression entities, so plain-sensor columns keep their existing values.

### Safety

Expressions are parsed with `ast` and walked against a strict node whitelist: only the arithmetic operators above, unary plus/minus and int/float constants are allowed; entity tokens resolve to their fetched Series. Anything else (function calls, attribute access, subscripts, comparisons, names that are not quoted entities, bool/str constants) is rejected, so a `var_list` entry cannot reach arbitrary code. Both operands of a power are magnitude-capped so a crafted `10 ** 1000000` or chained `(10 ** 1000) ** 100` cannot build a huge integer, and a pathological entry that still overflows or nests too deep (`OverflowError`, `RecursionError`, ...) fails the retrieval cleanly instead of propagating out and aborting the optimization.

### Validation

- Verified end to end against a real InfluxDB 1.8 container: the padded fetch fills the leading buckets, a mixed plain+expression `var_list` resolves, and a missing entity fails cleanly.
- `tests/test_retrieve_hass.py` passes, as does the full suite (595 passed, 22 xfailed).

### Tests added (`tests/test_retrieve_hass.py`)

- `test_influx_is_expression`, `test_influx_extract_expression_entities`, `test_influx_evaluate_expression` - the helper contracts (precedence, de-duplication, unknown token, malformed input, non-Series result).
- `test_influx_evaluate_expression_rejects_unsafe` - attribute access, subscript, comparison, boolean op, bool/str constants, function call, bare name, and the power-operand caps are all rejected; a reasonable exponent still works.
- `test_influx_expression_pads_and_slices_entity_fetch` - expression entities are fetched over the padded window and sliced back to `[start, end)`.
- `test_influx_expression_pathological_fails_soft` - a deeply nested or overflowing expression returns a clean failure instead of crashing.
- `test_get_data_influxdb_expression`, `_entity_cached`, `_mixed`, `_missing_entity` - the end-to-end paths with a mocked client, including single-query caching and the missing-entity hard fail.
- `test_get_data_influxdb_plain_missing_sensor_skipped` - a plain missing sensor is still skipped, preserving prior behaviour.

Docs: a new "Arithmetic expressions in the sensor list" subsection in `docs/passing_data.md`, including the alignment/lookback note and the InfluxDB 1.x off-grid-resolution caveat.

@dewi-ny-je if you get a chance to point this at your real InfluxDB setup that would be a great sanity check. Happy to adjust the lookback default, drop the `**`/`%` operators if you would rather keep the surface smaller, or split anything out.

## Summary by Sourcery

Add support for arithmetic expressions in InfluxDB var_list entries, safely evaluating multi-series expressions and integrating their results into data retrieval.

New Features:
- Allow InfluxDB var_list entries to be arithmetic expressions over multiple sensors using a {{ ... }} syntax, evaluating supported operations element-wise across aligned time series.

Enhancements:
- Introduce safe parsing and evaluation of InfluxDB arithmetic expressions with strict operator and constant whitelisting and protections against resource exhaustion.
- Align expression operands by querying referenced entities over a padded lookback window and slicing back to the requested range, while caching shared entities across expressions.

Documentation:
- Document how to use arithmetic expressions in InfluxDB-backed sensor lists, including alignment behavior, limitations, and failure modes.

Tests:
- Add comprehensive unit tests for expression detection, entity extraction, safe evaluation (including rejection of unsafe constructs), lookback padding behavior, error handling of pathological expressions, expression caching, mixed plain/expression var_lists, and differing failure behavior for expressions vs plain sensors.

### DIFF
--- a/docs/passing_data.md
+++ b/docs/passing_data.md
@@ -253,6 +253,27 @@ Then on the EMHASS configuration you need to set:
 Finally, if using the Add-on, you need to fill both "influxdb_password" and "influxdb_username" in the Add-on **Configuration** pane.
 If using the Docker standalone or legacy installation method, then you need to set these in the `secrets_emhass.yaml` file.
 
+### Arithmetic expressions in the sensor list
+
+When using InfluxDB, the entity id you would normally give to a sensor list parameter (such as `sensor_power_photovoltaics` or `sensor_power_load_no_var_loads`) can instead be an arithmetic expression that combines several InfluxDB time series. Wrap the expression in `{{ ... }}` and quote each entity id:
+
+```text
+{{'sensor.power_a' - 'sensor.power_b' * 1000}}
+```
+
+Each quoted entity is queried separately and the operation is applied element-wise on the values returned for the matching timestamps. The supported operators are `+`, `-`, `*`, `/`, `**`, `%`, unary `+`/`-` and numeric constants. Anything else (function calls, attribute access, comparisons, names that are not quoted entities, ...) is rejected, so an expression cannot run arbitrary code. An entity referenced more than once within expressions is only queried once. This is useful when the quantity EMHASS needs is not stored as a single sensor, for example a net power that is the difference of two meters, or a unit conversion.
+
+```{note}
+
+A few caveats apply because of how InfluxDB 1.x aggregates data:
+
+- Each series is bucketed with `GROUP BY time() FILL(previous)`, so every entity in an expression lands on the same time grid and the arithmetic stays aligned. For the leading buckets of the requested window, EMHASS queries each entity slightly earlier (by one day) so the first values can be forward-filled from the most recent prior reading; a sensor whose previous reading is older than that still starts with a short gap, which the downstream regression absorbs.
+- If a referenced sensor updates less often than once per day, that one-day lookback may not reach its previous reading and the expression starts with leading `NaN`. If that affects your setup, add the expression to `sensor_replace_zero` or `sensor_linear_interp` so the gap is filled before the optimization.
+- An entity used both as a plain sensor and inside an expression is queried twice (the plain entry over the requested window, the expression entity over the padded window), so its plain column and its value inside an expression can differ at the leading buckets.
+- InfluxDB 1.x cannot time-weight inside `GROUP BY`, so a source that updates more slowly than the optimization time step simply repeats its last value within each bucket. Keep this in mind when combining series that report at very different rates.
+- If any entity referenced by an expression returns no data, the whole InfluxDB retrieval fails (a plain sensor that returns no data is still skipped, as before).
+```
+
 
 ## Passing in secret parameters
 Secret parameters are passed differently, depending on which method you choose. Alternative options are also present for passing secrets, if you are running EMHASS separately from Home Assistant. _(I.e. not via EMHASS-Add-on)_ 
--- a/src/emhass/retrieve_hass.py
+++ b/src/emhass/retrieve_hass.py
@@ -866,9 +866,10 @@ class RetrieveHass:
 
         try:
             expression_result = self._evaluate_influx_expression(parsed_expression, series_mapping)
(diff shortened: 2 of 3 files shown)


LesIT1:
Thanks for the review. Pushed 40030e4 addressing both points.

1. TypeError in the soft-fail handler: done. `_build_influx_expression_df` now also catches `TypeError`, so a sensor that returned non-numeric (object dtype) values fails the single expression cleanly instead of aborting the whole retrieval.

2. Arithmetic-error test: added, with one adjustment worth flagging. A `Series / 0` does not raise in pandas, it yields `inf`, so `{{'sensor.a' / 0}}` would not actually exercise the `ArithmeticError` branch. A real `ZeroDivisionError` is only reachable when the division is between scalar constants (plain ints in the AST evaluator), so the new test uses `{{'sensor.a' + 1 / 0}}`. I also added a `TypeError` soft-fail test driven by an object-dtype series, which covers the new catch directly.

Both new tests assert the call returns None, mirroring the existing pathological cases. The full retrieve_hass suite passes and ruff is clean.

dewi-ny-je:
It won't be in a couple of days but I'll give it a test

davidusb-geek (approved):

dewi-ny-je:
I didn't/couldn't do much work but I'm happy I could give an idea someone picked up and implemented!

I'll see if I can adapt the docs to make the feature known to the users, because it could help new users to quickly get historical data

LesIT1:
@dewi-ny-je It was a very good idea, you did plenty, I am just glad I could help you get it over the line. Great idea on expanding the docs further, can never have too much doco(kinda).

dewi-ny-je:
#977

Issue #952, opened 2026-06-07
Feature request: optional minimum on-time / off-time per deferrable load


LesIT1:
A few common deferrable loads don't cope well with short-cycling, things like heat pumps, AC compressors, and pool or well pumps, where switching on and off rapidly causes wear and drops efficiency.

At the moment there are two extremes for how a deferrable runs. treat_deferrable_load_as_semi_cont lets the solver chop the run into lots of short bursts, and set_deferrable_load_single_constant forces it into one continuous block. There's nothing in between, like run in chunks of at least N steps.

What I'd suggest is an optional minimum_on_timesteps_of_each_deferrable_load (and maybe minimum_off_timesteps too), defaulting to 0 so it's off and nothing changes. When set, it adds the usual minimum-run-length constraints to the on/off binaries you already build, so once a load turns on it stays on for at least N steps.

No new dependencies, it's just CVXPY constraints on the existing binaries (the same ones the deferrable_load_max_cost work used). Off by default so existing plans are untouched, and it sits alongside semi_cont, single_constant and max_startups rather than adding anything new.

I'd start with minimum on-time since that's the half people usually ask for, and add off-time only if you'd want it. Open to this?

lutorm:
The startup cost is a "soft" way of doing this, right? But maybe there's a need for a hard floor on the runtime as well if you run very high rate optimizations.

LesIT1:
You're right that the startup cost is the soft version and it's the cheaper thing to reach for first. But it isn't the same as a guaranteed minimum runtime. A startup penalty discourages flapping, but the solver can still switch a load off one timestep after starting it if the prices line up, the penalty just has to be outweighed. A real min on-time / min off-time is a hard constraint: once on, the load stays on for N timesteps, which is what you want for compressors or pumps that physically shouldn't short-cycle.

So I think both have a place, startup cost for "prefer not to cycle" and min on/off-time for "must not cycle". The hard version is what I was after here. And you're right that it matters more at high optimization rates, since a per-start soft penalty gets relatively cheaper as you slice the timestep finer.

sokorn:
One thing to keep in mind with rolling windows and minumim runtime is, that you must substract the time, the load was running as you pass the runtime data. If you for example tell EMHASS to have a minimum runtime of 1h, but the load was already on for 30min, EMHASS will still plan 1h of remaining runtime, as it has no information of the past. It's a similar issue with the parameter for number of starts per deferrable load.

[...]

Issue #605, opened 2025-10-30
Feature Request: Enhance `def_current_state` to accept real power (W) and force this value for the first timestep


martinarva:
### Is your feature request related to a problem? Please describe.

Yes. The optimizer's main load forecast is based on `sensor_power_load_no_var_loads`. When a user manually starts a deferrable load (like an EV charger or pool heater), this sensor does **not** include the sensor_power_load_no_var_loads  power.

This means that for the *current* timestep (`i=0`), the `P_load[0]` value passed to the optimizer is **incorrect** (it's too low).

The current `def_current_state: [true, false, false]` parameter only prevents a *startup penalty*. It does **not** force the optimizer to plan the load as "ON," and critically, it does **not** correct the inaccurate `P_load[0]` value.

This can lead to two problems:
1.  The optimization for the current timestep is based on bad data (an incorrect power balance).
2.  The optimizer, not seeing the real-time load and seeking the lowest cost, may immediately plan to turn **off** the load that the user just manually turned **on**.

### Describe the solution you'd like

I propose enhancing `def_current_state` to accept a list of **numeric power values (in Watts)** instead of just booleans.

**Current (Boolean):**
`"def_current_state": [true, false, false]`

**Proposed (Numeric):**
`"def_current_state": [6820, 0, 0]`

This list would be populated by templates reading the *actual, real-time power* of the deferrable load entities. (A value of `0` would be treated the same as `False`).

This new numeric value should be used inside `optimization.py` in two ways:

1.  **Force the Power for Timestep 0:**
    Add a new constraint to force the deferrable power variable `P_deferrable[k][0]` to equal the provided value.
    * `if def_current_state[k] > 0:`
    * `P_deferrable[k][0] = def_current_state[k]` (e.g., `P_deferrable[0][0] = 6820`)

2.  **Implicitly Handle "ON" State:**
    * Forcing `P_deferrable[k][0]` to a positive value will implicitly require `P_def_bin2[k][0]` (the "ON" variable) to also be `1`.
    * The existing logic for `current_state` (which prevents the startup penalty) should be modified to check `if def_current_state[k] > 0` instead of `if def_current_state[k] is True`.

This change would perfectly "lock in" the real-world state for the current timestep. It fixes the inaccurate load forecast (by adding the deferrable load power back into the equation) and ensures the optimizer's plan aligns with the user's manual action for the first timestep.

### Describe alternatives you've considered

[...]

PR #980, opened 2026-06-14
feat: optional minimum on-time per deferrable load (#952)

LesIT1:
This adds an opt-in minimum on-time per deferrable load, the feature from #952 (thanks for the go-ahead). It is the hard-constraint complement to the existing soft startup penalty: once a load turns on it stays on for at least N consecutive timesteps, which is what compressors and pumps need to avoid short-cycling.

## What it does

- New config `def_minimum_on_time`: one integer per deferrable load, in timesteps. Default 0 disables it per load, so an unset or all-zero config is an exact no-op and existing plans are unchanged.
- It is a minimum up-time constraint on the on/off binaries you already build (`p_def_bin2` / `p_def_start`), with no new dependencies. It sits alongside `treat_deferrable_load_as_semi_cont`, `set_deferrable_load_single_constant` and `set_deferrable_max_startups` rather than replacing any of them.
- The primary target is `treat_deferrable_load_as_semi_cont` loads, where the binary forces full nominal power so min-on-time genuinely keeps the load running. For loads without semi_cont the constraint holds the on/off binary but does not pin a power level, and that is documented.

## Rolling-window state (the part #952 flagged as make-or-break)

A minimum on-time is only correct across MPC windows if the optimiser knows how much of it has already been served, otherwise every window re-plans a full N. To handle that there is an optional runtime input `def_current_on_timesteps`: the number of timesteps the load has already been on at the start of the horizon. When it is supplied for a currently-running load the first window enforces only the remainder (`max(0, N - elapsed)`); when it is not supplied it falls back to today's behaviour and only protects future starts inside the horizon. It is plumbed the same way as `def_current_state` (runtime param + treat_runtimeparams + cache allowlist).

This is the same runtime-state need that #605 is reaching for. I kept it as a separate opt-in input rather than overloading `def_current_state`, but I am happy to align the naming or shape with whatever you prefer for #605, since the two are clearly related.

## Behaviour changes worth flagging

- The existing force-on constraint `p_def_bin2[k] >= param_running_lb[k]` was moved out of the `set_deferrable_load_single_constant` branch so it now applies to all binary-logic non-sequence loads. The mask defaults to zeros, so for any load that is not being force-pinned this is `bin2 >= 0`, which is vacuous. `test_minimum_on_time_noop` asserts the default config produces the same plan as before.
- Minimum on-time is deliberately not applied to `set_deferrable_load_single_constant` loads. Those already run as one continuous block via their own currently-running pin, so a separate min-on would be redundant and could contradict their `sum(bin2) == required_timesteps` equality. There is a regression test for this.
- A start that cannot complete its minimum on-time before the load's operating window ends is simply not taken, so the problem stays Optimal. There is no forced infeasibility.

## Resilience

`def_minimum_on_time` is added to the stringly-typed xfail list in `test_config_value_resilience.py`, consistent with the other per-load deferrable params (operating_hours, nominal_power and so on) that do not yet have a #900 guard. It does raise a clear contextual error rather than a bare one.

## Tests

- `test_minimum_on_time_forward_red1`: a semi_cont load that the base solver short-cycles is kept on for at least N once started.
- `test_minimum_on_time_remainder_red2`: a currently-running load with elapsed time already served only owes the remainder.
- `test_minimum_on_time_noop`: default-off reproduces the existing plan.
- `test_minimum_on_time_feasibility_tight_window`: a start that cannot fit min-on in the window is not taken and the problem stays Optimal.
- `test_minimum_on_time_single_constant_currently_running_feasible`: single-constant plus min-on stays Optimal.

Docs are in config.md and differences.md, and openapi.json is regenerated.

Minimum off-time is the natural follow-up. I left it out of this PR to keep the change focused, and I am happy to add it next if the formulation here looks right to you.

## Summary by Sourcery

Add configurable minimum on-time constraints for deferrable loads, including runtime support for carrying over elapsed on-time and tests and docs to validate and describe the new behaviour.

New Features:
- Introduce per-load def_minimum_on_time configuration to enforce a minimum consecutive on-time for deferrable loads.
- Add def_current_on_timesteps runtime parameter to track elapsed on-time of currently running deferrable loads across optimization horizons.

Enhancements:
- Extend deferrable-load constraints to apply a shared force-on mask across single-constant pinning and minimum on-time remainder logic while keeping non-configured loads as a no-op.
- Ensure starts that cannot fulfill their minimum on-time within the operating window are simply avoided so optimization remains feasible.
- Wire def_minimum_on_time into configuration, hashing, OpenAPI, and parameter mappings so it participates correctly in caching and external integrations.

Documentation:
- Document def_minimum_on_time and its interaction with other deferrable-load options and runtime parameters in config and differences documentation.

Tests:
- Add regression and RED tests verifying minimum on-time enforcement, elapsed-on-time remainder handling, default no-op behaviour, feasibility with tight windows, and exclusion for single-constant loads.

Chores:
- Add def_minimum_on_time to the config resilience xfail list with contextual validation messaging.


LesIT1:
Thanks for the review. I pushed 2b3af977 addressing two of these:

- Shared the validation. `_coerce_min_on_timesteps` is now `_coerce_nonneg_timesteps(value, k, param_name)`, and `_update_def_current_on_timesteps_params` uses it too, so the integer and non-negative checks for both `def_minimum_on_time` and `def_current_on_timesteps` live in one place.
- The unused `n` in `_run_min_on_optim` now asserts the scenario dataframe length matches, so a misconfigured test fails fast.

On the ingest-time validation for `def_current_on_timesteps` in `utils`, I left that as is on purpose. It mirrors how `def_current_state` is handled: the parsing in utils stays light and the validation is centralised in the per-solve update method right before the values reach the solver. Adding a second validation pass at ingest would make this input inconsistent with its sibling, so I kept the single source of truth in the update method.

davidusb-geek (approved):

PR #981, opened 2026-06-14
fix: set_nodischarge_to_grid no longer blocks battery discharge to local load on AC-coupled systems (closes #936)

### LINKED ISSUE #936: set_nodischarge_to_grid over-constrains battery discharge since 0.17.3 (#796): E<=D forbids discharge-to-local-load during export-capable timesteps
### Summary

`set_nodischarge_to_grid` became much stricter in #796 (first released in 0.17.3) and now makes some previously-feasible setups infeasible. This is the root cause of #875. It surfaced from @g1za's report there, and I built the repro from his posted config.

### What #796 fixed (correctly)

#796 addressed #795: the old constraint let the optimizer discharge the battery to cover house load while exporting 100% of PV at the same time. For a hybrid inverter (Deye and similar) that prioritizes PV to load, that is physically impossible, so EMHASS plans desynced from reality, which mattered most during negative-price periods. That was a real problem and the PR was right to tackle it.

### The over-constraint

The old form was a power limit:

```python
p_grid_neg + p_pv >= 0   # only blocks exporting battery energy (export can't exceed PV)
```

#796 (commit 6901bef, `optimization.py:1307`) replaced it with a binary coupling:

```python
E <= D   # E=1 discharge, D=1 import: battery may only discharge while the grid is importing
```

The new rule does more than block battery-to-grid export. It also stops the battery from discharging to cover **local load** during any timestep where the grid is exporting (D=0). So whenever there is PV surplus going to the grid, the battery can't help with the house load in that timestep, even though none of that battery energy would actually reach the grid.

For a config that has to shed a lot of stored energy (`set_nodischarge_to_grid: true`, a large SOC to unload, e.g. `soc_init 1.0 -> soc_final 0.2`) and doesn't have many net-import timesteps, the battery has nowhere to put its energy and the solve goes infeasible. Non-hybrid (AC-coupled) systems are hit hardest, since for them battery discharge and PV export are physically independent to begin with.

The sibling `set_nocharge_from_grid` got the mirror change `D <= E` (`optimization.py:1303`) and likely has the same class of issue.

### Evidence

Bisect first-bad commit: `6901bef` (PR #796), first in 0.17.3.

From g1za's config alone (non-hybrid, `set_nodischarge_to_grid: true`, `soc_init 1.0`, `soc_final 0.2`, battery 27 kWh, 72h horizon at 15 min):

| scenario | v0.17.2 | v0.17.3 to v0.17.6 |
|---|---|---|
| PV ~7000 / load ~400, curtailment off | Optimal | Infeasible |
| PV ~9500 / load ~200, curtailment on | Optimal | Infeasible |

Trigger shape: `set_nodischarge_to_grid: true` plus a large SOC to shed and few import timesteps.

### Fix directions (for discussion, not a revert)

The goal is to keep #796's intent (no battery energy to the grid, PV prioritized to load on hybrids) without forbidding discharge to local load:

1. Restore the power-form export block (`p_grid_neg + p_pv >= 0`), which prevents battery energy reaching the grid without tying discharge to import-only timesteps.
2. Keep a binary formulation but allow battery-to-load during export, forbidding only battery-to-grid (constrain the battery's contribution to export rather than its discharge binary).
3. Apply the strict `E <= D` coupling only when `inverter_is_hybrid` is true, since PV-prioritized-to-load is a hybrid-inverter property and AC-coupled systems can discharge and export independently.

@torsteinelv, you wrote #796, so you'll know which form keeps the hybrid behavior you needed. @davidusb-geek for the call. Happy to send the PR once there's agreement on the form.

Root cause of #875.

LesIT1:
## The problem

`set_nodischarge_to_grid` became much stricter in #796 (first released in 0.17.3). The old power-form export block was replaced with a binary coupling `E <= D` (E = battery discharging, D = grid importing), so the battery may only discharge in timesteps where the grid is importing.

That does more than block battery-to-grid export. It also forbids the battery discharging to cover **local load** in any timestep where the grid is exporting (D = 0). For a config that has to shed a lot of stored energy (`set_nodischarge_to_grid: true`, a large SoC to unload, e.g. soc_init 1.0 to soc_final 0.2) with few net-import timesteps, the battery has nowhere to put its energy and the solve goes infeasible. AC-coupled (non-hybrid) systems are hit hardest, since for them battery discharge and PV export are physically independent. This is the root cause of #875.

#796 was right to tackle the original report (#795): on a hybrid inverter that prioritises PV to the load, the battery genuinely cannot discharge while PV is exporting, so the strict coupling matches reality there. This keeps that behaviour for hybrids and only loosens it where it is wrong.

## The fix (option 3 from the issue)

Gate the constraint on inverter topology, using the existing `inverter_is_hybrid` parameter:

- **Hybrid** (`inverter_is_hybrid: true`): keep `E <= D` unchanged. PV is prioritised to load, so the battery cannot discharge during export.
- **AC-coupled** (non-hybrid): bound grid export to the available PV instead. Battery energy can never reach the grid, but the battery may still discharge to local load while PV exports.

No new config parameter and no plumbing changes. The AC-coupled form mirrors the existing `set_nocharge_from_grid` constraint (`p_sto_neg + p_pv >= 0`).

## A second change, found while testing the first

While validating the AC-coupled branch I found a related leak when `compute_curtailment` is on. The export bound used the uncurtailed PV forecast `p_pv`, but the non-hybrid power balance injects `p_pv - p_pv_curtailment`. So when the solver curtails PV it could still export up to the full forecast and fill the gap from the battery, putting battery energy onto the grid. The bound now uses the same effective-PV expression as the power balance:

- non-hybrid + `compute_curtailment`: `p_grid_neg + p_pv - p_pv_curtailment >= 0`
- non-hybrid, no curtailment: `p_grid_neg + p_pv >= 0`

This only bites when curtailing, so the feasibility fix above is retained.

## Behaviour change

For AC-coupled users of `set_nodischarge_to_grid`, configs that were infeasible (or that could leak battery energy to the grid when curtailing) now solve correctly. Hybrid behaviour is byte-identical. Anyone not using `set_nodischarge_to_grid` is unaffected. Documented in docs/config.md.

## Alternatives

The issue lists two other forms: (1) restore the pure power-form export block for everyone, and (2) constrain the battery's contribution to export rather than its discharge binary. I went with option 3 (the topology gate) because it preserves #796's correct hybrid behaviour exactly while fixing the AC-coupled case, with the smallest surface. Happy to switch to another form if you and @torsteinelv prefer.

## Out of scope

`set_nocharge_from_grid` has the same uncurtailed-`p_pv` pattern in its existing constraint, but it is pre-existing and not part of #936, so I left it untouched. Worth a separate look if you want it tightened too.

## Tests

- `test_nodischarge_to_grid_ac_coupled_feasible_issue936`: non-hybrid with a large SoC to shed is now feasible (RED on the old `E <= D`, both the MIP and its LP relaxation infeasible).
- `test_nodischarge_to_grid_hybrid_still_blocks_discharge_issue936`: hybrid still applies `E <= D` and stays infeasible for the same shed, proving the hybrid branch is unchanged.
- `test_nodischarge_to_grid_curtailment_no_battery_export_issue936`: with curtailment on, no battery energy reaches the grid (RED before the curtailment-net fix).

Repro and the option list are from #936 (@OptimalNothing90); the hybrid intent is from #796 (@torsteinelv).

## Summary by Sourcery

Adjust battery discharge-to-grid constraints based on inverter topology to restore feasibility for AC-coupled systems while preserving existing hybrid behaviour, and tighten curtailment handling so battery energy cannot leak to the grid.

Bug Fixes:
- Allow AC-coupled systems using set_nodischarge_to_grid to discharge the battery to local load during PV export by constraining grid export to available PV instead of forcing E<=D.
- Fix a curtailment-related leak where, with compute_curtailment enabled, the grid export bound used uncurtailed PV, allowing battery energy to reach the grid despite set_nodischarge_to_grid.

Enhancements:
- Document the topology-dependent behaviour of set_nodischarge_to_grid in the configuration reference.

Tests:
- Add regression tests covering AC-coupled feasibility, unchanged hybrid behaviour, and absence of battery-to-grid export when curtailment is enabled under set_nodischarge_to_grid.


LesIT1:
Thanks Sourcery. Took the docs grammar fix (pushed in 9aede2eb).

On the two high-level points:

- The direct `plant_conf["inverter_is_hybrid"]` / `["compute_curtailment"]` access is deliberate. It matches how the surrounding code reads `plant_conf` (for example `_add_main_power_balance_constraints` uses `self.plant_conf["compute_curtailment"]` directly), and both keys ship in `config_defaults.json` so they are always present after config load. Switching only this branch to `.get()` would make it inconsistent with its neighbours without removing a real KeyError path. Happy to revisit if you'd want the whole module moved to defensive access.

- On the test duplication: the scenario construction is already shared through the `_make_nodischarge_scenario` helper. What is left per test is the `Optimization` build and the assertion, which I kept explicit so each topology case (AC-coupled feasible, hybrid blocked, curtailment no-export) reads on its own. I would rather keep them legible than fold the invocation into the helper, but can extract it if you prefer.

davidusb-geek:
Might need a rebase en solve conflicts?

LesIT1:
@davidusb-geek Rebased!

davidusb-geek (approved):

Issue #596, opened 2025-10-13
[Feature Request] Configure energy meters for deferrable loads


c0ldtech:
Right now one needs to supply an energy meter to EMHASS that does not contain data of the deferrable loads - which one needs to create as a virtual meter first and wait for it to have enough data so EMHASS can effectively use it. 
If one then wants to add another existing load as deferrable, the virtual meter needs to be adjusted and once again, the number of "historic_days_to_retrieve" needs to pass so appropriate data gets populated before the meter can be used by EMHASS.

It would be much easier and more flexible if the configuration would contain a sensor for the deferrable load itself so that EMHASS can subtract its values from the regular load meter. That way deferrable loads could be removed or added easily.

Issue #765, opened 2026-03-22
Starting deferable load again if it is already scheduled


han-alink:
If I set set_deferrable_max_startups to 1, a new MPC run will schedule a new start even if the deferable load is already running. That is probably not the intention of this (new) parameter.

Micr0mega:
I noticed this as well, worked around / hacked it by setting `set_deferrable_max_startups = 0.1` if the load is running, `1` otherwise. Works because 0.1 is not 0 (unlimited starts), but smaller than 1, so it schedules 0 starts ;-)

RikBast:
Indeed doesn't work the normal way, was trying also

davidusb-geek:
Solved by PR #982

PR #982, opened 2026-06-15
feat: optional def_current_power to pin a deferrable load's current power at t=0 (#605)

### LINKED ISSUE #605: Feature Request: Enhance `def_current_state` to accept real power (W) and force this value for the first timestep

LesIT1:
## What this does

Adds an opt-in runtime input `def_current_power` so the optimiser can be told the real power a deferrable load is drawing right now, at the first timestep of the horizon. This fixes #605.

When a deferrable load is started by hand (an EV charger, a pool heater), the main load sensor (`sensor_power_load_no_var_loads`) does not include it, so `P_load[0]` handed to the optimiser is too low. Two things go wrong: the t=0 power balance is built on bad data, and the optimiser, not seeing the load, often plans to switch off at t=0 exactly the thing the user just switched on. The existing `def_current_state` only suppresses the startup penalty. It does not force the load on and it does not correct the power value.

`def_current_power` is a per-load list of watts. When an entry is greater than 0 the optimiser, for that load:

1. pins the deferrable power at the first timestep to the supplied value, so the t=0 balance matches reality,
2. forces the load on at the first timestep, so it is not immediately switched off, and
3. treats the load as already running for startup detection, so no phantom startup penalty fires at t=0.

Default unset (or all zeros) is an exact no-op, so behaviour is unchanged for anyone who does not set it. The only behaviour change is when the new parameter is supplied.

## Why a separate parameter

The issue suggested overloading `def_current_state` to accept watt values instead of booleans. I went with a separate input instead so the binary on/off decision stays strict 0/1 and a fractional watt value can never weaken it. `def_current_state` is untouched.

## Scope by load type

The power pin only applies where the first-timestep power is a free variable: ordinary loads and loads with a minimum power. For `treat_deferrable_load_as_semi_cont` loads the power is strictly `nominal * bin`, so the pin is omitted and the force-on alone holds the load on at nominal (the supplied wattage acts only as an on/off signal, which is correct for an on/off device). Single-constant, sequence (list-valued nominal power) and thermal loads ignore the input. A currently-running single-constant load is already handled by `def_current_state`, which pins its remaining required timesteps, so injecting a below-nominal power there would fight the required-energy block.

For a load with a minimum power, the supplied value should sit in `[minimum_power, nominal]`. A value outside that range makes the solve infeasible rather than being silently clamped, so the input stays honest about what the hardware reported.

## How it is wired

`def_current_power` is a runtime-only input, plumbed like `def_current_state`: it lives in `runtime_params.json`, is handled in `treat_runtimeparams`, and is in the optim_conf runtime-key allowlist so a new value updates the pin on every MPC tick without rebuilding the solver (the value and an activation flag go through CVXPY Parameters, and the pin uses a parametric big-M that is a structural no-op when inactive). It composes with `def_current_on_timesteps` from #980 (how long a load has been running) by adding what power it is running at, so the optimiser can start from the real mid-run state at t=0. It is also a step towards #596 (per-deferrable metering), which is the standing-config counterpart of this runtime signal.

## Tests

In `tests/test_optimization.py`:
- pin forces `P_def[0]` to the supplied watts for an ordinary load (red against current master, which turns the load off at t=0),
- absent / all-zero input reproduces today's result (no-op, with a cost incentive that would otherwise turn the load off),
- semi-continuous load is forced on at nominal, not the partial value,
- phantom startup is suppressed at t=0,
- single-constant load ignores the input (dispatch identical to omitting it),
- window mask is widened when a load is reported running before its configured window,
- length-mismatch warning and invalid-value error mirror the sibling parameters.

Full suite green locally.

## Summary by Sourcery

Introduce an optional per-load runtime parameter to pin deferrable load power and on-state at the first timestep, aligning the optimizer with real-world load conditions while preserving existing behavior by default.

New Features:
- Add runtime optim_config parameter def_current_power to specify current per-load power at t=0 and influence deferrable load dispatch accordingly.

Enhancements:
- Wire def_current_power through optimization parameter updates and deferrable-load constraints to pin eligible loads' power at t=0, force them on, and suppress phantom startup penalties without rebuilding the solver cache.
- Extend window mask and running lower-bound logic so loads reported as running before their configured window remain feasible when def_current_power is used.
- Document def_current_power configuration, semantics, and interactions with different deferrable load types in the config reference.

Tests:
- Add optimization tests covering pin behavior, no-op semantics when absent or zero, semi-continuous handling, phantom-startup suppression, window-mask widening, single-constant no-op behavior, and validation of malformed def_current_power inputs.
- Add runtime parameter tests and wiring to ensure def_current_power is correctly parsed from runtime inputs and excluded from config hashing for cache invalidation.

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -117,6 +117,27 @@ Example:
   **Composes with `set_deferrable_max_startups`:** the two constraints are independent (count vs window) and coexist cleanly. A pathological combination of low max-startups, long min-on, and a short operating window can over-constrain a load; if you use both, ensure `max_startups x min_on_time <= available_timesteps`.
 
   **Initial-condition remainder:** to carry the remaining on-time across MPC ticks when the load is already running, also pass `def_current_on_timesteps` at runtime (see Passing data at runtime).
+- `def_current_power`: Per-load actual power (watts) each deferrable load is drawing right now at the start of the optimization horizon. **Default `[0, 0]` (or absent key) is an exact no-op** -- today's behaviour is preserved. One non-negative float per deferrable load, in the same order as `nominal_power_of_deferrable_loads`. Supply the real sensor reading in watts.
+
+  Unit: W (watts). Runtime-overridable on every MPC tick without rebuilding the solver cache.
+
+  When `def_current_power[k] > 0`, the optimizer applies three coordinated effects:
+
+  1. **Pin:** fixes `P_deferrable[k][t=0]` to the supplied power, correcting the t=0 power balance when the load is excluded from the main load sensor.
+  2. **Force ON:** forces the load's binary `bin2[0] = 1` at t=0, preventing the optimizer from immediately scheduling a stop.
+  3. **Suppress phantom startup:** sets the internal initial-state to ON so no startup penalty fires at t=0 (equivalent to also passing `def_current_state[k]=true`).
+
+  This is a separate runtime input rather than an extension of `def_current_state`, so the binary on/off decision stays strict 0/1 and a fractional watt value can never weaken it.
+
+  **Semi-continuous loads (`treat_deferrable_load_as_semi_cont: true`):** because semi-continuous power is strictly `nominal * bin` (cannot be an arbitrary fraction of nominal), the power pin is omitted for these loads. Supplying any non-zero value still triggers the force-ON and phantom-startup suppression, so the load is held ON at its full nominal power at t=0. The supplied wattage acts as an on/off signal only.
+
+  **Loads with minimum power (`minimum_power_of_deferrable_loads > 0`):** the supplied value should be at or above the configured minimum to avoid an infeasible t=0 power balance. Supplying a value below the minimum and above zero will produce an infeasible solve (the pin forces `p = supplied` while the min-power constraint forces `p >= min_power`).
+
+  **Supplied value exceeds nominal:** the existing window-mask upper bound (`p <= nominal * mask`) prevents values above nominal from being pinned; the solve will be infeasible. Supply values within `[min_power, nominal]`.
+
+  **Single-constant, sequence and thermal loads** ignore `def_current_power` (it is a no-op for those load types). A single-constant load runs as one fixed block, so its currently-running state is handled by `def_current_state` (which pins the remaining required timesteps); pinning a below-nominal power there would fight the required-energy target. Use `def_current_state` for a currently-running single-constant load.
+
+  **Typical MPC use:** read the load's power sensor at the start of each optimization tick and pass the reading in watts via the runtime API so the optimizer knows the load is running and at what level.
 - `deferrable_load_groups`: Define groups of deferrable loads that share a physical actuator (e.g. a heat pump serving both hot water and underfloor heating). Each group can enforce a shared power budget, mutual exclusion, or both. This is a list of group objects, each with the following fields:
 	- `names`: List of deferrable load names in the group (e.g. `["deferrable0", "deferrable1"]`).
 	- `max_power` *(optional when `mutual_exclusion` is `true`)*: Maximum combined power in Watts for all loads in the group at any timestep. Required when `mutual_exclusion` is `false`.
--- a/src/emhass/command_line.py
+++ b/src/emhass/command_line.py
@@ -270,6 +270,8 @@ class OptimizationCache:
             # Per-call elapsed on-time for min-on remainder (issue #952); value
             # is read via cp.Parameter so no rebuild on cache hit.
             "def_current_on_timesteps",
+            # Per-call current power in watts (issue #605); pin value is a cp.Parameter.
+            "def_current_power",
             "minimum_power_of_deferrable_loads",
             "cost_forecast_per_deferrable_load",
             # shared_thermal_tanks has its own structural hash field above
--- a/src/emhass/optimization.py
+++ b/src/emhass/optimization.py
@@ -371,6 +371,25 @@ class Optimization:
             cot.value = 0.0
             self.param_current_on_timesteps.append(cot)
 
+        # Current-power parameters (issue #605).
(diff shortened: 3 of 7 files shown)


LesIT1:
Thanks Sourcery.

Took the test fix: the phantom-startup check was guarded by `if "P_def_start_0" in res.columns`, which would skip silently if that debug column ever stopped being emitted. It now asserts the column is present first, so the check fails loudly instead (5c306c7a).

On the two high level points:

- Eligibility duplication: the affected / pin-eligibility is re-derived in both `_update_def_current_power_params` and the constraint builder on purpose. The load type has to be known at build time, to decide whether the pin constraints are added at all, and again at update time, to set the activation parameter to match what was baked in. This follows the existing min-on-time and single-constant handling, which re-derive the same way. Happy to pull it into a small shared helper if you'd rather have it centralised.
- `def_current_power` length in `treat_runtimeparams`: `n_loads` is used for the scalar broadcast case. A supplied list is passed through as-is rather than padded or truncated, matching `def_current_state` and `def_current_on_timesteps`. A wrong-length list is surfaced by the optimizer as a warning (missing entries default to 0, extras ignored) instead of being silently reshaped, so a user mistake stays visible.

davidusb-geek:
Also solves #765 right?

LesIT1:
@davidusb-geek yes, it does, through `def_current_state`.

#765 is the max-startups budget getting eaten by the in-progress run: with `set_deferrable_max_startups: 1`, a load that is already running registers a startup at t=0 on every fresh solve, so the one allowed start is spent before the horizon even begins.

The lever is the t=0 startup detection, which is already parameterised on master:

```python
p_def_start[k][0] >= p_def_bin2[k][0] - param_def_current_state[k]
...
cp.sum(p_def_start[k]) <= max_starts
```

When the running load is flagged (`def_current_state[k]=True`), `start[0]` is free to be 0, so the in-progress run no longer counts against `set_deferrable_max_startups`. So #765 is actually fixable today with `def_current_state` alone; the reporters hit it because they were not setting it (hence the `max_startups = 0.1` workaround).

What #982 adds is that `def_current_power[k] > 0` bumps `param_def_current_state[k]` internally, so a user who supplies the current power for a running load gets the same startup suppression for free, on the load types `def_current_power` applies to (ordinary / minimum-power / semi-continuous). For single-constant, sequence and thermal loads, which `def_current_power` deliberately skips, `def_current_state` is still the route.

The one caveat worth being explicit about: this is supervisor-driven, not inferred. EMHASS still has no memory across solves, so the running signal (`def_current_state` or `def_current_power`) has to be fed each tick by whatever drives the load. Same theme as #983.

Happy to bolt a small regression test onto this PR that pins it down: `set_deferrable_max_startups = 1`, load running, assert that flagging it (via either `def_current_state` or `def_current_power`) keeps the solve feasible and spends zero of the startup budget, with the RED being the same case left unflagged. Plus a line in the docs mapping `def_current_state` / `def_current_power` to #765 so the next person finds it, and then #765 can close when this merges. Want me to add those?

davidusb-geek (approved):

Issue #983, opened 2026-06-15
Carry over completed operating hours for must-run deferrables across the rolling horizon (#952 follow-on)


LesIT1:
@davidusb-geek following on from the min-on-time work in #980 and the t=0 pinning in #982, I hit a related gap while running a must-run deferrable live, and I think it is the remaining piece of the rolling-horizon carryover story from #952. I wanted to lay it out and check you are happy for me to PR it.

## The problem

In naive-MPC the optimiser re-plans every cycle, and the required run time for a deferrable is rebuilt from `operating_hours` on every solve with no memory of how much the load has already run today. So a load that must run N hours per day keeps being planned for the full N hours on each cycle.

For a flexible load that is free to fragment this is mostly harmless. For a must-run load pinned to a daily window it is not: as the day advances and the full block no longer fits the remaining window, the optimiser pushes it later and eventually rolls the whole block to the next day, so the load under-runs its daily target.

## Worked example (real)

I run a pool filtration pump as a deferrable: must-run, about 4 hours per day in winter, `set_deferrable_load_single_constant: true`, pinned to a 09:00 to 15:00 cheap/solar window, 15-minute MPC. Over one day I saw:

- morning solves: a clean single 4-hour block inside the window
- once the clock passed the point where 4 hours still fit before 15:00, each solve moved the block later, then rolled the whole block to tomorrow
- net result: the pump ran a couple of short fragments totalling about 75 minutes instead of 4 hours, and the published plan showed the full block sitting on the next day

The fragments come from the block being re-placed between solves; the under-run comes from there being no "I have already done X of my N hours today" signal, so once the remaining window cannot hold the full N, the optimiser abandons today.

## Why #980 and #982 do not cover this

I checked both against current master (they are mine, so this is just mapping what they actually do):

- `def_current_on_timesteps` (#980) computes `remaining = max(0, def_minimum_on_time - elapsed)` and only uses it to force the tail of an in-progress min-on-time window. It does not touch `param_required_timesteps` or `target_energy`, and the constraint that consumes it is explicitly gated `not is_single_const`. So it carries over the min-on tail, not the daily hours, and not for single-constant loads.
- `def_current_power` (#982) pins `P_def[k][0]` and forces the load on at the first timestep for a hand-started load. It is a single-timestep signal and is also excluded for single-constant loads.
[...]

PR #987, opened 2026-06-17
feat: optional minimum off-time per deferrable load (#952)

LesIT1:
This adds the off-time half of #952, the sibling of the minimum on-time that landed in #980. Some deferrable loads cope badly with short-cycling: heat pumps, AC compressors, and pool or well pumps lose efficiency and wear out when they restart too soon after stopping. A startup penalty discourages that softly, but it can always be outweighed by the right prices. This is the hard version: once a load turns off it stays off for at least N timesteps.

Like the on-time side it is fully opt-in. Default 0 is an exact no-op, and when no load sets it the solver problem is byte-identical to today, with no new variables and no new constraints.

## What it adds

- `def_minimum_off_time` (config, runtime-overridable, one int per load, unit timesteps, default `[0, 0]`): once a load stops it must stay off for at least N consecutive timesteps.
- `def_current_off_timesteps` (runtime-only): how many timesteps each load has already been off at the start of the horizon. This is the part @sokorn flagged on the issue, and it is what makes a minimum off-time correct under a rolling MPC. The first window enforces only the remainder, `max(0, N - elapsed)`, so a load that has already been off for a while is not held off for a fresh full N every tick. If the key is absent there is no initial force (it is not assumed zero), so existing setups are unaffected. Plumbed the same way as `def_current_on_timesteps` (runtime param + treat_runtimeparams + cache allowlist).

## How the constraint works

Min-on reused the existing rising-edge `p_def_start` binary. Off-time needs the falling edge, so this introduces `p_def_stop[k]`, created only for loads with `def_minimum_off_time > 0`. It is pinned exactly to the falling edge with three constraints rather than just a lower bound:

```
stop[t] >= bin2[t-1] - bin2[t]   # fires on a real stop
stop[t] <= bin2[t-1]             # only if it was on
stop[t] <= 1 - bin2[t]           # only if it is now off
```

The two upper bounds matter. A stop variable with no objective cost would otherwise be a free degree of freedom, and in a price tie the solver could set `stop[t] = 1` with no real falling edge and force a spurious off window. Pinning `stop[t] = 1` to exactly when there is a falling edge removes that. The forward constraint is then symmetric to min-on: `sum(1 - bin2[t:window_end]) >= (window_end - t) * stop[t]`, with the window clamped to the horizon end so a stop near the end stays feasible (the solver just will not take it if the off window cannot fit). The initial-condition remainder forces the first `remaining` steps off through a new force-off mask `param_running_ub` (default 1.0, so a no-op unless set).

## Scope and exclusions

Same gating as min-on: `set_deferrable_load_single_constant` and sequence loads are excluded, since single-constant already runs as one block and sequence loads are shaped by convolution rather than the on/off binary. It composes with `def_minimum_on_time` and `set_deferrable_max_startups` (independent constraints). No new dependency, just CVXPY constraints on the existing binaries plus the one new stop binary.

## Tests

In `tests/test_optimization.py`, mirroring the min-on tests:

- forward (RED): the base solver short-cycles off; with N=3 every interior off run is at least 3, with an anti-vacuity guard so the assertion cannot be skipped silently.
- remainder (RED): elapsed off-time forces the first `remaining` steps off; with no elapsed supplied t=0 is free.
- no-op: an all-zero `def_minimum_off_time` reproduces the no-key baseline plan exactly.
- feasibility: a too-tight window keeps the problem Optimal (the load just does not start).
- single-constant exclusion: a running single-constant load stays Optimal.

Full suite green locally (615 passed, 28 xfailed). Docs are in config.md and differences.md, param help is in param_definitions.json, and openapi.json is regenerated.

## Notes

This pairs with #983 (carrying completed operating hours across the horizon), which is the same idea applied to total run time rather than a single on/off window. I have kept that separate. Happy to adjust the formulation or the param names if you would shape it differently.

## Summary by Sourcery

Add configurable minimum off-time constraints for deferrable loads to prevent short-cycling, including support for rolling-horizon remainder handling and corresponding configuration, runtime parameters, and documentation updates.

New Features:
- Introduce per-load def_minimum_off_time configuration to enforce a minimum number of timesteps a deferrable load must remain off after stopping.
- Allow passing def_current_off_timesteps at runtime so remaining minimum off-time is correctly enforced across MPC horizons.
- Add force-off masking for deferrable load binaries to implement minimum off-time remainders without altering existing behaviour when disabled.

Enhancements:
- Extend deferrable load optimization to track current-off timesteps and add falling-edge decision variables for loads with minimum off-time enabled.
- Update configuration hashing, defaults, and parameter definitions to recognize the new minimum off-time and elapsed-off-time parameters.

Documentation:
- Document the new def_minimum_off_time and def_current_off_timesteps options, their units, interactions with other constraints, and behaviour at horizon boundaries in the configuration and differences documentation.

Tests:
- Add optimization tests covering forward minimum off-time enforcement, elapsed off-time remainder behaviour, default no-op behaviour, feasibility under tight windows, and exclusion for single-constant loads.


LesIT1:
Thanks Sourcery, two fair questions. Here is the thinking on each.

On moving `p_def_stop[k]` into `_initialize_decision_variables`: the list and its `vars_dict` entry are declared there, but the actual `cp.Variable` is created lazily in the constraint pass, and that part is deliberate. `p_def_start`, `p_def_bin1` and `p_def_bin2` are created unconditionally because they always exist, whereas a stop binary is only needed when a load sets `def_minimum_off_time > 0`. Creating it for every load would add a boolean variable per timestep to every existing user's MILP even when the feature is off, which breaks the exact no-op the default is meant to be (the no-op test asserts a byte-identical plan, and that holds precisely because no stop variable is created when the feature is unused). Creating it eagerly in the init method would also mean recomputing the same per-load gating (min-off > 0, not single-constant, not sequence) in two places, which is the divergence risk your second point is about. Keeping the instantiation next to the gating that decides it keeps that logic in one spot.

On `is_sequence_load_rem`: fair point. It is the same expression as the `is_sequence_load` used in the forward constraint, just re-derived because the remainder block runs in a different method where that flag is not in scope. It also mirrors how the merged min-on-time code re-derives its `is_single_const` in the same remainder block, so the two paths stay consistent with each other today. I would rather not add a one-off helper only on the off-time side, since that would make it inconsistent with the on-time path right next to it. If you would like the gating defined once everywhere, I am happy to follow up with a small shared `_is_sequence_load(k)` helper applied across both the on and off paths as a separate tidy-up, rather than folding a refactor of the already-merged code into this PR.

davidusb-geek (approved):

PR #990, opened 2026-06-17
docs(cookbook): add EV recipe - EMHASS as planner, evcc as executor

LesIT1:
Adds the first EV recipe to the cookbook, plus a link from the cookbook index.

It documents the setup where EMHASS stays the master MPC planner (the EV is one deferrable in the whole-home optimisation) and evcc only executes the plan on the car: reads live SoC, holds the charge current in the charger's safe range, speaks the vehicle API or OCPP, keeps the car awake. EMHASS does the time-shifting, evcc actuates. This is the opposite coupling to evcc-io/evcc#29815 (evcc calling EMHASS as its backend optimiser), so it does not depend on that landing and works today on released evcc and EMHASS. The index EV section is updated to say that, rather than gating all EV recipes on #29815.

You greenlit this one as the first cookbook addition in discussion #824. It is written against a live deployment running exactly this setup.

Notes:
- Docs only. No code, no new params, openapi unchanged.
- Step 5 (handling a manual mid-window charge) uses `def_current_power`, which is merged on master via #982 / #605 but not in v0.17.7 yet. The prerequisite and a caveat both say it needs a release after v0.17.7. Steps 1-4 work on any current release.
- The runtime payload uses `def_total_hours`, which maps to `operating_hours_of_each_deferrable_load` via associations.csv, the same as the EV study case.

## Summary by Sourcery

Add a cookbook recipe describing how to use EMHASS as the EV charging planner with evcc as the executor, and link it from the EV section of the cookbook index.

Documentation:
- Document an end-to-end EV charging setup where EMHASS performs whole-home optimization and evcc executes the EV charging plan, including prerequisites, runtime parameters, HA/evcc wiring, and caveats.
- Update the cookbook index EV section to link to the new recipe and clarify how this EMHASS–evcc coupling relates to the alternative evcc-as-frontend architecture discussion.

### DIFF
--- a/docs/cookbook/ev_evcc_executor.md
+++ b/docs/cookbook/ev_evcc_executor.md
@@ -1,0 +1,212 @@
+# EV charging with EMHASS as planner and evcc as executor
+
+## Goal
+
+Let EMHASS decide when and how much to charge an EV, as one deferrable load in the whole-home
+optimization, and let **evcc** carry that plan out on the vehicle. EMHASS does the time-shifting:
+which slots to charge in, against the tariff and the PV forecast. evcc is the vehicle-aware executor:
+it talks to the car properly, reads live SOC, holds the charge current inside the charger's safe
+limits, speaks the vehicle API or OCPP, and keeps the car awake. EMHASS stays the planning brain and
+evcc is the actuator. The two never fight over the plan.
+
+> This is the **opposite** coupling to
+> [evcc-io/evcc#29815](https://github.com/evcc-io/evcc/discussions/29815), which proposes evcc
+> calling EMHASS as its optimizer backend. Here evcc never optimizes the EV. It only carries out
+> EMHASS's per-timestep decision. The two recipes can coexist later. This one works today with
+> released evcc and EMHASS.
+
+## Prerequisites
+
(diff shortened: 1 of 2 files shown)


LesIT1:
﻿Thanks. Took the first two overall comments and both nits:

- Added a caveat on securing the EMHASS and evcc control endpoints (keep them on the LAN or behind a reverse proxy; evcc's loadpoint API has no auth of its own).
- Added a note that `sensor.p_deferrable2` follows the deferrable order in Step 1 and needs adjusting if that order changes.
- "study-case" -> "study case" for consistency with the referenced doc.

On the source-location comments, I'm keeping the `file:line` citations. The cookbook's own `_template.md` requires every `<!-- source: -->` comment to cite a real line number and a local check enforces it, so dropping the numbers would break the contributor rules. They already name the function next to the line (`treat_runtimeparams`), so if a refactor moves the line the function pointer is still there to re-anchor on.

davidusb-geek (approved):

scruysberghs:
> * evcc's loadpoint API has no auth of its own

I saw in the release notes recently evcc has an optional long lived access token setup now

Issue #898, opened 2026-05-27
Thermal cooling (#888): sense: null crashes optimization; sense=cool with positive heating_rate warms instead of cools


OptimalNothing90:
Two issues in the cooling support from #888, both verified with repros. Same null-handling thread as #876 / #878.

**1. `sense: null` raises ValueError (crash).** When `sense` is present-but-null in a thermal_battery or thermal_config block, `hc.get("sense", "heat")` returns `None` (the default only applies when the key is absent). `normalize_heat_cool_mode(None)` does `str(None).strip().lower()` = `"none"`, which isn't in `{"heat","cool"}`:

```
ValueError: thermal_battery: invalid sense "None". Expected "heat" or "cool".
```

Three call sites: `utils.py:456`, `optimization.py:1441`, `optimization.py:1651`. The HA add-on UI can persist an unset optional field as null (same shape as #876). Fix: `hc.get("sense") or "heat"` so a present-but-null value collapses to the default, matching the fallback convention from #878 / #879.

**2. `sense: cool` with positive `heating_rate` warms instead of cools.** In `_add_thermal_load_constraints`, `sense_coeff` is applied to the penalty terms but not to the temperature-dynamics constraint (`optimization.py:~1462`):

```python
predicted_temp[1+L:] == predicted_temp[L:-1]
  + (p_deferrable[:-1-L] * heat_factor)   # sense_coeff missing
  - (cool_factor * (predicted_temp[L:-1] - outdoor_temp[L:-1]))
```

`_add_thermal_battery_constraints` (`~1907`) applies `sense_coeff * q_input` correctly, so the fix pattern is in the same file. The existing cooling test uses negative `heating_rate` as a workaround, so it passes without exercising `sense=cool` + positive rate. Repro (executed): heat_factor=0.5, p=1000, T=24, cool_factor=0.1, outdoor=20 gives T=24.5 without sense_coeff (warms) vs 23.5 with it (cools).

Both are narrow since cooling is new, but #1 is a hard crash for anyone with `sense: null` and #2 silently does the opposite of the intent. Fixes are small. Happy to PR them, or leave it to @mime24 if you'd prefer the original author. Whatever's easiest, @davidusb-geek.

mime24:
Obviously, I haven't tested properly. Please apologize!
And thank you @OptimalNothing90 for reporting this. 

If you have the possibility to fix it on short notice, feel free to do so. It's really appreciated. 

Otherwise I can try tomorrow

[...]

PR #992, opened 2026-06-18
fix: schedule cooling for shared-tank comfort_sense: cool zones

### LINKED ISSUE #991: heat_topology cooling (comfort_sense: cool) never schedules cooling
Running a `heat_topology` config on my Altherma where a heat-pump source feeds a `comfort_sense: cool` zone, the source never cools: it stays at 0 W and the zone drifts up to its `max_temperatures` — or the run goes infeasible on a hot day, since a heat-only source can't satisfy a max it's already above.

Looks like the same problem as #994, but on the shared-tank / `thermal_source` path instead of the per-load `thermal_battery` one. Digging in, the shared-tank path doesn't carry the cooling sense the way the per-load paths do (#888/#898): the source COP stays clamped to 1.0, the tank dynamics only add heat, and there's no comfort penalty pulling the zone toward target — so doing nothing is cost-optimal.

I put a fix up in #992. Tested on a Daikin Altherma 3 + underfloor: cooling runs and holds the zone near target, heating path unchanged.

Happy to share the repro if it helps.

Thimpey:
Fixes #991. On a `heat_topology` (shared-tank) setup, a heat-pump source feeding a `comfort_sense: cool` zone never schedules cooling — it sits at 0 W, or goes infeasible on a hot day.

Three small things in the shared-tank path that the per-load thermal paths already handle (#888/#898):

- the source COP was clamped to 1.0 because `compile_heat_topology` didn't pass the zone's `comfort_sense` down to the source — now it does;
- the tank dynamics used a fixed `+` sign, so a source could only ever heat — added the `sense_coeff` the per-load paths use;
- there was no penalty toward `desired_temperatures`, so idling was cost-optimal — added the same directional comfort penalty.

Builds on #975 (now merged), which keeps a shared-tank source active so it can schedule at all — this adds the cooling behaviour on top. Reproducer tests are included and the heating path is unchanged. Running it on a Daikin Altherma 3 + underfloor.

Happy to adjust the approach if you'd rather handle any of these differently.

### DIFF
--- a/src/emhass/optimization.py
+++ b/src/emhass/optimization.py
@@ -2615,15 +2615,16 @@ class Optimization:
         where cop_k is resolved via utils.resolve_thermal_battery_cop (Carnot
         for heat pumps, flat for constant-efficiency sources like gas).
 
-        Returns: (predicted_temp_var, heating_demand_arr) for downstream
-        plotting / publishing.
+        Returns: (predicted_temp_var, heating_demand_arr, penalty_term) where
+        penalty_term is the signed comfort penalty (<= 0, added to the Maximize
+        objective) or None when no desired_temperatures are configured.
         """
         tank = self._get_shared_thermal_tanks()[tank_idx]
         tank_id = tank.get("id", f"tank{tank_idx}")
         required_len = self.num_timesteps
         load_ids = [int(k) for k in tank.get("load_ids", [])]
         if not load_ids:
-            return None, None
+            return None, None, None
 
         # Tank physics
         volume = tank["volume"]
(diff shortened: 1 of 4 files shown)


Thimpey:
Thanks @sourcery-ai 🙏 — two quick notes:

**Membership set:** `shared_tank_membership` is already a `dict[int, int]` (load index → shared-tank index, built in `_load_shared_tank_membership`), so `k in shared_tank_membership` is an O(1) hash lookup rather than a linear scan. I kept it a dict because the mapped value is used later (e.g. the `window_outside_horizon` handling), which a plain set of indices would drop.

**Shared overshoot/Big-M helper:** good call on the sign-flip readability. The same heat/cool Big-M block already exists verbatim in `_add_thermal_load_constraints` and `_add_thermal_battery_constraints` (#898); this PR's shared-tank block mirrors those exactly so the inequalities stay consistent across all three paths. Extracting a common helper would touch both pre-existing methods, which feels out of scope for a behavioural cooling fix — happy to do it as a dedicated follow-up refactor so the three paths converge.

Thimpey:
Rebased on master now that #975 is in — this is just the cooling changes now.

Thimpey:
@snauwaertc heads-up — this overlaps with your heat-topology-full-stack work. Your branch already cools shared tanks: `sense_coeff` + the directional comfort penalty in `_add_shared_thermal_tank_constraints`, and `comfort_sense` propagated through `compile_heat_topology`. This PR does the same three things, just on top of current master.

I don't want two parallel implementations competing here. Are you planning to bring the shared-tank cooling upstream yourself as part of the topology stack? If so I'll close this in favour of that.

snauwaertc:
Hi @Thimpey, thanks for the heads-up. Happy to integrate, but testing still continues on my side (and others) for the full-stack repo, so feel free to propose PR and I'll rebase afterwards if you prefer.

davidusb-geek (approved):

Issue #12, opened 2022-06-16
Selected module throws "KeyError" in the log


vdiogo:
I have updated the module by choosing _Trina Solar TSM-340DD14A(II)_ from the PEV DB the one that matches my solar panels, and I get the following exception in the log.

```
The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/usr/local/lib/python3.9/dist-packages/flask/app.py", line 2077, in wsgi_app
    response = self.full_dispatch_request()
  File "/usr/local/lib/python3.9/dist-packages/flask/app.py", line 1525, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/usr/local/lib/python3.9/dist-packages/flask/app.py", line 1523, in full_dispatch_request
    rv = self.dispatch_request()
  File "/usr/local/lib/python3.9/dist-packages/flask/app.py", line 1509, in dispatch_request
    return self.ensure_sync(self.view_functions[rule.endpoint])(**req.view_args)
  File "/usr/local/lib/python3.9/dist-packages/emhass/web_server.py", line 134, in action_call
    input_data_dict = set_input_data_dict(config_path, str(config_path.parent), costfun,
  File "/usr/local/lib/python3.9/dist-packages/emhass/command_line.py", line 76, in set_input_data_dict
    P_PV_forecast = fcst.get_power_from_weather(df_weather)
  File "/usr/local/lib/python3.9/dist-packages/emhass/forecast.py", line 317, in get_power_from_weather
    module = cec_modules[self.plant_conf['module_model'][i]]
  File "/usr/local/lib/python3.9/dist-packages/pandas/core/frame.py", line 3505, in __getitem__
    indexer = self.columns.get_loc(key)
  File "/usr/local/lib/python3.9/dist-packages/pandas/core/indexes/base.py", line 3623, in get_loc
    raise KeyError(key) from err
KeyError: 'Trina Solar TSM-340DD14A(II)'
```

davidusb-geek:
![image](https://user-images.githubusercontent.com/32926550/174106198-0822c672-bbe6-4ad5-b656-37c927dd7740.png)

PR #996, opened 2026-06-20
fix(thermal): compute cooling demand for thermal_battery sense=cool (#994)

### LINKED ISSUE #994: thermal battery with sense=cooling doesn't work as expected
**Describe the bug**
a configured thermal battery with sense=cooling doesn't schedule the cooler to cool.
A heating demand is calculated for the chilly night hours instead of cooling demand for the hot daytime.
it seems that `calculate_heating_demand_physics` doesn't handle the cooling sense.

```
# Calculate temperature difference (only heat when outdoor < indoor)
    temp_diff = indoor_target_temperature - outdoor_temps
    temp_diff = np.maximum(temp_diff, 0.0)
```

**To Reproduce**
configure a thermal battery with sense=cool.
e.g.
```
"thermal_battery": {
            "sense": "cool",
            "indoor_target_temp": 24,
            "volume": {{ 120 * 0.1 }},
            "u_value": 0.252,
            "envelope_area": {{ 392 - 50 }},
            "window_area": 10,
            "shgc": 0.6,
            "ventilation_rate": 0.3,
            "heated_volume": 597,
            "supply_temperature": 18.5,
            "min_temperatures": {{ [22] * horizon }},
            "max_temperatures": {{ [26] * horizon }},
            "start_temperature": {{  states('sensor.raumtemperatur') | float(24.0) }},
            "thermal_inertia_time_constant": 1
            }
```

**Expected behavior**
a cooling demand should be calculated for hours when it's hot outside.

**Screenshots**
calculated demand in night time
Image

 - Home Assistant OS

**Your hardware**
- OS: HA OS

**EMHASS installation type**
 - Add-on

**Additional context**
@OptimalNothing90 and myself already touched the cooling sense but it still seems not working. I'm trying out some fixes but I'm not so familiar with the code and coding. so help is appreciated :)
 
I also think it might be worth to label entities related to cooling with cool/cooling instead of heat.

LesIT1:
## What this fixes

Closes #994. A `thermal_battery` configured with `sense: "cool"` was scheduling the cooler at night instead of during the hot part of the day. The cause is in `calculate_heating_demand_physics`: it only ever computed a heating load, capping the temperature delta at `indoor - outdoor` so demand was non-zero only when it was colder outside than the target. `sense` was never passed into the function from any of its call sites, so the cooling direction never reached the demand calculation. The actuator side (`sense_coeff`) was already wired up from #12, but the demand magnitude was still heating-only, which is why the optimizer saw a phantom load on cold nights.

## The change

`calculate_heating_demand_physics` now takes a `sense` argument (default `"heat"`).

- `"heat"` is byte-identical to the old behaviour. The temperature delta, the solar and internal gain subtraction, and the floor at zero are all unchanged. There's a test asserting the default and `sense="heat"` produce exactly the same array.
- `"cool"` reverses the delta to `outdoor - indoor`, so the load lands on the hot timesteps, and solar and internal gains now add to the load instead of reducing it. The result is returned as a signed negative value (a heat gain), which matches the sign convention already used by `calculate_thermal_loss_signed`. The shared temperature equation subtracts the demand term, so a negative demand correctly raises the indoor temperature when it's hot, and the cooler removes it.

`sense` is threaded through all four call sites in `optimization.py` (the build path, its fallback, the warm-start cache update, and the shared-tank path).

## Behaviour change to be aware of

The published `heating_demand_heater{k}` value is now negative for a cooling load (it represents heat the building gains rather than heat delivered). I updated `plan_output_schema.md` to spell that out. Heating loads are unaffected.

## A couple of notes

- Cooling demand is only modelled by the physics method (`u_value`, `envelope_area`, `ventilation_rate`, `heated_volume`). The degree-day method (`specific_heating_demand` + `area`) is still heating-only. I documented this in `thermal_battery.md` rather than changing the degree-day path, to keep the diff focused, but happy to add a cooling branch there too if you'd prefer.
- When `indoor_target_temperature` is omitted it defaults to the lower temperature bound, which slightly overstates a cooling load. I left the default alone and the docs now tell cooling users to set `indoor_target_temperature` explicitly. Easy to switch the cooling default to the upper bound if you'd rather.
- @mime24, in the config on the issue the key was `indoor_target_temp`, but the code reads `indoor_target_temperature`, so your 24 was being ignored and it fell back to the min. Worth fixing in your config once this lands.

## Tests

- `tests/test_utils.py`: a behavioural test that fails on the current code (cooling demand lands on the cold steps instead of the hot ones), a test that cooling demand is signed as a heat gain and that solar gains increase it, and a no-op test that `sense="heat"` matches the default exactly.
- `tests/test_optimization.py`: an end-to-end test through `perform_optimization` showing a physics cooling load runs the cooler more on a hot outdoor profile than a cool one.

I don't have a cooling device here so I can't field-test this on real hardware. The maths and the optimizer behaviour check out, but a confirmation from someone running an actual cooler would be welcome.

Thanks to @sokorn for the original thermal_battery model and @metzi for the cooling sense groundwork.

## Summary by Sourcery

Add cooling-mode support to the physics-based thermal demand calculation and wire it through thermal battery optimization so cooling loads are scheduled on hot periods with correct sign conventions.

Bug Fixes:
- Fix thermal battery with sense="cool" erroneously scheduling cooling at night by computing demand based on outdoor heat gain instead of heat loss only.

Enhancements:
- Extend calculate_heating_demand_physics with a sense parameter to support both heating and cooling while preserving existing heating behaviour.
- Propagate the thermal sense configuration through all physics-based thermal battery and shared tank demand calculations.

Documentation:
- Clarify thermal battery documentation to describe how sense affects physics-based heating vs cooling demand, and note that the degree-day method remains heating-only.
- Update the plan output schema to explain that heating_demand_heater{k} is positive for heating loads and zero or negative for cooling loads.

Tests:
- Add unit tests for cooling-mode physics demand covering timestep targeting, sign conventions, and solar gain effects.
- Add an end-to-end optimization test verifying that a physics-based cooling thermal battery runs more on hot outdoor profiles than on cool ones.


LesIT1:
Thanks for the review. Addressed in a102acd9.

On the bug_risk finding: good catch. The degree-day (`specific_heating_demand` + `area`) path is heating-only by design, and it was silently dropping `sense="cool"` and producing heating demand on cold periods. Rather than extend the HDD model to cooling (degree-days are a heating concept, so that is a larger change for another time), I went with option (a): the optimizer now logs a warning at all three HDD call sites (single-source parameterized and fallback paths, plus the shared-tank path) telling the user the degree-day model is heating-only and to use the physics parameters for cooling. Added a test that asserts the warning fires, and noted the limitation in the docs.

On the sign convention: added a comment where `heating_demand` is consumed in the tank dynamics, spelling out that it is `>=0` for heating and `<=0` for cooling so the sign is explicit for anyone touching those constraints later.

On centralizing the `sense` default: fair point, left it as-is for now to keep this PR focused, but it is a reasonable cleanup for a follow-up.

davidusb-geek (approved):

PR #998, opened 2026-06-21
feat: carry over completed operating timesteps for must-run deferrables (closes #983)

### LINKED ISSUE #983: Carry over completed operating hours for must-run deferrables across the rolling horizon (#952 follow-on)

LesIT1:
## What this is

This is the concrete version of the carryover proposal in #983. I built it because the constraint handling and the state handoff are much easier to discuss against real code than in the abstract, and it slots in alongside #980 and #982. It is entirely opt-in and a no-op by default, so it costs nothing for anyone not using it. Whether it belongs in core is your call, I am happy to reshape it or drop it.

Closes #983 if you want it.

## The problem

In naive-MPC the optimiser re-plans every cycle and rebuilds the required run for each deferrable from `operating_hours` on every solve, with no memory of how much the load has already run today. For a flexible load that can fragment this is harmless. For a must-run load pinned to a daily window it is not: as the day advances and the full block no longer fits the remaining window, the optimiser pushes it later and eventually rolls the whole block to tomorrow, so the load under-runs its daily target.

Real example from my own setup: a pool pump as a must-run `set_deferrable_load_single_constant` load, about 4 hours/day, pinned to a 09:00 to 15:00 window, 15-minute MPC. Once the clock passed the point where 4 hours still fit before 15:00, each solve moved the block later and then rolled it to the next day. The pump ran a couple of short fragments instead of its 4 hours.

#980 (`def_current_on_timesteps`) only carries the min-on-time tail, and is gated `not is_single_const`. #982 (`def_current_power`) only pins t0. Neither decrements the remaining required run, so nothing tells the optimiser that part of today's hours is already done.

## The change

A new runtime-only, per-load signal `def_current_operating_timesteps`: a list of integers, one per deferrable load, saying how many operating timesteps each load has already completed today at the start of the horizon. Each solve then decrements that load's `required_timesteps` and `target_energy` by the elapsed amount, clamped at 0, so it schedules only the remainder. It applies to standard and single_constant must-run loads.

On the design question I raised in the issue (extend `def_current_on_timesteps` vs a sibling): I went with a sibling. `def_current_on_timesteps` is deliberately gated `not is_single_const` and only fires when `def_minimum_on_time > 0`, so overloading it would have changed its existing behaviour and still needed a separate single-constant path. A sibling mirrors the #982 runtime-only measurement pattern exactly and keeps both features clean. Easy to change if you would rather it were one parameter.

The supervisor still owns the elapsed tracking and the daily reset boundary (which, as @scruysberghs rightly noted, is not always midnight), exactly the same split as #980 and #982. It just feeds the number.

## Behaviour change disclosure

- Default unset, or an all-zero list, is a byte-exact no-op. There is a test that pins this.
- When a load's elapsed reaches its required run, that load is treated as satisfied for the day: it is released, not forced. It is dropped from the must-run forcing (the single-constant pin, the min-on-time tail, and the `def_current_power` t0 pin are all released for that load), so it is free to run if it is economical but is never forced to start a fresh block. This only affects a load you are actively feeding the new signal for.
- This is a fed, trusted signal. If you tell EMHASS a load has done 3 hours, it believes you, the same as the existing current-state inputs.

## Non-goals

- EMHASS does not track wall-clock day rollover or decide when "today" resets. The supervisor feeds the elapsed count, same as #980/#982.
- No automatic elapsed tracking inside EMHASS.
- Flexible (fragmentable) loads are unchanged beyond the remainder decrement.

## Tests

In `tests/test_optimization.py`:
- `test_current_operating_timesteps_single_const_red` red/green proof: a must-run single-constant load with elapsed > 0 schedules only the remainder (fails on master).
- `test_current_operating_timesteps_noop_no_key`: absent key reproduces today's plan.
- `test_current_operating_timesteps_standard_load`: standard load decrement.
- `test_current_operating_timesteps_clamp_elapsed_ge_required` and `_over_elapsed_clamp`: elapsed >= required clamps to 0, model stays feasible, load not forced.
- `test_current_operating_timesteps_length_mismatch_warns`: a length mismatch logs a warning and still solves.
- `test_current_operating_timesteps_running_singleconst_clamp`: a currently-running single-constant load at/over its target solves cleanly with no phantom block.
- `test_current_operating_timesteps_satisfied_with_current_power_feasible`: a satisfied load that is also reporting current power stays feasible.
- `test_current_operating_timesteps_satisfied_standard_min_on_feasible`: a satisfied standard load that also has a min-on-time remainder solves to plain Optimal (not the relaxed fallback).

`tests/test_runtime_params.py` covers the runtime-key plumbing (it stays disjoint from the config schema, like `def_current_power`). Full suite green locally (629 passed, 28 xfailed).

Documented in `docs/config.md` next to the other `def_current_*` runtime signals.

## Summary by Sourcery

Introduce a runtime-only signal to decrement completed operating timesteps for must-run deferrable loads so MPC only schedules remaining required run and releases loads whose daily requirement is already satisfied.

Bug Fixes:
- Prevent infeasible or relaxed fallback solutions caused by conflicts between must-run requirements, minimum-on-time constraints, and current-power pins when loads have already met their operating-hour targets.
- Ensure single-constant and standard must-run loads are not forced to run phantom extra blocks once their daily operating requirement is fully satisfied.

Enhancements:
- Add CVXPY parameters and optimization logic to apply completed-operating-timestep decrements across standard and single-constant deferrable loads, including safe clamping and interaction with load-activation flags.
- Extend runtime parameter handling and config hashing to support the new completed-operating-timesteps signal without affecting cached problem structures.

Documentation:
- Document the new def_current_operating_timesteps runtime parameter, its units, interaction with other deferrable load signals, and typical MPC usage in the configuration guide.

Tests:
- Add optimization tests verifying decrement behavior, clamping when elapsed time meets or exceeds requirements, length-mismatch handling, and interactions with current power and minimum-on-time constraints for both standard and single-constant loads.
- Add runtime-parameter tests to cover plumbing for the new def_current_operating_timesteps key.

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -161,6 +161,19 @@ Example:
   **Single-constant, sequence and thermal loads** ignore `def_current_power` (it is a no-op for those load types). A single-constant load runs as one fixed block, so its currently-running state is handled by `def_current_state` (which pins the remaining required timesteps); pinning a below-nominal power there would fight the required-energy target. Use `def_current_state` for a currently-running single-constant load.
 
   **Typical MPC use:** read the load's power sensor at the start of each optimization tick and pass the reading in watts via the runtime API so the optimizer knows the load is running and at what level.
+- `def_current_operating_timesteps`: Per-load integer list: how many operating timesteps each must-run deferrable load has already completed today at the start of this optimization horizon. The optimizer schedules only the REMAINDER: `required_timesteps` and `target_energy` are each decremented by the elapsed amount, clamped at 0. **Default `null` (absent key) or `[0, 0]` is an exact no-op** -- today's plan is unchanged. One non-negative integer per deferrable load, in the same order as `nominal_power_of_deferrable_loads`.
+
+  Unit: timesteps. Convert from time: `elapsed_hours / optimization_time_step`. With the default 30-minute step, 2 elapsed hours = 4 timesteps.
+
+  Applies to **both standard and single_constant must-run loads** (unlike `def_current_on_timesteps` which is gated on `def_minimum_on_time > 0` and excluded for single_constant loads). If a load has no required operating hours configured (`operating_hours_of_each_deferrable_load[k] == 0`), the elapsed value is ignored.
+
+  When the elapsed count equals or exceeds the total required timesteps, the remainder clamps to 0 and the constraint is fully relaxed -- the load is no longer forced to run in this tick, and the solve stays Optimal.
+
+  **Interaction with `def_current_state`:** if you also pass `def_current_state[k]=true` for a single_constant load, the optimizer uses both signals: the currently-running pin (block A) uses the *decremented* `required_timesteps`, so the running-block anchor also shrinks automatically.
+
+  **Daily reset boundary:** EMHASS does not track wall-clock day rollover. The caller (supervisor / automation) is responsible for resetting the elapsed count to 0 at the start of each day, exactly as with `def_current_on_timesteps` and `def_current_power`.
+
+  **Typical MPC use:** at each optimization tick, read the load's run-time counter (e.g. from a Home Assistant energy meter or runtime helper) and pass the elapsed timesteps so the optimizer schedules only the outstanding run.
 - `deferrable_load_max_cost`: Make a deferrable load *optional* by capping how much it may cost to run. This is a list of floats, one value per deferrable load, in the same currency units as your load cost forecast. The default `0` keeps a load **mandatory**: the optimizer must deliver its full `operating_hours_of_each_deferrable_load` of energy within the horizon, as it always has. A value above `0` makes the load **optional**: the optimizer schedules it only when its complete run can be done for less than that amount (the cost of the energy it consumes plus any `set_deferrable_startup_penalty`), otherwise the load is left off for the whole horizon. This suits a "nice to have" load that is only worth running when energy is cheap enough, for example a heat pump boosting hot water above its normal setpoint, or an immersion heater that should only soak up surplus when prices are very low. The cap is an all-or-nothing budget for the load's entire run, not a per-timestep price limit. It applies to every deferrable load type (standard, semi-continuous, single-constant and sequence loads) except thermal loads configured with a `thermal_config` or `thermal_battery`, which follow their own temperature targets instead. For example:
 	```json
 	"deferrable_load_max_cost": [0, 0.65]
--- a/src/emhass/command_line.py
+++ b/src/emhass/command_line.py
@@ -275,6 +275,9 @@ class OptimizationCache:
             "def_current_off_timesteps",
             # Per-call current power in watts (issue #605); pin value is a cp.Parameter.
             "def_current_power",
+            # Per-call completed operating timesteps today (issue #983); decrements
+            # required_timesteps + target_energy via cp.Parameter (no rebuild on cache hit).
+            "def_current_operating_timesteps",
             "minimum_power_of_deferrable_loads",
             "cost_forecast_per_deferrable_load",
             # shared_thermal_tanks has its own structural hash field above
--- a/src/emhass/optimization.py
+++ b/src/emhass/optimization.py
@@ -414,6 +414,16 @@ class Optimization:
             active.value = 0.0
             self.param_def_current_power_active.append(active)
 
+        # Completed operating-timesteps parameters (issue #983).
+        # param_current_operating_timesteps[k]: how many operating timesteps load k has
+        # already run today. Used to decrement required_timesteps and target_energy in the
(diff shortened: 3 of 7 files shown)


LesIT1:
Thanks, both fair points.

On centralising the satisfied-load handling: I did look at pulling it into a helper, but the three touch points write to different CVXPY parameters in separate loops (the decrement builds the satisfied set, param_load_active is its own pass, and the min-on lower bound sits in the running-lb block), so one helper would have to thread all of those through and I think it ends up more indirection than it saves. What actually guards the regression you are describing is the test side: the three interaction tests (currently-running single-const, satisfied plus current power, satisfied standard plus min-on) pin that a satisfied load is released from every force-on path, so a future constraint that forgets it fails the suite. Happy to add the helper as a follow-up if you would prefer it.

On _make_cots_overrides: fair, it is deliberately scoped to these tests for now. If more runtime-signal tests land I will fold it into the shared scaffolding rather than copy it.

davidusb-geek (approved):

lutorm:
Question about this parameter:  is this the correct way to tell the optimizer that a sequence load has already run partway through its sequence? I'm confused about this vs `def_current_operating_timesteps` and how/whether any of them affect scheduling of sequence loads.

PR #999, opened 2026-06-21
fix(thermal): keep solar gains when pv_power_forecast is passed (#997)

### LINKED ISSUE #997: Passing pv_power_forecast disables thermal solar gains (no way to supply GHI when weather method is 'list')
**Describe the bug**

If you pass your own `pv_power_forecast` as a runtime parameter (which a lot of people do, e.g. feeding Solcast in themselves), the thermal battery's physics-based demand silently stops accounting for solar gains. The window solar gain term just disappears, for heating as well as cooling.

This came up while looking at #994 (@mime24 spotted it). It's a separate problem from the cooling fix in #996, so I'm splitting it out.

**What's happening**

Passing `pv_power_forecast` flips `weather_forecast_method` to `"list"` in `treat_runtimeparams` (`utils.py`, the loop around line 1711, `params["optim_conf"][forecast_methods[method]] = "list"`). The `"list"` weather path returns only the PV power series, so the `ghi` column never gets created.

The physics demand then guards on that column:

```python
if "ghi" in data_opt.columns and window_area is not None:
    ...solar_irradiance = ...
```

(`optimization.py`, in `_add_thermal_battery_constraints`). With no `ghi` column the branch is skipped and solar gains are dropped without any warning. The same applies to the surface solar gain path (pool / outdoor tank), which also reads `data_opt["ghi"]`.

The catch is there's no way to supply GHI on its own. `ghi` / `solar_irradiance_forecast` isn't an accepted runtime forecast key (`pv_power_forecast`, `load_power_forecast`, `load_cost_forecast`, `prod_price_forecast`, `outdoor_temperature_forecast` are the only ones), and `outdoor_temperature_forecast` has the same issue: once you pass your own PV you also have to pass your own outdoor temperature, and even then there's no slot for irradiance. So the moment you bring your own PV forecast, the thermal model loses both the GHI-based solar gain and (unless you also pass it) the outdoor temperature it needs.

**To reproduce**

1. Configure a `thermal_battery` with the physics method (`u_value`, `envelope_area`, `ventilation_rate`, `heated_volume`) and a `window_area` so solar gains apply.
2. Run a `naive-mpc-optim` (or dayahead) passing `pv_power_forecast` as a runtime parameter.
3. The demand is computed with no solar gain contribution. The debug log shows `Retrieving weather forecast data using method = list` and there's no `ghi` in the data.

**Expected behavior**

Bringing my own PV forecast shouldn't switch off the solar-gain modelling. Either the weather/GHI forecast should be decoupled from the PV-power forecast, or there should be a way to pass GHI (and irradiance generally) as a runtime list the way the other forecasts can be passed.

**Possible directions** (whichever fits your design best)

- Accept `solar_irradiance_forecast` (GHI) as a runtime forecast key, same shape as the others, so people who pass their own PV can also pass GHI.
- Or let the weather forecast (for GHI) still run via open-meteo even when `pv_power_forecast` is supplied as a list, so the two aren't coupled through a single `weather_forecast_method`.

Happy to put a PR together once you've got a preferred direction. I don't want to guess at the API surface you'd want here.

**EMHASS version**: master (v0.17.7 and later)

LesIT1:
## What this fixes

Closes #997. If you pass your own `pv_power_forecast` as a runtime parameter, the thermal battery's physics demand silently stops accounting for solar gains, for heating and cooling.

The cause is that passing any list forecast (including `pv_power_forecast`) flips `weather_forecast_method` to `"list"` in `treat_runtimeparams`. The `"list"` weather path only carries the PV power, so the `ghi` column never gets built, the `if "ghi" in data_opt.columns` guards in the demand calculation skip the solar term, and the gains disappear with no warning. The same thing happens to the outdoor-temperature fallback: once you bring your own PV you also lose the open-meteo `temp_air` the thermal model would otherwise fall back on.

@mime24 spotted the original symptom while we were looking at #994, and we agreed on the approach in the issue thread: keep the weather/GHI fetch independent of the PV list rather than adding a new runtime key.

## The change

When the PV power is supplied as a list and a thermal load is actually configured, EMHASS now still fetches the open-meteo weather forecast and grafts its weather columns (`ghi`, `temp_air`, and the rest) onto the list frame. The PV power stays exactly the list you passed. With `ghi` and `temp_air` present again, the existing GHI merge and the temp-air fallback in `command_line` pick them up the way they already do for the open-meteo path, so solar gains and the outdoor-temperature fallback survive.

It is all in `forecast.py`. `command_line.py` is untouched because its merge already keys off whether the column exists.

## Behaviour change to be aware of

- The extra open-meteo fetch only happens when a `thermal_config` or `thermal_battery` load is present. A pure PV/battery setup that passes `pv_power_forecast` is unaffected and makes no extra weather call, so it is a no-op for everyone not using a thermal load.
- It is fail-soft. If open-meteo is unreachable the run logs a warning and carries on without solar gains, the same as today, rather than failing. So an offline or air-gapped setup keeps working.
- One small related change in the same path: when the weather cache is enabled, a stale cache that does not cover the requested window was being served best-effort (zero-filled GHI) under the `list` method. Since the only way `list` reaches that cache is through this open-meteo augmentation, and open-meteo has no rate limit, I let it refetch fresh instead, the same as the normal open-meteo path. This only affects the `list` + thermal + caching combination.

## The alternative I did not take

The other option in the issue was to accept `solar_irradiance_forecast` (GHI) as a new runtime forecast key. @mime24 and I both preferred keeping the weather fetch decoupled from the PV list because it means solar gains just keep working with no new key to learn. Happy to switch to the explicit-key shape instead if you would rather have that.

## Tests

In `tests/test_forecast.py`:

- a thermal load plus a passed PV list now gets `ghi` and `temp_air` from open-meteo while `yhat` stays the passed list (fails on master),
- the no-op case: no thermal load means no open-meteo fetch and the frame stays PV-only,
- two fail-soft cases (open-meteo raising and returning nothing) leave the plain list frame and warn,
- the stale-cache refetch under the `list` method.

Full suite passes locally (625 passed, 28 xfailed).

## Summary by Sourcery

Ensure thermal loads retain solar gains and outdoor temperature fallbacks when pv_power_forecast is provided as a runtime list by augmenting list-based forecasts with weather data.

Bug Fixes:
- Preserve GHI and outdoor temperature data for thermal loads when pv_power_forecast is supplied via the list weather method.
- Treat stale cached weather data for list-based forecasts like open-meteo data, forcing a refetch instead of serving zero-filled irradiance.

Enhancements:
- Add helper logic to detect thermal loads and augment list-based PV forecasts with open-meteo weather columns while leaving the passed PV list unchanged.

Documentation:
- Document how solar gains behave when passing pv_power_forecast and clarify that GHI is sourced automatically from the weather forecast in this case.

Tests:
- Add tests covering list-method weather augmentation for thermal loads, the no-op behavior for non-thermal setups, fail-soft handling of open-meteo failures, and stale cache refetching under the list method.

### DIFF
--- a/docs/thermal_battery.md
+++ b/docs/thermal_battery.md
@@ -898,7 +898,7 @@ rest_command:
 Important notes:
 - `outdoor_temperature_forecast` is required for thermal battery optimization (building heating mode)
 - `start_temperature` should ideally come from a real sensor (floor temp for underfloor heating, tank sensor for hot water)
-- If using solar gains, ensure your forecast data includes `ghi` (global horizontal irradiance)
+- If using solar gains, the optimization data needs a `ghi` (global horizontal irradiance) column. This comes from the weather forecast automatically, including when you pass your own `pv_power_forecast` (see "Solar gains not working" in Troubleshooting)
 - Add `"thermal_inertia_time_constant": 2.0` to the `thermal_battery` dict to enable the thermal inertia filter
 - In MPC mode, `Q_input` auto-persists between solves; use `"q_input_initial": 0.5` to manually override
 - For hot water tanks, add `"density": 997, "heat_capacity": 4.184` and a `"draw_off_demand"` profile or `"specific_heating_demand": 0.0, "area":1.0`
@@ -1053,6 +1053,21 @@ Requirements for solar gains:
 3. Must use physics-based method (not HDD method)
 
 Check the logs for: "Using physics-based heating demand with solar gains"
+
+#### Passing your own `pv_power_forecast`
+
+Supplying `pv_power_forecast` as a runtime parameter switches the weather forecast
+method to `list`, which on its own only carries the PV power and no `ghi` or
+`temp_air`. To keep solar-gain modelling working in that case, EMHASS still fetches
+the open-meteo weather forecast for the `ghi` and outdoor-temperature columns
+whenever a thermal load is configured, while using your passed list for the PV
+power. So bringing your own PV forecast no longer silently disables solar gains.
+
+This extra weather fetch only happens when a `thermal_config` or `thermal_battery`
+load is present, so pure PV/battery setups that pass `pv_power_forecast` are
+unaffected and make no extra weather call. If open-meteo is unreachable the run
+falls back to no solar gains (a warning is logged) rather than failing. A passed
+`outdoor_temperature_forecast` still takes precedence over the fetched temperature.
 
 ### Hot water tank temperature drops too fast
 
--- a/src/emhass/forecast.py
+++ b/src/emhass/forecast.py
@@ -764,6 +764,61 @@ class Forecast:
         data.set_index("ts", inplace=True)
         return data
 
+    def _list_method_needs_weather(self) -> bool:
+        """Whether a configured deferrable load consumes weather-derived data.
+
+        The ``list`` weather path only carries the passed PV power (``yhat``). A
+        thermal load's physics demand additionally needs ``ghi`` (solar gains) and
+        ``temp_air`` (the outdoor-temperature fallback), so when one is configured
+        we still fetch open-meteo to supply those columns (issue #997). Pure
+        PV/battery setups never trigger the extra fetch, keeping the list path a
+        no-op for them.
+        """
+        return any(
+            isinstance(cfg, dict)
+            and (
+                isinstance(cfg.get("thermal_config"), dict)
+                or isinstance(cfg.get("thermal_battery"), dict)
+            )
+            for cfg in self.optim_conf.get("def_load_config", []) or []
+        )
+
+    async def _augment_list_with_open_meteo(
+        self, data: pd.DataFrame, w_forecast_cache_path: str, use_legacy_pvlib: bool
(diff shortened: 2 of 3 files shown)


LesIT1:
Thanks. Took the first one in e2526b35: the catch in `_augment_list_with_open_meteo` is now `(aiohttp.ClientError, OSError, ValueError, KeyError)`, which covers the network, cache IO and malformed-response cases the open-meteo fetch can actually hit, while letting a real programming error propagate instead of being hidden as a warning.

On the reindex sanity check, I left it as is. By the time the weather frame reaches that line it has already been reindexed onto `self.forecast_dates` inside `_get_weather_open_meteo` (and padded via interpolate), and the list frame is built on `self.forecast_dates_tz`, which is the same horizon. So there isn't a second horizon that could drift out of alignment, the `nearest` only soaks up sub-second rounding between the two indexes. command_line also re-reindexes the merged `ghi`/`temp_air` with a 1h tolerance plus a ffill/bfill afterwards, so anything odd there would get caught and smoothed too.

davidusb-geek (approved):

PR #1004, opened 2026-06-26
fix: hold-last fill for dict forecasts with pre-horizon points (#1003)

### LINKED ISSUE #1003: Incorrect backfill / forward fill of dictionary forecast data when value is 0
I have 2 pricing sensors that has the following characteristics:
- Load:
  - Between 11:00 and 14:00, the price is $0
  - Between 16:00 and 23:00, the price is $0.407
  - At all other times, the price is $0.308
- Production:
  - Between 18:00 and 21:00, the price is $0.10
  - At all other times the price is $0

This is the runtime parameters between passed to the naive mpc optim endpoint:
```json
{
   "prediction_horizon":576,
   "soc_init":0.51,
   "soc_final":0.1,
   "battery_minimum_state_of_charge":0.1,
   "battery_maximum_state_of_charge":1,
   "load_cost_forecast":{
      "2026-06-26T08:00:00+10:00":0.308,
      "2026-06-26T11:00:00+10:00":0,
      "2026-06-26T14:00:00+10:00":0.308,
      "2026-06-26T16:00:00+10:00":0.407,
      "2026-06-26T23:00:00+10:00":0.308,
      "2026-06-27T11:00:00+10:00":0,
      "2026-06-27T14:00:00+10:00":0.308,
      "2026-06-27T16:00:00+10:00":0.407,
      "2026-06-27T23:00:00+10:00":0.308,
      "2026-06-28T11:00:00+10:00":0,
      "2026-06-28T14:00:00+10:00":0.308,
      "2026-06-28T16:00:00+10:00":0.407,
      "2026-06-28T23:00:00+10:00":0.308
   },
   "prod_price_forecast":{
      "2026-06-26T08:00:00+10:00":0,
      "2026-06-26T18:00:00+10:00":0.1,
      "2026-06-26T21:00:00+10:00":0,
      "2026-06-27T18:00:00+10:00":0.1,
      "2026-06-27T21:00:00+10:00":0,
      "2026-06-28T18:00:00+10:00":0.1,
      "2026-06-28T21:00:00+10:00":0
   }
}
```

The problem is that the backfilling of the production forecast price is incorrect, per the generated graph in EMHASS:
Image

It is backfilling the price before 18:00 to $0.10, when the actual price is $0 for anytime before 18:00. In my case, the supplied price forecase dictionary does include an entry at 8:00 that sets the price to $0, so this is unexpected.

The load pricing is also incorrect:
Image

Pricing before 11:00 should be $0.308, but it is backfilled as $0.

LesIT1:
Fixes #1003.

When you pass `load_cost_forecast` or `prod_price_forecast` as a timestamp-to-value dict, EMHASS filled the horizon wrong whenever a point was defined before the start of the optimization window. The price before the first in-window change came out as the *next* value instead of the one that should have been held.

The reporter's case: a production price of 0 before 18:00 showed up as the 0.10 peak, and a load price that should have been 0.308 before 11:00 showed up as 0.

The cause is in `treat_runtimeparams` (the dict branch). It did:

```python
resample(step).aggregate({"value": "mean"}).reindex(forecast_dates, method="nearest")
# then .ffill().bfill()
```

`method="nearest"` snaps each horizon slot to the nearest provided timestamp rather than holding the last value, and it drops any point that sits before `forecast_dates[0]`. With that anchor gone, the trailing `bfill()` fills the leading slots with the next value.

The fix keeps the resample-mean aggregation (still correct for sub-step inputs) but aligns with hold-last semantics: union the provided index with the horizon, forward-fill, then reindex onto the horizon. The union keeps a pre-horizon point alive as a fill anchor, so each value holds until the next change. I parse `forecast_dates` (a list of ISO strings) to a UTC index so the union lines up by instant across DST edges.

Behaviour change to disclose: dict forecasts now use step / hold-last interpolation instead of nearest-point snapping. List forecasts are untouched. For a dict whose first point is already at or after the window start the output is unchanged, so the only thing that moves is the previously-wrong leading region.

Tests: added `test_treat_runtimeparams_dict_forecast_holds_last_value`. It pins the clock, passes a dict with a point one step before the horizon start, and checks that the leading slots hold the anchor value and the full step profile holds across the horizon. The test fails on master (the leading load slot reads 0 instead of 0.308) and passes with the fix. Full suite is green.

I also added a short note to `docs/passing_data.md` covering the timestamped-dict form, since only the list form was documented before.

## Summary by Sourcery

Fix handling of dict-based load and production price forecasts so values are held until the next change, including when an initial point lies before the optimization window, and document and test the corrected behaviour.

Bug Fixes:
- Ensure dict-based load_cost_forecast and prod_price_forecast correctly use hold-last semantics across the optimization horizon, including when an initial point occurs before the horizon start.

Documentation:
- Document how to pass forecasts as timestamp-to-value dictionaries and clarify that EMHASS uses step interpolation, holding each value until the next change.

Tests:
- Add a regression test verifying that dict-based forecasts hold the last value across the horizon and correctly anchor leading slots with pre-horizon points.

### DIFF
--- a/docs/passing_data.md
+++ b/docs/passing_data.md
@@ -55,6 +55,20 @@ The possible dictionary keys to pass data are:
 
 - `prod_price_forecast` for the PV production selling price forecast.
 
+#### Passing a forecast as timestamped values
+
+Instead of a plain list, any of these forecast keys can be passed as an object that maps ISO 8601 timestamps to values. EMHASS aggregates the points to the optimization time step and then holds each value until the next provided point (step interpolation), so you only need to supply a point where the value changes. A point whose timestamp falls before the start of the optimization window is used to anchor the first values of the horizon.
+
+```json
+{
+  "load_cost_forecast": {
+    "2024-01-01T00:00:00+01:00": 0.20,
+    "2024-01-01T06:00:00+01:00": 0.15,
+    "2024-01-01T18:00:00+01:00": 0.30
+  }
+}
+```
+
 ### Passing other data at runtime
 
 It is possible to also pass other data during runtime to automate energy management. For example, it could be useful to dynamically update the total number of hours for each deferrable load (`operating_hours_of_each_deferrable_load`) using for instance a correlation with the outdoor temperature (useful for water heater for example). 
--- a/src/emhass/utils.py
+++ b/src/emhass/utils.py
@@ -1728,19 +1728,27 @@ async def treat_runtimeparams(
                     ).dt.tz_convert(time_zone)
 
                     # Aggregate any sub-step points to the optimization time step.
+                    # Resample in the local time_zone so the buckets line up with the
+                    # forecast grid, which get_forecast_dates floors in local time
+                    # (this matters for sub-hour UTC offsets such as +05:30).
                     forecast_data_df = forecast_data_df.resample(
                         pd.to_timedelta(optimization_time_step, "minutes"),
                         on="time",
                     ).aggregate({"value": "mean"})
-                    # Align with forecast_dates using hold-last (step) semantics: each
-                    # value holds until the next provided point. Union the two indexes
-                    # first so points defined before forecast_dates[0] still anchor the
-                    # forward-fill; reindexing straight onto forecast_dates with
-                    # method="nearest" dropped that anchor and let the trailing bfill
-                    # fill the leading slots with the NEXT value instead (issue #1003).
-                    # forecast_dates is a list of ISO strings; parse to a tz-aware
-                    # (UTC) index so the union aligns by instant across DST edges.
+                    # Now move to UTC so the union/reindex below align by instant
+                    # across DST edges without mixing two differently-localized
+                    # indexes. forecast_dates is a list of ISO strings; parse it to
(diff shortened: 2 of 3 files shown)


LesIT1:
Thanks. I took the clarity fix in 9b6926e6: the resampled index is now converted to UTC before the union, so the union and reindex both run on UTC indexes instead of mixing a local-tz index with a UTC one.

I kept the resample itself in local time on purpose. `get_forecast_dates` floors the horizon grid in the local timezone, so resampling in local time keeps the aggregation buckets aligned with that grid. For a sub-hour offset like +05:30 (Asia/Kolkata) resampling in UTC would shift the buckets off the local-clock grid, so I only normalized the index for the union rather than moving the whole thing to UTC.

On the underlying concern: pandas unions tz-aware indexes by instant, so the previous mixed-tz union wasn't actually shifting timestamps. I checked the DST spring-forward and fall-back days both ways and the output was identical, no NaN and correct length. The explicit UTC normalize just makes that obvious and drops the version-dependent ambiguity. I confirmed the output is byte-identical before and after this change across Europe/Paris, Australia/Brisbane and Asia/Kolkata.

davidusb-geek (approved):

Issue #963, opened 2026-06-09
[Feature request] opt-in battery self-identification (learn capacity + round-trip efficiency from HA history, the same way set_use_adjusted_pv learns PV)


LesIT1:
## The gap

EMHASS already learns two of the three things it needs from Home Assistant history. PV is self-calibrated through `set_use_adjusted_pv` (a regression fit against measured PV), and load is learned by the mlforecaster. The one set of values that stays a hand-entered constant is the battery physics: `battery_nominal_energy_capacity`, `battery_charge_efficiency`, and `battery_discharge_efficiency`. Those are read straight from `plant_conf` and fed into the optimizer (in `optimization.py` they land at `cap`/`eff_dis`/`eff_chg` around lines 1417-1419 and drive the SoC equation `power_flow = (p_sto_pos * (1/eff_dis)) + (p_sto_neg * eff_chg)` at line 1472). Whatever number you typed in is what the optimizer trusts, forever.

## Why it matters

People copy nameplate numbers off a spec sheet. Usable capacity is almost never the nameplate figure once you account for the reserved window, and it fades with age. Round-trip efficiency is frequently guessed. When those constants are wrong the optimizer silently mis-sizes every charge/discharge decision: it over- or under-fills against price windows, and the SoC trajectory it plans drifts from reality. For anyone doing tight arbitrage or deep cycling, a capacity that is 10-15% off is a real money and battery-wear error that nobody ever notices because there is no feedback loop. The data to close that loop (signed battery power plus measured SoC) is usually already sitting in HA.

## Proposed approach

An **opt-in, default-off** feature that learns the battery constants from HA history, structurally cloning the `adjust_pv_forecast` pattern so it reuses the plumbing you already accept: `is_model_outdated` for cadence, the existing `retrieve_home_assistant_data` path, the `data_path` persistence convention, and the `optim_conf` toggle style.

Master toggle `set_use_battery_identification`, default `false`, exactly mirroring how `set_use_adjusted_pv` ships `false` in `config_defaults.json`. When it is off the feature is a true no-op: no extra history retrieval, no fit, no sensors, no config mutation. The optimizer reads `plant_conf` precisely as it does today.

When it is on, EMHASS retrieves two new signals via new `associations.csv` rows mirroring the PV rows (`var_PV -> sensor_power_photovoltaics`):

- `sensor_power_battery` (signed battery power, AC side)
[...]

PR #1001, opened 2026-06-22
fix: atomic entity-file writes so continual_publish survives a partial read

LesIT1:
## The bug

On a `continual_publish` setup, an automation calling `/action/naive-mpc-optim`
starts returning 400 and the result sensors in Home Assistant stop updating.
Later optimization runs log fine, but the sensors stay frozen at the timestamp
of the first failure until both EMHASS and HA are restarted. Reported in #1000
on 0.17.7, reproduces on master.

The traceback:

```
ERROR in app: Exception on request POST /action/naive-mpc-optim
  File "command_line.py", in continual_publish
    freq = await _publish_and_update_freq(...)
  File "command_line.py", in _publish_and_update_freq
    await publish_json(...)
  File "command_line.py", in publish_json
    entity_data = pd.read_json(entity_path / entity, orient="index")
ValueError: Expected object or value
```

## Root cause

`post_data(save_entities=True)` writes each `data/entities/<id>.json` in place
with a truncating `open(..., "w")`. The `continual_publish` background task reads
those same files on its own schedule, in the same event loop. When a read lands
in the window between truncate and write it sees a zero-byte file, and
`pd.read_json` raises `ValueError: Expected object or value`.

That is the visible error, but the reason a restart is needed is the second part:
`continual_publish` is a bare `while True:` with no exception handling, so that
one read error propagates out and the background task dies. After that, optim
runs still write fresh entity files but nothing republishes them on the continual
cadence, so the sensors stay frozen until the task is respawned at startup. Quart
logs the dead task against the request that started it, and that ERROR line is
what turns the next action response into a 400.

The `metadata.json` write directly below the entity write was already hardened
with a temp file plus `os.replace` in e749da98. The per-entity data file was the
remaining in-place write.

## The fix

- Write each entity file to a unique temp file then `os.replace` it into place,
  mirroring the existing `metadata.json` commit, so a concurrent reader always
  sees a complete document.
- `continual_publish` now catches a transient per-cycle exception, logs it, and
  retries on the next interval instead of exiting. `CancelledError` still
  propagates, so shutdown is unaffected.
- The reader skips in-flight `.tmp` files, since the atomic writes now place them
  briefly in the entities directory. This also covers the pre-existing
  `metadata.json` temp file.

## Behaviour change

No config or output change. On the happy path the published values, entity schema
and metadata format are identical. The only differences are that the entity write
is now atomic and the publisher loop no longer dies on a transient error.

## Tests

[...]

PR #1007, opened 2026-07-01
feat: Opt-in battery self-identification (learn capacity + round-trip efficiency from HA history)

LesIT1:
## Opt-in battery self-identification (learn capacity + round-trip efficiency from HA history)

Implements the v1 we agreed on in #963: an opt-in, default-off feature that learns the battery's usable capacity and round-trip efficiency from Home Assistant history, built as a sibling of `set_use_adjusted_pv`. Advisory only for now (observe/suggest), so it never changes the values the optimizer uses.

Closes the last hand-entered piece of the battery model. PV is already self-calibrated through `set_use_adjusted_pv` and load is learned by the mlforecaster, but `battery_nominal_energy_capacity`, `battery_charge_efficiency` and `battery_discharge_efficiency` stay whatever number you typed in, forever, even as the pack ages. This lets EMHASS measure them from the signed battery power and measured SoC that are usually already sitting in HA.

### What it does

When `set_use_battery_identification` is on (and `set_use_battery` is true), on the same cadence as the adjusted-PV model it retrieves history for two new sensors, segments it into charge/discharge runs, and estimates:

- **usable capacity** from a total-least-squares (through-origin) fit of integrated throughput against dSoC, with a Theil-Sen cross-check and a block-bootstrap confidence interval;
- **round-trip efficiency** from the discharge/charge slope ratio (which is capacity-independent), cross-checked against a closed-window energy balance.

Two trust tiers, exactly as discussed:

- `observe` (default when the feature is on): writes the estimate, its CI and the assumptions to a JSON under `data_path` and to the log. Nothing published, `plant_conf` untouched.
- `suggest`: also publishes two read-only sensors (`sensor.battery_identified_capacity`, `sensor.battery_identified_round_trip_efficiency`, each carrying the CI and sample counts as attributes) and logs a "configured vs identified" line.

Guardrails keep a bad fit from ever surfacing: minimum deep-cycle counts, a relative-CI ceiling, TLS-vs-Theil-Sen agreement, and hard capacity/efficiency sanity bounds. Anything that fails stays in observe and keeps your configured values. If there isn't enough data (shallow cyclers), it publishes nothing.

### What is and isn't identifiable (the honest part)

From a single AC-side meter plus reported SoC you can identify the round-trip **product** but not split it into separate charge/discharge efficiencies, so the feature never invents an asymmetric split: it sets `eff_chg = eff_dis = sqrt(round_trip)` and says so. Capacity is per *reported* SoC, which is fine because the optimizer plans in those same units. Power-dependent efficiency (thanks @dewi-ny-je for the point on #963), standby draw, and the DC-side split are documented as known limitations and left for a possible v2, along with `auto_apply`.

### Behaviour changes to flag

- **Default off, and off is a true no-op.** When `set_use_battery_identification` is false nothing extra is retrieved, no sensors are published, and the optimizer reads `plant_conf` exactly as today. There's a test pinning that the retrieval variable list is byte-identical when the feature is off.
- **v1 never writes `plant_conf`**, in either tier, on disk or in memory. As agreed, `auto_apply` (in-memory substitution before the optimizer reads it) is deferred to v2.
- **Never raises into an optimization run.** `identify_battery` is wrapped so any unexpected error just logs a warning and returns, leaving the configured values in force.

### A couple of structure decisions worth flagging

- The estimator lives in a new module `src/emhass/battery_identification.py` rather than as methods on `Forecast`. It shares zero `Forecast` state (no var_pv/lat/lon/solar), so the `adjust_pv` pattern is mirrored at the `command_line.py` orchestration layer (`identify_battery`), which is where the reuse targets you named actually live (`is_model_outdated`, `retrieve_home_assistant_data`, the `data_path` convention). `is_model_outdated` gained an optional `label` param so its log lines read correctly for both callers; the PV call site is untouched and keeps its wording (there's a test for that).
- The suggest tier needs to publish a scalar sensor with a small attributes dict, which `post_data` (horizon-shaped) doesn't do cleanly, so there's one additive method `RetrieveHass.post_scalar_sensor`. It reuses the same URL/auth/session and honours `get_data_from_file`; no existing path changes.
- I documented the known limitations in the note in `docs/config.md` rather than `docs/differences.md`, since that file is the legacy-naming mapping table and the adjusted-PV params aren't in it either.

### New config

- `set_use_battery_identification` (bool, default false)
- `battery_identification_trust_tier` (`observe` | `suggest`, default `observe`)
- `battery_identification_model_max_age` (int hours, default 24)
- `sensor_power_battery`, `sensor_battery_state_of_charge` (only retrieved when the feature is on)

All plumbed through `associations.csv`, `config_defaults.json`, `param_definitions.json`, the regenerated `openapi.json`, and `docs/config.md`.

### Tests

New `tests/test_battery_identification.py`:

- recovers known capacity and RTE from synthetic history (ground truth controlled), including a different capacity value and under SoC quantisation noise
- the symmetric split multiplies back to the measured RTE
- segmentation is robust to sub-deadband sign-boundary flicker
- sign convention is auto-detected either way (negating the power channel gives the same answer), and pulsed charge-then-discharge does not spuriously flip
- degenerate near-zero-width CI at the segment floor is treated as low confidence, and an undefined CI fails closed (never publishes)
- insufficient/shallow data and out-of-band capacity are rejected with the configured values kept
- orchestrator: observe writes JSON and publishes nothing and never touches plant_conf; suggest publishes both sensors; a failed fit writes nothing and leaves an existing file untouched; missing sensor columns degrade gracefully
- the `is_model_outdated` label change keeps the PV wording

Full suite stays green.

I can test this on live hardware: an AC-coupled SigenStor with signed battery power and SoC in HA and a deep-cycling strategy that produces the large dSoC segments this needs.

## Summary by Sourcery

Add an opt-in battery self-identification feature that estimates usable capacity and round-trip efficiency from Home Assistant history, persists results, and optionally exposes advisory sensors without affecting the optimizer’s configured battery values.

New Features:
- Introduce a pure battery identification estimator that derives battery usable capacity and round-trip efficiency from signed power and state-of-charge history.
- Add an identify_battery orchestration path that caches identification results, reuses them based on age, and runs alongside live optimizations without modifying plant configuration.
- Expose optional advisory Home Assistant sensors for identified battery capacity and round-trip efficiency in a suggest trust tier.
- Extend Home Assistant retrieval to support a dedicated battery_id mode that fetches battery power and state-of-charge series when battery identification is enabled.

Enhancements:
- Generalize the model staleness check helper to support custom labels so it can be reused for both adjusted PV and battery identification artifacts.
- Add a scalar sensor publishing helper to the Home Assistant retrieval client for posting single-value advisory sensors with metadata attributes.

Documentation:
- Document the new battery identification configuration flags, required sensors, behaviour, and limitations in the configuration reference.

Tests:
- Add unit tests for the battery identification estimator covering ground-truth recovery, robustness to sign conventions and noise, guardrails, and serialization.
- Add orchestration tests to verify JSON persistence, publish behaviour by trust tier, error handling, and interaction with existing results files.
- Add tests to ensure the updated model staleness helper preserves existing adjusted-PV logging while supporting custom labels.


LesIT1:
Thanks, took two of these:

- Added `exc_info=True` to the never-raise handler in `identify_battery` so the traceback is kept in the log while the optimizer stays protected.
- Added a per-call uuid to the atomic temp filename in `_save_battery_identification`. You're right that pid alone can collide if two identify_battery calls overlap in the same process. This matches the same fix I made on #1001 for the entity-file writes.

Two I'm declining, with reasons:

- The `orjson.dumps(...).decode("utf-8")` in `post_scalar_sensor`: this mirrors the existing `post_data` method (retrieve_hass.py:1613), which sends the body the same way. The content-type header is set to application/json explicitly, so bytes vs str makes no functional difference here. I'd rather keep the two post paths consistent than change only the new one.
- Surfacing the `post_scalar_sensor` boolean at the call sites: the method already logs a warning on a non-OK status and an error on an exception internally, so a failed advisory-sensor post is already visible in the log. Logging again at the call site would just double it.

Also added a docs page for the feature (`docs/battery_identification.md`) since it was only described in the config reference before.

davidusb-geek (approved):

PR #1008, opened 2026-07-01
feat: add forecast-calibration action to compare load forecast methods on your own history (#993)

### LINKED ISSUE #993: [Feature Request] a forecast accuracy / calibration report that works for any forecast method, not just the mlforecaster backtest
### The gap

EMHASS can produce a forecast several ways (naive, mlforecaster, typical, csv, list), but there is no built-in way to see how good a forecast actually turned out against the realised values. The only accuracy reporting today is:

* the mlforecaster's own backtest goodness-of-fit (MAE/RMSE/R2/MAPE), which only covers its own ML fit, and
* the R2 on the adjusted-PV bias-correction model.

So anyone on `naive`, `typical`, `csv`, or `list`, including everyone who supplies their own forecast through the `list` method, has no built-in signal on whether that forecast is any good. When a plan looks wrong it is hard to tell whether the optimiser is at fault or the forecast that fed it.

I hit this directly: I run EMHASS with an externally supplied load forecast via the `list` method, wanted to measure its accuracy against actuals, and found there was nothing in EMHASS to do it, so I had to build a separate harness outside the project.

### What I am proposing

A method-agnostic calibration report. It pairs the forecasts EMHASS published with the realised actuals already available from the recorder/history, and reports:

* overall error (MAE, RMSE, bias, MAPE),
* a breakdown by lead time / horizon (does accuracy decay further out?),
* a breakdown by hour of day,
* a breakdown by season or temperature bucket (the "does it track summer vs winter" question),
* a naive-persistence skill score (same hour yesterday / same hour last week) so a user can see whether the forecast actually beats the trivial baseline.

It would work for both load and PV, for any `*_forecast_method`, and would be reporting only. No change to the optimiser or any core constraint, opt-in, no behaviour change for existing runs.

### Why it is worth having

* It closes the "is my forecast any good" blind spot for the methods that currently have no measurement, which is most of them.
* It gives a principled way to compare forecast methods on a given installation instead of guessing.
* It connects to the request that came up around #841, where validating on real forecast-vs-actual data was the open question. A built-in calibration report is the tool that makes that kind of validation routine rather than a one-off script.

### Questions for you before I build anything

I am happy to build this and would route it through the usual process, but I would rather agree the shape first:

1. Where should it live? A new `/action/forecast-calibration` style endpoint, a CLI subcommand, or fold it into the existing backtest plumbing so it is reachable for non-ML methods too?
2. Should it persist a rolling paired log itself (forecast published now, actual filled in later), or compute on demand by re-pulling actuals from the recorder for a past window?
3. Load and PV in one report, or start with one?

If you would rather keep this out of core and as a cookbook recipe plus an external script, that is a reasonable call too and I can go that way instead.

LesIT1:
## What this adds

A new `forecast-calibration` action that measures how well the load forecast methods track your own consumption, so you can see which one to use. This is the reporting tool from #993. It is reporting only: it does not change any optimization, add any config parameter, or save any model or file.

Fixes #993.

## Why

Right now the only forecast accuracy reporting is the mlforecaster's own backtest. Anyone on `naive` or `typical` has no built-in way to check whether their load forecast is any good, or to compare the methods on their own data. This closes that gap for the load forecast, which is the one forecast with no external service behind it.

## How it works

A POST to `/action/forecast-calibration` (empty body works):

1. It retrieves your recent load history (at least about 90 days, since `typical` needs a long window).
2. It splits the history chronologically into three windows: `train`, `test` and `val`.
3. For each method (`naive`, `typical`, `mlforecaster`) it runs a day-ahead walk-forward over the held-out `test` and `val` days. To predict a given day a method only sees history strictly before that day, so there is no look-ahead.
4. It reports MAE, RMSE, R2, MAPE, a persistence skill score and the sample count per method and per split, plus a graph of actual load versus each method over the validation window.

The result renders in the add-on web UI as a graph and a metrics table, using the same injection-dict path as the existing forecast-model actions (no template change).

## Behaviour notes (disclosing the choices)

- Opt-in and a true no-op for existing runs: nothing happens unless you call the action.
- No new config parameter. The window and split sizes are internal constants; `historic_days_to_retrieve` is honoured if it asks for more than the default.
- The `train` column is an in-sample score for `mlforecaster` only. `naive` and `typical` have no fit step so their `train` cell is `N/A` rather than a made-up number.
- The skill score is `1 - method_MAE / naive_MAE`, computed over the days a method and `naive` both cover, so a method that covers fewer days is not flattered by comparing against naive over a different sample.
- `mlforecaster` is fit fresh in memory on the `train` window with a LinearRegression baseline. It never reads a model you trained with `forecast-model-fit`, so the report works even if you have never run a fit, and it will not stall on a slow configured estimator.
- `typical` here is derived from the retrieved history rather than the long-term typical profile, and it needs prior same-month same-weekday history, so on a short window it can cover fewer days than the others. The `n` column makes that visible.

I also pulled the backtest metric calculation out of `machine_learning_forecaster.py` into a shared `utils.compute_forecast_metrics` helper so both the ML backtest and this report use the same code. That is a small net reduction and it is byte-identical to the old inline block (there is a test that locks it).

## What I left for later

I focused on the load forecast as you suggested. I did not include:

- `csv` and `list` load methods. Those are supplied at runtime, so there is nothing to reconstruct on demand without a stored log, which you did not want.
- PV forecast calibration.
- Using your own trained model or the long-term typical profile. The report fits fresh so it always works, but that does mean the numbers are a fair method comparison rather than an exact reading of your production model. Happy to add a "use my saved model" variant later if that is useful.

## Tests

New `tests/test_forecast_calibration.py`: the metrics helper against sklearn and its NaN guards, the RED capability proof, the no-look-ahead invariant, naive matching the production persistence rule, the same-sample skill score, the divide-by-zero guard, and the insufficient-history 400 path. Full suite green locally.

## Summary by Sourcery

Add an on-demand load forecast calibration action that compares built-in forecast methods on Home Assistant history and exposes its report via the CLI and web server.

New Features:
- Introduce a forecast-calibration action and CLI endpoint to compare naive, typical, and machine-learning load forecast methods on the user’s historical data.
- Add a forecast calibration web UI view that renders a validation window plot and metrics table from the calibration results.

Enhancements:
- Extract shared forecast accuracy metric computation into a utils.compute_forecast_metrics helper reused by the ML forecaster backtest and the calibration report.

Documentation:
- Document the new load forecast calibration action, including usage, caveats, and CLI invocation in the forecasts, usage guide, ML forecaster, and developer documentation.

Tests:
- Add unit and integration-style tests for the forecast calibration report, shared metrics helper edge cases, RED capability proof, and key behavioral invariants such as no-look-ahead, naive persistence behavior, fair skill score computation, and insufficient-history handling.

### DIFF
--- a/docs/develop.md
+++ b/docs/develop.md
@@ -264,6 +264,7 @@ curl -i -H 'Content-Type:application/json' -X POST -d {} http://localhost:5000/a
 curl -i -H 'Content-Type:application/json' -X POST -d {} http://localhost:5000/action/forecast-model-fit
 curl -i -H 'Content-Type:application/json' -X POST -d {} http://localhost:5000/action/forecast-model-predict
 curl -i -H 'Content-Type:application/json' -X POST -d {} http://localhost:5000/action/forecast-model-tune
+curl -i -H 'Content-Type:application/json' -X POST -d {} http://localhost:5000/action/forecast-calibration
 curl -i -H "Content-Type:application/json" -X POST -d  '{"csv_file": "heating_prediction.csv", "features": ["degreeday", "solar"], "target": "hour", "regression_model": "RandomForestRegression", "model_type": "heating_hours_degreeday", "timestamp": "timestamp", "date_features": ["month", "day_of_week"], "new_values": [12.79, 4.766, 1, 2] }' http://localhost:5000/action/regressor-model-fit
 curl -i -H "Content-Type:application/json" -X POST -d  '{"mlr_predict_entity_id": "sensor.mlr_predict", "mlr_predict_unit_of_measurement": "h", "mlr_predict_friendly_name": "mlr predictor", "new_values": [8.2, 7.23, 2, 6], "model_type": "heating_hours_degreeday" }' http://localhost:5000/action/regressor-model-predict
 curl -i -H 'Content-Type:application/json' -X POST -d {} http://localhost:5000/action/publish-data
--- a/docs/forecasts.md
+++ b/docs/forecasts.md
@@ -234,12 +234,24 @@ See the [machine learning forecaster](mlforecaster.md) section for more details.
 
 With three load forecast methods available (`naive`, `typical` and `mlforecaster`), it is not always obvious which one tracks your own consumption best. The `forecast-calibration` action produces an on-demand accuracy report to help you pick. It is reporting only: it does not change any optimization and it saves no model or file.
 
-The action retrieves your recent load history (always at least about 90 days, regardless of the `historic_days_to_retrieve` setting, since the typical method needs a long window to be meaningful), splits it chronologically into three windows (`train` / `test` / `val`) and scores every method with a day-ahead walk-forward. To predict a given day, a method may only see history strictly before that day, so there is no look-ahead. The report is shown in the add-on web UI as a graph (actual load versus each method over the validation window) and a metrics table (MAE, RMSE, R2, MAPE, a persistence skill score and the sample count `n`, per method and per split).
+The action retrieves your recent load history (90 days by default, since the typical method needs a long window to be meaningful), splits it chronologically into three windows (`train` / `test` / `val`) and scores every method with a day-ahead walk-forward. To predict a given day, a method may only see history strictly before that day, so there is no look-ahead. The report is shown in the add-on web UI as a graph (actual load versus each method over the validation window) and a metrics table (MAE, RMSE, R2, MAPE, a persistence skill score and the sample count `n`, per method and per split).
 
-Run it with an empty body:
+You can launch it from the add-on web UI with the **Load forecast calibration** button on the advanced page, or over the REST API with an empty body:
 
 ```bash
 curl -i -H 'Content-Type:application/json' -X POST -d '{}' http://localhost:5000/action/forecast-calibration
+```
+
+The three day windows are optional runtime parameters, so you can trade report length against how far back it looks without editing any config. They default to the values above when omitted:
+
+- `calibration_days_to_retrieve`: how much history to pull (default `90`). Lower it for a quicker run or raise it to cover more of the year. If it is too short for the split below, the action returns a clear "not enough history" message instead of a report.
+- `calibration_test_days`: size of the held-out `test` window in days (default `14`).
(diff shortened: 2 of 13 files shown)


LesIT1:
Thanks, took most of these in bd144a80:

- Added the two missing commas in the docs.
- Documented why the ml `last_window` uses `2 * num_lags` (skforecast only needs the last `num_lags` to seed the recursion, the extra is a safety margin).
- Added a test that locks the N/A behaviour when a method has no coverage in a split.

Two I'm declining, with reasons:

- The `plot_df.plot()` / matplotlib concern: EMHASS sets `pd.options.plotting.backend = "plotly"` at import (utils.py), so `.plot()` returns a Plotly figure here, and `.layout` / `.to_html()` work. This mirrors the existing `get_injection_dict_forecast_model_fit` and `_tune` builders, which use the same pattern. I confirmed it renders valid Plotly HTML.
- Surfacing the insufficient-history reason in the 400: the reason is already logged with `logger.error` inside `compute_forecast_calibration`, and the web handler returns `grab_log(...)` in the 400 body, so that message reaches the response and the log already. Happy to thread it more explicitly if you'd prefer, but it isn't lost today.

davidusb-geek:
Thanks for this @LesIT1. 
Now it would be nice if we could launch this report with a button from the webui. Can you add that feature? 
Also the parameters (number of history days) are hard coded right? We could add the possibility to modify these at runtime?

LesIT1:
Thanks David. Both are in the latest push.

The advanced web UI page now has a "Load forecast calibration" button that runs the action.

The day windows are runtime parameters now, so you can change them without editing any config, and they fall back to the current 90/14/14 defaults when omitted:

- calibration_days_to_retrieve (default 90)
- calibration_test_days (default 14)
- calibration_val_days (default 14)

One thing worth flagging: I used a dedicated calibration_days_to_retrieve key for the retrieval window rather than the existing historic_days_to_retrieve. That key is always pre-set (bumped to at least 9 for the fit/predict actions), so reusing it here could not express a shorter calibration window and would have muddied the default. The new key is unambiguous and lines up with the two window params. Docs and tests updated.

davidusb-geek (approved):

Issue #1002, opened 2026-06-23
Bug: `set_battery_first_priority` makes `dayahead-optim` infeasible even on trivial inputs (works in `naive-mpc-optim`)


twobeass:
## Summary

The `set_battery_first_priority` constraint added in #957 makes
`perform_dayahead_forecast_optim` infeasible across a wide range of
otherwise-feasible inputs, including trivial single-binding-constraint
setups. The same model is solved as Optimal by `perform_naive_mpc_optim`
on identical inputs.

Symptom matches none of the documented infeasibility edges (load >
Pd_max is not violated): the issue reproduces with `Pd_max = 20 kW`,
`load = 1 kW` constant, `PV = 0`, `soc_init = 1.0`, `cap = 7700 Wh`.

## Reproducer

Two runs against the same EMHASS v0.17.7 add-on, same `config.json`
(everything off except `set_use_battery: true`, `set_use_pv: true`,
`set_total_pv_sell: false`), same runtime params.

```json
// runtime params
{
  "pv_power_forecast": [/* 96 slots, all Solcast values, or [0]*96 */],
  "load_power_forecast": [1000.0],   // 96 × constant 1 kW also reproduces
  "soc_init": 1.0
}
```

**Result A — `POST /action/dayahead-optim`:**

```
INFO  command_line: Performing day-ahead forecast optimization with soc_init: None  ← runtime soc_init is NOT picked up
WARNING optimization: Optimization failed with status: 'infeasible'.
INFO  optimization: Solving relaxed problem (LP)...
ERROR optimization: Relaxed optimization also failed with status: infeasible
```

Cost function is never evaluated. Solve time ~1.2 s with the relaxed
LP, so it's a structural infeasibility, not a numerical edge.

**Result B — `POST /action/naive-mpc-optim`** (same `config.json`,
same payload plus `"prediction_horizon": 48, "soc_final": 0.1`):

```
INFO  command_line: Performing optimization with soc_init: 1.0  ← runtime soc_init picked up
INFO  optimization: Total value of the Cost function = 0.22
```

Optimal. The plan correctly drains the battery before any grid import
(all import slots have SoC ≤ SoCmin + 1 % tolerance).

**Toggle test:** removing `set_battery_first_priority: true` from
`config.json` and re-running Result A's payload against
`/action/dayahead-optim` → Optimal, Cost = -3.79. So the
infeasibility is exclusively introduced by the flag in the dayahead
codepath.

## What's different between the two codepaths

[...]

PR #1012, opened 2026-07-04
fix: plumb soc_final through dayahead-optim (#1002)

twobeass:
## Summary

Scoped-down follow-up to #1002. Complementary to #1014, not a competitor
— I originally proposed both a plumbing change and a default-fallback
change; @LesIT1's #1014 addresses the infeasibility at its source (the
hard import-gate becomes a soft penalty), which is the better fix and
makes my default-fallback change unnecessary. I've reverted that part.

What stays: **plumb `soc_final` through the dayahead path.** ee56fdf
already routes runtime `soc_init` through
`dayahead_forecast_optim` → `perform_dayahead_forecast_optim` →
`perform_optimization`; this adds the symmetric `soc_final` step for the
same three functions. Currently the only way to pin a terminal SoC on
the dayahead path is to switch to `naive-mpc-optim`, which changes the
horizon and cadence semantics — this closes that gap.

Orthogonal to the constraint model, useful with or without
`set_battery_first_priority`.

## Regression test

`test_dayahead_forecast_optim_honours_soc_final` proves the runtime
`soc_final` actually reaches `perform_optimization` by contrasting two
runs with different targets (`soc_final=0.9` vs `soc_final=0.1`) and
asserting `SOC_opt.iloc[-1]` differs by more than the SoC-recovery
tolerance. No dependency on `set_battery_first_priority`.

## Order-of-merge

Either order works — the two PRs touch different concerns and don't
conflict:

- #1014 first, this after: nothing to do here; the plumb is independent.
- This first, #1014 after: same.

I've kept the two commits on the branch (the initial change + the
scope-down revert) so the reasoning is visible in history. Happy to
squash if you'd prefer a single-commit landing.

Related: #834, #957, #1014.

### DIFF
--- a/src/emhass/command_line.py
+++ b/src/emhass/command_line.py
@@ -1902,7 +1902,10 @@ async def dayahead_forecast_optim(
     """
     _t0 = _time.monotonic()
     soc_init = input_data_dict["params"]["passed_data"].get("soc_init")
-    logger.info(f"Performing day-ahead forecast optimization with soc_init: {soc_init}")
+    soc_final = input_data_dict["params"]["passed_data"].get("soc_final")
+    logger.info(
+        f"Performing day-ahead forecast optimization with soc_init: {soc_init}, soc_final: {soc_final}"
+    )
     # Prepare forecast data with costs, prices, outdoor temp, and GHI
     with stage_timer(input_data_dict["stage_times"], "price_prep", logger):
         df_input_data_dayahead = prepare_forecast_and_weather_data(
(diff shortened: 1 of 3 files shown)


Pwitty1985:
Confirming this matches what I independently found while troubleshooting dayahead-optim infeasibility with set_battery_first_priority: true (see my report and logs on #1002). My testing showed the exact same pattern described here — Initial SOC and Final SOC always ending up identical regardless of battery_target_state_of_charge in config, which lines up with the "no soc_final plumbed through, defaults to soc_init" root cause described in this PR. Switching to naive-mpc-optim as a workaround (which does accept soc_final) resolved it for me, consistent with this fix addressing the underlying gap on the dayahead-optim side specifically. Looking forward to this landing — would let me go back to dayahead-optim instead of the MPC workaround. Happy to test against my config if useful.

davidusb-geek (approved):

Issue #1011, opened 2026-07-04
naive-mpc-optim crashes with KeyError when pv_power_forecast is passed as a runtime list


Pwitty1985:
# Title
naive-mpc-optim crashes with KeyError when pv_power_forecast is passed as a runtime list

## EMHASS version
0.17.8 (Docker image `ghcr.io/davidusb-geek/emhass:latest`)

## Deployment
Docker Compose, standalone (non-add-on), `network_mode: host`, Home Assistant on the same host.

## Describe the bug
Calling `naive-mpc-optim` with a `pv_power_forecast` runtime list crashes with `HTTP 500` / `KeyError: 'sensor.pv_production_watts'` (the entity configured as `sensor_power_photovoltaics`). The traceback shows the failure inside `get_mix_forecast`, which attempts to blend the first element of the supplied forecast list with a live reading of the configured PV sensor via an `alpha`/`beta` weighted average:

```
first_fcst = alpha * df_forecast.iloc[0] + beta * df_now[col].iloc[-1]
```

`df_now` does not contain the expected column, causing the `KeyError`. This occurs even when `alpha`/`beta` are explicitly passed as `1`/`0` — i.e. even when the live-value term should be weighted to zero, the code still appears to evaluate `df_now[col]` unconditionally, so the lookup fails regardless of the weighting.

## Steps to reproduce
1. Configure `sensor_power_photovoltaics: sensor.pv_production_watts` (a working, live-updating Home Assistant sensor — confirmed independently to report correct values).
2. Call `naive-mpc-optim` with a `pv_power_forecast` runtime list (48 values, matching `prediction_horizon: 48`):
   ```bash
   curl -i -H 'Content-Type:application/json' -X POST -d '{"prediction_horizon":48,"soc_init":0.99,"soc_final":0.9,"pv_power_forecast":[0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,750.9,2320.8,3548.2,4629.7,5567.2,6455.6,7214.1,7718.8,7754.8,7439.2,6860.1,6117.7,5382.3,4658.7,3819.6,2930,1922.7,814.5,217.4,15.3,0,0,0,0,0,0,0]}' http://localhost:5000/action/naive-mpc-optim
   ```
3. Observe `HTTP 500` and the traceback below.
4. Repeat with `alpha:1, beta:0` added explicitly — same crash.
[...]

Issue #764, opened 2026-03-21
naive-mpc-optim crashes with KeyError when pv_power_forecast is passed as runtime parameter


hofmannsven-beep:
EMHASS Version: v0.17.1
Description:
When calling naive-mpc-optim via the REST API and passing pv_power_forecast as a runtime parameter (list), EMHASS crashes with a KeyError on the sensor name. The same parameter works correctly with dayahead-optim.

Steps to reproduce:
bashcurl -X POST http://localhost:5000/action/naive-mpc-optim \
  -H "Content-Type: application/json" \
  -d '{"pv_power_forecast": [0,500,1000,2000,3000,2000,1000,500,0,0], "soc_init": 0.5, "soc_final": 0.6}'
```
**Expected behavior:**
EMHASS uses the passed `pv_power_forecast` list directly for MPC optimization, as it does for `dayahead-optim`.

**Actual behavior:**
EMHASS crashes with:
KeyError: 'sensor.your_pv_power_sensor'

File "/app/src/emhass/forecast.py", line 670, in get_mix_forecast
    first_fcst = alpha * df_forecast.iloc[0] + beta * df_now[col].iloc[-1]
                                                      ~~~~~~^^^^^
KeyError: 'sensor.your_pv_power_sensor'

Root cause analysis:
When pv_power_forecast is passed, weather_forecast_method is set to list internally. However, naive-mpc-optim always calls _get_naive_mpc_pv_forecast() → get_power_from_weather() → get_mix_forecast().
get_mix_forecast() tries to combine the forecast list with live sensor data from HA statistics (df_now[col]), but the sensor is not found in the statistics DataFrame → KeyError.

Workaround:
Remove pv_power_forecast from the MPC call and use set_use_pv: true with Open-Meteo + PVLIB instead. This works but is less accurate for non-standard PV configurations (e.g. East-West orientation with multiple roof sections).

Suggested fix:
In get_mix_forecast(), if beta=0 or if the sensor column is not found in df_now, skip the live data combination and use only the forecast list (alpha=1, beta=0).

**EMHASS installation type**
 - Add-on

davidusb-geek:
I suppose that the issue here is actually in your EMHASS configuration.

EMHASS is looking for the default placeholder sensor name because your actual Home Assistant PV sensor entity hasn't been configured in the EMHASS settings.

Please update sensor_power_photovoltaics in your configuration to match your actual HA sensor name.

[...]

PR #1013, opened 2026-07-05
fix: guard get_mix_forecast against a missing sensor column (#764)

LesIT1:
### The bug

`naive-mpc-optim` crashes with `HTTP 500` / `KeyError: <pv sensor>` when `pv_power_forecast` is passed as a runtime list (reported in #764, hit again in #1011). The same crash happens for a runtime `load_power_forecast` list.

The mix correction in `Forecast.get_mix_forecast` blends the latest real sensor value into the first forecast step:

```python
first_fcst = alpha * df_forecast.iloc[0] + beta * df_now[col].iloc[-1]
```

When the forecast comes from a runtime list instead of the sensor, `df_now` has no column for the configured sensor, so `df_now[col]` raises `KeyError` and the whole optimisation aborts.

### The fix

If `df_now` is empty or does not contain the sensor column there is no live value to blend, so return the forecast unchanged and skip the correction. One guard in the shared static method covers both the PV and load callers. When the column is present the blend is unchanged, so there is no behaviour change for the normal sensor-backed path.

### Tests

`tests/test_forecast.py::TestGetMixForecast`:
- missing sensor column returns the forecast unchanged (was a KeyError)
- empty `df_now` returns the forecast unchanged
- present sensor column still blends the first step (guards the happy path)

The first two fail on master with the reported KeyError; all three pass with the fix. Full suite stays green.

## Summary by Sourcery

Guard mix forecast computation against missing or empty sensor data to prevent runtime crashes when forecasts are provided as lists.

Bug Fixes:
- Prevent KeyError in get_mix_forecast when df_now lacks the configured sensor column or is empty.

Tests:
- Add unit coverage for get_mix_forecast to ensure missing sensor columns and empty current-value frames return forecasts unchanged, and valid sensor data still triggers mix correction.

### DIFF
--- a/src/emhass/forecast.py
+++ b/src/emhass/forecast.py
@@ -1070,6 +1070,14 @@ class Forecast:
         if ignore_pv_feedback:
             return df_forecast
 
+        # The mix correction blends the latest real sensor value into the first
+        # forecast step. When the forecast was supplied as a runtime list rather
+        # than read from the sensor, df_now holds no column for that sensor, so
+        # there is no live value to blend. Skip the correction and return the
+        # forecast unchanged instead of raising a KeyError (issue #764).
+        if df_now is None or col not in df_now.columns or df_now.empty:
+            return df_forecast
+
         first_fcst = alpha * df_forecast.iloc[0] + beta * df_now[col].iloc[-1]
         df_forecast.iloc[0] = int(round(first_fcst))
         return df_forecast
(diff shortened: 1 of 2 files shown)


Pwitty1985:
Confirming this fixes a crash I hit independently as well (see my report on #764 / my separate issue with full traceback and config). In my case, set_use_pv: true avoided the KeyError but silently substituted EMHASS's own physical-model PV forecast instead of respecting the pv_power_forecast I was passing at runtime — so the crash was "fixed" but the actual goal (using an external forecast, in my case Solcast data) wasn't achievable. This guard against a missing/empty df_now sounds like it addresses the root cause properly rather than routing around it. Will test with real pv_power_forecast data once this merges and report back.

davidusb-geek (approved):

Issue #1009, opened 2026-07-02
set_battery_first_priority: true invalidates OptimizationCache on every MPC tick (10-20x slowdown)


sidoon:
**Describe the bug**
Setting `set_battery_first_priority: true` in config.json causes the CVXPY problem structure to be rebuilt from scratch on every naive-mpc-optim call instead of reusing the cached optimizer. This results in 10-20x slower optimization times (21s vs 1.6s per call).

Logs show "Building CVXPY problem structure..." on every single MPC tick when `set_battery_first_priority: true`, but only on first call after restart when set to `false`.

**To Reproduce**
1. Set `set_battery_first_priority: true` in config.json
2. Restart EMHASS
3. Call naive-mpc-optim repeatedly (e.g. every 5 minutes via Node-RED)
4. Observe "Building CVXPY problem structure..." in logs on every call (~21s total, optim_solve=10.7s)
5. Set `set_battery_first_priority: false` and restart
6. Call naive-mpc-optim repeatedly
7. Observe "Building CVXPY problem structure..." only on first call after restart (~1.6s total, optim_solve=0.8s)

**Expected behavior**
`set_battery_first_priority` should be evaluated once at cache build time. Subsequent MPC calls should reuse the cached CVXPY problem structure regardless of this setting, as is the case for other boolean parameters like `set_use_battery`.

**Screenshots**
Log with `set_battery_first_priority: true` (every call rebuilds):
INFO in optimization: Building CVXPY problem structure...
INFO in command_line: Optimization completed in 21.5s (top: optim_solve=10.7s, 49%)
INFO in optimization: Building CVXPY problem structure...
INFO in command_line: Optimization completed in 21.5s (top: optim_solve=10.8s, 49%)

Log with `set_battery_first_priority: false` (only first call builds):
INFO in optimization: Building CVXPY problem structure...
INFO in command_line: Optimization completed in 1.6s (top: optim_solve=0.8s, 47%)
INFO in command_line: Optimization completed in 1.6s (top: optim_solve=0.8s, 47%)

**Home Assistant installation type**
- Home Assistant OS

**Your hardware**
- OS: Home Assistant OS
- Architecture: amd64
- CPU: Intel N100 (Gentec N1 mini-PC)
- RAM: 16GB

**EMHASS installation type**
- Add-on

**Additional context**
Configuration used:
- optimization_time_step: 30 min
- prediction_horizon: 96 slots
- num_def_loads: 1
- set_use_battery: true
[...]

PR #1014, opened 2026-07-05
Make set_battery_first_priority a soft penalty so it cannot cause infeasibility (#1002)

### LINKED ISSUE #1002: Bug: `set_battery_first_priority` makes `dayahead-optim` infeasible even on trivial inputs (works in `naive-mpc-optim`)

LesIT1:
## What this fixes

Closes #1002.

With `set_battery_first_priority: true`, `dayahead-optim` comes back non-optimal (infeasible) on trivial inputs where `naive-mpc-optim` solves the identical model fine. Reproduced with PV = 0, a constant 1 kW load, 20 kW of discharge headroom, `soc_init = 1.0` and a terminal SoC target of 0.6: day-ahead is infeasible, naive-MPC is Optimal.

The cause is the import gate I added in #957. It is a hard constraint: `p_grid_pos <= max_from_grid * import_gate`, with the gate forced to 0 in any timestep where the battery is still above its minimum SoC. Day-ahead enforces a terminal SoC target, so when there is little or no PV the battery has to import to recharge back up to that target. But importing raises the SoC above the minimum, which forces the gate shut, which forbids the import. That is a deadlock, so the solver returns infeasible. Naive-MPC has a softer terminal target, so it slips past the same trap.

## The change

I converted the hard gate into a soft penalty. The boolean gate stays, so the per-slot "drain the battery before importing" behaviour is unchanged on a normal flat tariff. But instead of forbidding grid import while the battery is charged, importing in that situation is now penalized in the objective at a large multiple of the import price (an internal constant, 100x). That dwarfs any realistic tariff gradient, so draining first still wins, but the penalty is finite, so the optimizer can always fall back to importing when that is the only feasible option (recharging to a target with no PV, or a load that exceeds the battery's discharge power) instead of returning infeasible.

The penalty is priced off a non-negative clip of the import tariff. Pricing it off the raw signed tariff would turn it into a reward in a negative-price timestep (routine on day-ahead markets) and drive the unbounded penalty variable to infinity, so that clip matters.

## Behaviour change (please read)

This flips a documented behaviour. Before, `set_battery_first_priority` was a hard constraint that could make the optimization infeasible when the load exceeded the battery's discharge power. Now it never makes the problem infeasible; it imports the shortfall and pays the penalty. I think that is the right trade for an opt-in flag, since the old behaviour handed users an infeasible plan with no output at all, but it is a real change so I am calling it out. The default is still off and off is byte-identical to before (no new variable, constraint, or objective term is created).

I updated the docs and the config help text to describe it as a soft penalty rather than a hard constraint. The old test that pinned the infeasible-when-load-exceeds-discharge behaviour is replaced by one that asserts the feasible outcome.

## What this does NOT fix

Not #1009. That one is the 10-20x slowdown when the flag is on, which comes from the boolean gate making the problem a MILP (no warm-start between MPC ticks). This PR keeps the boolean, so it does not help the speed. A boolean-free version would fix both at once, but it can only approximate the per-slot drain-first ordering rather than pin it exactly (the exact "import while charged" term is bilinear), so I kept them separate. Happy to follow up on #1009 that way if you want it.

## Tests

- `test_battery_first_priority_dayahead_feasible_with_soc_target` - the #1002 repro, day-ahead now solves (RED on master).
- `test_battery_first_priority_stays_bounded_with_negative_price` - a negative-price slot stays bounded and optimal.
- `test_battery_first_priority_feasible_when_load_exceeds_discharge` - replaces the old `..._infeasible_...` test.
- `test_battery_first_priority_drains_before_import` - unchanged, confirms per-slot drain-first still holds.

Full suite green.

Thanks to @twobeass for the report and @Pwitty1985 for confirming it.

## Summary by Sourcery

Make battery-first priority an objective penalty instead of a hard constraint so optimizations remain feasible while still preferring to drain the battery before importing from the grid.

New Features:
- Introduce a configurable battery-first import penalty tied to a non-negative clip of the grid import tariff to enforce drain-first behavior without causing infeasibility.

Bug Fixes:
- Prevent day-ahead optimization with battery-first priority from becoming infeasible when recharging to a terminal SoC target without PV.
- Ensure battery-first priority remains feasible when the load exceeds the battery’s maximum discharge power by allowing penalized grid import of the shortfall.
- Avoid unbounded or non-optimal solutions in negative-price tariff timesteps by clipping the import tariff used in the battery-first penalty to non-negative values.

Enhancements:
- Refine the battery-first gate and constraints so importing while the battery is charged is penalized rather than forbidden, preserving drain-first behavior while guaranteeing feasibility.
- Extend optimization parameters to include a non-negative import tariff clip used solely for battery-first penalty calculations and kept in sync across warm-started solves.

Documentation:
- Update configuration documentation and API parameter definitions to describe battery-first priority as a soft penalty that prefers draining before importing while remaining always feasible.

Tests:
- Add and update optimization tests to cover day-ahead feasibility with terminal SoC targets, bounded behavior under negative prices, and feasibility when load exceeds discharge power under battery-first priority.


LesIT1:
Thanks for the review.

I took the second one. The drain-first test now references the configured min SoC (a named `soc_min`) instead of a literal `0.1`, and I noted that the `0.02` band is there to clear the optimizer's internal 1% SoC gate tolerance, so the assertion won't silently drift if that minimum changes. Pushed in bf8dcd5c.

On the penalty factor, I kept it as the internal constant. It is already relative to the tariff: it multiplies `unit_load_cost` per timestep, so it scales with whatever the market's price magnitude is (a market quoted 1000x higher gets a 1000x penalty). The 100x headroom is there to dominate the intra-horizon price gradient rather than the absolute price level, and 100x clears any realistic gradient. Making it configurable would mean plumbing a new setting through associations, config_defaults, param_definitions, openapi and docs for a robustness knob that isn't really an economic dial, and the repo leans toward internal constants for that kind of thing (same reasoning as the curtailment tie-break epsilon). Happy to expose it if you'd prefer, but I don't think it earns a config entry.

twobeass:
Thanks @LesIT1 — this is the right shape for the fix, and I like it
noticeably better than what I put up in #1012.

Attacking the constraint model directly (hard gate → soft penalty)
resolves the two failure modes at their source in one move:

1. The dayahead-with-terminal-SoC deadlock from #1002 — the case I
   originally reported and Pwitty1985 independently confirmed.
2. The load-exceeds-discharge sharp edge pinned by
   `test_battery_first_priority_infeasible_when_load_exceeds_discharge`.

My #1012 only ever addressed (1), and only by defaulting `soc_final`
under the flag — a symptom-level workaround that would have left (2)
still broken. Your penalty framing makes both go away without needing
to touch defaults, which is the cleaner invariant.

Two details from the PR body that I want to call out for
@davidusb-geek's review since they're the load-bearing design calls:

- **Non-negative clip of the import tariff.** Pricing the penalty off
  the raw signed tariff would turn it into a reward in negative-price
  timesteps and drive the unbounded penalty variable to infinity. The
  clip is what keeps the penalty bounded. The new
  `test_battery_first_priority_stays_bounded_with_negative_price` pins
  this — worth pointing out because negative-price slots are routine on
  day-ahead spot markets and this is the kind of edge that would be
  easy to miss.
- **100× multiplier**. Large enough to dwarf any realistic tariff
  gradient, small enough to stay finite. On a flat tariff this gives
  identical drain-first behaviour to the old hard gate; on a rising
  intraday tariff it still overrides "import early" as the unique
  baseline optimum.

**Behaviour-change disclosure**: I agree with the trade you called out
— a documented behaviour change from "hard constraint, sometimes
infeasible" to "soft penalty, always feasible" is the right one for an
opt-in flag. The default-off byte-identical guarantee (no new variable,
constraint, or objective term when off) is exactly what makes this safe
to merge without a bump.

I've scoped my #1012 down to just the `soc_final`-plumb bit — that's
orthogonal to the constraint model (you can pin a terminal SoC on the
dayahead path without switching to naive-mpc-optim) and is useful with
or without the flag. The default-fallback part of #1012 is gone; it
would have conflicted with your fix and is redundant against it. Either
order-of-merge is fine.

+1 from the original reporter — this is the fix.

LesIT1:
Thanks @twobeass, appreciate the look and the +1.

Agreed on keeping #1012 for the soc_final plumbing on the dayahead path. That part is genuinely orthogonal (pinning a terminal SoC there is useful with or without this flag), and merge order doesn't matter either way.

Good call flagging the non-negative clip for @davidusb-geek. That one wasn't obvious. The penalty variable has no upper bound, so pricing it off the raw signed tariff quietly turns it into an unbounded reward the moment a slot goes negative, and the solver runs it to infinity. Clipping the tariff to non-negative is what keeps it bounded, and negative day-ahead slots are common enough that it would bite in practice, so there's a test pinning it.

davidusb-geek (approved):

Issue #916, opened 2026-05-31
[BUG] get_cached_forecast_data stale-cache fallback zeroes out irradiance for Open-Meteo (regression since v0.17.3)


tomvanacker85:
# [BUG] `get_cached_forecast_data` stale-cache fallback zeroes out irradiance for Open-Meteo (regression since v0.17.3)

## Summary

Since **v0.17.3**, when the `weather_forecast_data.pkl` cache does not cover the requested
forecast window, `get_cached_forecast_data` silently serves **zero-filled irradiance data**
instead of deleting the stale cache and triggering a fresh Open-Meteo fetch.  
This causes all PV generation forecasts to be **flat/zero** indefinitely — without any error.

---

## Affected versions

- ✅ v0.17.2 — working correctly  
- ❌ v0.17.3 — regression introduced  
- ❌ v0.17.4 — still broken  
- ❌ v0.17.5 — still broken  

---

## How to reproduce

1. Configure EMHASS with `weather_forecast_method: open-meteo`.
2. Let the system run for ~24 h so that `weather_forecast_data.pkl` is written.
3. Wait until the pkl's timestamp index no longer covers the *current* rolling forecast window
   (this happens every day once the start of the window advances past the cached start time —
   typically ~1–2 days after the cache was written, depending on `delta_forecast_daily`).
4. Trigger any optimization or `weather-forecast-cache` action.
5. **Observed**: all `ghi`/`dni`/`dhi` columns are `0.0`, PV forecast is flat.  
   **Expected**: the stale pkl is deleted and a fresh request is sent to the Open-Meteo API.

---

### The breaking change in `get_cached_forecast_data` (forecast.py ~line 1829)

**v0.17.2 (correct behaviour for Open-Meteo):**

```python
if self.forecast_dates[0] in data.index and self.forecast_dates[-1] in data.index:
    data = data.loc[self.forecast_dates[0] : self.forecast_dates[-1]]
    self.logger.info("Retrieved forecast data from the previously saved cache.")
else:
    self.logger.error("Unable to obtain cached forecast data within the requested timeframe range.")
    self.logger.warning("Removing old forecast cache file. Next optimization will pull data from forecast API...")
    os.remove(w_forecast_cache_path)   # ← deletes stale pkl
    return False                        # ← caller re-fetches from API
```

**v0.17.3+ (broken for Open-Meteo):**

```python
else:
    # Serve best-effort stale cache: reindex to requested dates,
[...]

PR #1017, opened 2026-07-07
fix: weather-forecast-cache never refreshes a stale Solcast cache (#1016)

### LINKED ISSUE #1016: [BUG] weather-forecast-cache never refreshes a stale Solcast cache — _get_weather_solcast returns cached data before fetching (v0.17.3+)
## Summary

When using `weather_forecast_method: solcast` with the two-tier caching pattern (periodic `weather-forecast-cache` action + frequent `cache_only` MPC runs, exactly as recommended in the docs), the Solcast cache **never refreshes once it ages beyond its coverage window**. The `weather-forecast-cache` action — which the docs describe as fetching fresh data — is short-circuited by the read-cache-first logic in `_get_weather_solcast` and silently serves the existing stale cache without ever hitting the Solcast API. The cache becomes effectively **write-once, then frozen forever**, and every subsequent optimization is fed a zero-filled/`ffill`-ed forecast.

This is the Solcast counterpart to the Open-Meteo stale-cache regressions from v0.17.3 (#916, #917) and is in the same code area as #932 — Open-Meteo got a delete-on-stale fix, but the Solcast path still preserves-and-serves stale indefinitely.

Observed on **v0.17.7**; confirmed still present in current `master`.

## Root cause

`Forecast._get_weather_solcast()` reads the cache **before** deciding whether to fetch, and returns early whenever the cache file exists:

```python
async def _get_weather_solcast(self, w_forecast_cache_path: str) -> pd.DataFrame:
    """Helper to retrieve weather data from Solcast or cache."""
    cached_data = await self._get_cached_forecast_or_none(w_forecast_cache_path)
    if cached_data is not None:
        return cached_data          # <-- returns before any fetch
    ...
    # fetch from Solcast ...
    if self.params["passed_data"].get("weather_forecast_cache", False) or \
       self.params["passed_data"].get("weather_forecast_cache_only", False):
        data = await self.set_cached_forecast_data(w_forecast_cache_path, data)
    return data
```

The `weather_forecast_cache` flag (set to `True` by the `weather-forecast-cache` action in `command_line.weather_forecast_cache()`) only controls **saving after a fetch** — it does **not** bypass the early cache return. So the dedicated refresh action can never fetch fresh data while any cache file is present.

For Solcast this is compounded by `get_cached_forecast_data()`: when the cache does **not** cover the requested timeframe, the Open-Meteo branch deletes the stale cache and returns `None` (→ triggers a fresh fetch), but the Solcast branch instead reindexes + zero-fills/`ffill`s and returns a **non-`None`** DataFrame ("Serving best-effort stale data ... Cache preserved to protect API rate limits."). Because it is non-`None`, `_get_weather_solcast` returns it early and never fetches. There is no age check and no delete-on-stale for Solcast, so nothing ever breaks the loop.

Net effect: the only Solcast fetch that can ever happen is the very first one (when no cache file exists). After that the cache is frozen; once it ages past its coverage window (~1 day, given ~2× horizon is cached), every optimization silently runs on a zero-filled forecast.

## Reproduction

1. `weather_forecast_method: solcast`, `weather_forecast_cache_only: true` on MPC runs.
2. Schedule the `weather-forecast-cache` action a few times/day (per docs).
3. Let it run for >1–2 days.

Result: `weather_forecast_data.pkl` stops updating (frozen at the first fetch's timestamp), the daily `emhass_solcast_calls_YYYY-MM-DD.count` file stops being created (zero API calls after day 1), and the PV forecast collapses to a flat/zero-filled line. Manually triggering `weather-forecast-cache` logs *"Cache does not fully cover the requested timeframe. Serving best-effort stale data (reindexed + zero-filled). Cache preserved to protect API rate limits."* and makes **no** API call. Deleting `weather_forecast_data.pkl` and re-triggering the action fetches fresh correctly (until it freezes again).

## Impact

Silent and severe: the optimizer plans against a flat/zero PV forecast, so battery charge/discharge, grid, and deferrable-load scheduling are all wrong, with no error surfaced. In my case the cached forecast was frozen for ~12 days and the "PV plan" was a flat ~2.1 kW line while actual production was a normal bell curve peaking ~5.5 kW.

## Suggested fix

Make the explicit refresh action bypass the cache read so it always fetches fresh (it is quota-guarded by `_solcast_rate_limit_ok`, and it is only invoked a handful of times/day by design):

```python
async def _get_weather_solcast(self, w_forecast_cache_path: str) -> pd.DataFrame:
    """Helper to retrieve weather data from Solcast or cache."""
    force_refresh = self.params["passed_data"].get("weather_forecast_cache", False)
    cached_data = None if force_refresh else await self._get_cached_forecast_or_none(w_forecast_cache_path)
    if cached_data is not None:
        return cached_data
    ...
```

Optionally/additionally, give the Solcast branch of `get_cached_forecast_data()` an age/coverage-based delete-on-stale (as Open-Meteo already has) so a stale cache is refreshed on the next run even outside the explicit action — while still keeping the `_solcast_rate_limit_ok` daily cap to protect quota.

## Related

- #916 / #917 — Open-Meteo irradiance zeroed by the same v0.17.3 stale-cache-reuse change (fixed via delete-on-stale).
- #932 — Solcast stale-cache fallback crash in `get_power_from_weather` (same code path).

## Environment

- EMHASS 0.17.7 (add-on, `ghcr.io/davidusb-geek/emhass:v0.17.7`), Python 3.12, behavior confirmed in current `master`.

SpengeSec:
## What

Make the `weather-forecast-cache` action always fetch a fresh Solcast forecast, instead of short-circuiting on an existing stale cache.

## Why

`_get_weather_solcast` reads the cache **before** deciding to fetch and returns early whenever a cache file exists:

```python
cached_data = await self._get_cached_forecast_or_none(w_forecast_cache_path)
if cached_data is not None:
    return cached_data
```

The `weather_forecast_cache` flag (set by the `weather-forecast-cache` action) only controls **saving after a fetch** — it does not bypass this early return. For Solcast, `get_cached_forecast_data` serves reindexed/zero-filled stale data (non-`None`) when the cache no longer covers the requested timeframe, so the dedicated refresh action can never fetch while any cache file is present.

Result: the Solcast cache is effectively **write-once then frozen**. The only fetch that ever happens is the first (when no file exists); after the cache ages past its coverage window (~1 day, given ~2× horizon is cached), every optimization silently runs on a flat/zero PV forecast. This is the Solcast counterpart of the Open-Meteo stale-cache regressions from v0.17.3 (#916/#917) — Open-Meteo got delete-on-stale, Solcast did not.

Full analysis, reproduction and evidence in #1016.

## Change

Bypass the cache read when `weather_forecast_cache` is set, so the explicit refresh action always fetches fresh. Normal `cache_only` MPC runs are unchanged, and the fetch is still quota-guarded by `_solcast_rate_limit_ok`.

## Testing

Verified on a live add-on (v0.17.7, behavior identical in `master`): with a stale cache present, the `weather-forecast-cache` action served stale and made zero API calls; after this change it fetches fresh and rewrites the cache, while `cache_only` MPC runs continue to read the cache.

Fixes #1016

## Summary by Sourcery

Bug Fixes:
- Prevent the weather-forecast-cache action from returning stale Solcast cache data by bypassing cache reads when a forced refresh is requested.

### DIFF
--- a/src/emhass/forecast.py
+++ b/src/emhass/forecast.py
@@ -563,7 +563,21 @@ class Forecast:
 
     async def _get_weather_solcast(self, w_forecast_cache_path: str) -> pd.DataFrame:
         """Helper to retrieve weather data from Solcast or cache."""
-        cached_data = await self._get_cached_forecast_or_none(w_forecast_cache_path)
+        # The explicit `weather-forecast-cache` action sets weather_forecast_cache
+        # and is meant to refresh the cache from the Solcast API. Reading the cache
+        # first here would return a stale-but-present cache early (for Solcast,
+        # get_cached_forecast_data serves reindexed/zero-filled stale data instead
+        # of returning None), so the refresh action would never fetch and the cache
+        # would freeze permanently once it aged past its coverage window. Bypass the
+        # cache read for that action so it always fetches fresh; normal cache_only
+        # MPC runs still read the cache. The fetch stays quota-guarded by
+        # _solcast_rate_limit_ok.
+        force_refresh = self.params["passed_data"].get("weather_forecast_cache", False)
+        cached_data = (
+            None
+            if force_refresh
+            else await self._get_cached_forecast_or_none(w_forecast_cache_path)
+        )
         if cached_data is not None:
             return cached_data
         # Incompatible/stale cache was discarded (issue #932); fall through


davidusb-geek:
@SpengeSec Thanks for this fix!
Currently some failing tests, can you take a look at those?

src/emhass/forecast.py
```diff
@@ -563,7 +563,21 @@ def _parse_pv_quantile_bias(self) -> float:
...
+        # The explicit `weather-forecast-cache` action sets weather_forecast_cache
+        # and is meant to refresh the cache from the Solcast API. Reading the cache
+        # first here would return a stale-but-present cache early (for Solcast,
+        # get_cached_forecast_data serves reindexed/zero-filled stale data instead
+        # of returning None), so the refresh action would never fetch and the cache
+        # would freeze permanently once it aged past its coverage window. Bypass the
+        # cache read for that action so it always fetches fresh; normal cache_only
+        # MPC runs still read the cache. The fetch stays quota-guarded by
+        # _solcast_rate_limit_ok.
+        force_refresh = self.params["passed_data"].get("weather_forecast_cache", False)
```
JoshC1994, line 575:
```suggestion
        force_refresh = self.params.get("passed_data", {}).get("weather_forecast_cache", False)
```

JoshC1994 (commented):
Looks like the test failures are unrelated to this PR and caused by `test_is_electric_load_excludes_load_from_grid_balance` in `test_optimization.py`  where it is comparing grid import against a hardcoded 2000W threshold while running against a real-clock-dependent baseload 'typical', which exceeded that threshold on this run's date.
If it ran again it would likely pass.  

Probably needs to be put under another PR as a bug fix. 

suggestion tests/test_optimization.py line 3812
```
only_l1_slots = (p_load0 == 0) & (p_load1 > 0)
if only_l1_slots.any():
    grid_in_l1_only = p_grid_pos[only_l1_slots]
    # Compare against actual baseload at these slots instead of a fixed
    # ceiling, since self.p_load_forecast varies with the real calendar
    # date/time (typical-curve method).
    baseline_at_slots = self.p_load_forecast[only_l1_slots]
    leaked_power = (grid_in_l1_only - baseline_at_slots.values).max()
    self.assertLess(
        leaked_power,
        500,  # tolerance for solver/rounding noise, not baseload magnitude
        "Non-electric load (load 1) appears to be pulling from the grid",
    ) 


```

davidusb-geek (approved):

PR #1025, opened 2026-07-14
fix: continuous time-of-day encoding for the adjusted PV forecast (removes hour-boundary sawtooth)

scruysberghs:
## Problem

With `set_use_adjusted_pv` enabled and a sub-hourly `optimization_time_step` (e.g. 15 min), the adjusted PV forecast shows a sawtooth: a discrete jump at every hour boundary, with smooth behaviour in between.

Root cause: `adjust_pv_forecast_data_prep` / `adjust_pv_forecast_predict` feed the regressor the default `add_date_features` set, which includes the **raw integer `hour` (0-23)**. That feature is piecewise constant: within an hour it does not move, at every `:00` it steps by 1. A linear model (the default `LassoRegression`) therefore produces a discontinuity of exactly its hour-coefficient at every hour boundary.

The effect is amplified by collinearity: `solar_azimuth` advances almost linearly (~15 degrees/hour), so the fit can trade a large smooth azimuth term against a large stepped hour term. Both cancel on hourly averages (small train-MSE difference), but at 15-min resolution the cancellation breaks: within the hour the azimuth term keeps rising while `hour` is flat, then `hour` steps at `:00`. Result: a sawtooth with the amplitude of the hour coefficient.

## Real-world measurement

On a production system (15-min timestep, LassoRegression, `historic_days_to_retrieve` window of recent data):

- Published adjusted forecast over a 48 h horizon (192 steps): **all 23 hour-boundary deltas negative** (mean −641 W) while within-hour deltas averaged +175 W.
- Recovering the live model's coefficients from its own input/output via least squares (the model is exactly linear; residual RMS 81 W on an ~7 kWp system) gives:

  | feature | coefficient |
  |---|---|
  | forecast | 1.01 |
  | solar_elevation | −3.9 W/deg |
  | solar_azimuth | **+54.9 W/deg** (≈ +823 W/h smooth ramp) |
  | hour | **−894 W per hour tick** |

  i.e. the azimuth/hour cancellation pair described above, producing a ~900 W sawtooth on a ~7 kWp installation. Since the adjusted forecast feeds the LP, the battery/deferrable plan inherits the artifact.

## Fix

- Encode time of day continuously: replace the raw `hour` feature with `hour_sin` / `hour_cos` computed from the **fractional** hour (`hour + minute/60`), in both the fit and predict paths (new `Forecast.add_cyclic_hour_features`). The other calendar features are unchanged; they only step at midnight where PV is zero anyway.
- Backward compatibility: a saved `adjust_pv_regressor.pkl` trained on the old feature set makes scikit-learn raise on predict (feature-name mismatch). `adjust_pv_forecast` now catches this, re-fits once with the current feature set, and retries, so existing installs self-heal on upgrade instead of erroring until the pickle is deleted.

## Tests

- `test_pv_forecast_adjust`: asserts the raw `hour` column is gone and `hour_sin`/`hour_cos` are present.
- New `test_pv_forecast_adjust_no_hour_boundary_jump`: regression test that a model with weight on the time-of-day features cannot introduce hour-boundary jumps on a constant input curve (fails on the old integer-hour encoding).
- New `test_adjust_pv_forecast_stale_feature_model_refit`: the predict-fails → re-fit → retry path for pre-existing pickles.

Validated against 10 days of real production data (measured PV + forecast history): with the cyclic encoding the predicted curve's hour-boundary steps are statistically identical to within-hour steps (no sawtooth), while the old encoding reproduces the measured ~900 W jumps.

## Summary by Sourcery

Encode time-of-day features continuously in the adjusted PV forecast to eliminate hour-boundary discontinuities and ensure existing persisted models self-heal after the feature-set change.

New Features:
- Add cyclic hour features (hour_sin and hour_cos computed from fractional hour) as the time-of-day encoding for PV forecast adjustment.

Bug Fixes:
- Prevent sawtooth artifacts in the adjusted PV forecast at sub-hourly time steps by removing the raw integer hour feature from the regression model inputs.
- Handle failures when predicting with legacy adjusted PV models trained on outdated feature sets by automatically re-fitting once and retrying, falling back to the unadjusted forecast if recovery fails.

Enhancements:
- Align training and prediction feature sets for adjusted PV forecasts using calendar features without raw hour plus cyclic hour encoding and solar angles.

Tests:
- Extend PV forecast adjustment tests to assert use of cyclic hour features instead of raw integer hour.
- Add regression test ensuring the adjusted PV forecast cannot introduce artificial jumps at hour boundaries on a constant input curve.
- Add test verifying that stale adjusted PV models trained on outdated feature sets are re-fitted automatically when prediction fails.


scruysberghs:
Both points addressed:

- The re-fit path now catches only `ValueError` (what scikit-learn raises on the feature-name mismatch when a saved model predates the feature change); other exception types propagate again.
- `add_cyclic_hour_features` now validates that the DataFrame has a `DatetimeIndex`, consistent with `utils.add_date_features`, with a test covering the non-datetime case.

scruysberghs:
Side-by-side validation on a real installation (7 kWp, 15-min timestep, LassoRegression): the production add-on on current code and a shadow instance built from this PR received the **same rendered runtime payload at the same moment**, both fitting on the same live history. The shadow's result was read back via the new `GET /api/v1/plan` endpoint.

adjusted PV forecast, current code vs this PR

Quantified on the daytime horizon of this run:

| | mean delta at :00 | mean delta within hour | implied hour jump | negative :00 deltas |
|---|---|---|---|---|
| current code | −667 W | +200 W | **−867 W** | 21/21 |
| this PR | −1 W | +3 W | **−3.6 W** | 15/28 (random sign) |

Same downward correction of the raw forecast in both, but the hour-boundary discontinuity is gone with the cyclic encoding.

davidusb-geek (approved):

davidusb-geek:
Hi @scruysberghs, can you please rebase and solve the current conflicts here?
Tests are passing fine

PR #1028, opened 2026-07-17
fix: bound set_nodischarge_to_grid export by PV surplus, not raw PV (#795 regression)

### LINKED ISSUE #1023: Regression: set_nodischarge_to_grid battery-to-grid loophole (#795) reintroduced by #981 on AC-coupled systems
## Description

PR #981 (shipped in 0.17.8) relaxed the `set_nodischarge_to_grid` constraint for AC-coupled systems to *"bound grid export to available PV"*. That is exactly the constraint issue #795 identified as logically flawed: `P_export <= P_PV` only caps the **amount** of export, it does not constrain the **source** of the exported energy.

The optimizer exploits this: it discharges the battery to cover the *entire* local load, which frees the PV production for export. Physically and economically this is indistinguishable from discharging the battery straight into the grid — the constraint simply cannot see it.

My case is a stronger demonstration than the original #795, because in every affected timestep **`load > PV`** — there is no PV surplus whatsoever — and the plan still exports:

| time  | PV (W) | load (W) | p_batt (W, + = discharge) | p_grid (W, − = export) |
|-------|-------:|---------:|--------------------------:|-----------------------:|
| 20:00 |    328 |      800 |                  **+800** |               **−328** |
| 20:30 |    247 |      800 |                  **+800** |               **−247** |
| 21:00 |    103 |      837 |                  **+837** |               **−103** |
| 21:30 |     20 |      837 |                  **+837** |                **−20** |

Note that `p_grid == -P_PV` in every slot — the signature of the `P_export <= P_PV` bound being the binding constraint. The battery discharges the full load rather than the deficit. The pattern repeats identically on both days of the 48 h horizon, so this is systematic and not a solver artefact.

## How the regression happened

- **#795** — original report: the battery drains itself to cover the load so that PV can be exported.
- **#796** (0.17.3) — fixed by a strict `E <= D` coupling (battery discharge limited to grid import).
- **#936** — that fix over-constrained AC-coupled systems: it also blocked legitimate battery → local-load discharge, causing infeasibility.
- **#981** (0.17.8) — relaxed the rule to `export <= PV` for AC-coupled systems, which reopens #795.

The complaint in #936 was entirely valid; the remedy is what reintroduces the old flaw. Both requirements can be met simultaneously — see below.

## Reproduction

1. AC-coupled system: `inverter_is_hybrid: false`, `set_use_battery: true`, `set_nodischarge_to_grid: true`, `costfun: profit`, `optimization_time_step: 30`, `maximum_power_to_grid: 9000`.
2. Pass `load_power_forecast`, `load_cost_forecast` and `prod_price_forecast` as runtime lists, together with `soc_init` (battery not full — SoC ≈ 32 % here).
3. Run `dayahead-optim` over an evening where PV is declining but still non-zero and `load > PV`.
4. Inspect `p_batt_forecast` / `p_grid_forecast`: the battery discharges the full load and `p_grid == -P_PV`.

## Expected behaviour

With `set_nodischarge_to_grid: true`, the battery should discharge only the net deficit, and no export should occur while `load >= PV`:

```
p_batt = max(0, load − PV)      ->  20:00: 800 − 328 = 472 W
p_grid = 0
```

## Root cause

`src/emhass/optimization.py`, AC-coupled branch (`inverter_is_hybrid == False`), as introduced by #981:

```python
constraints.append(self.vars["p_grid_neg"] + p_pv >= 0)
```

i.e. **`export <= p_pv`**.

The bound is on PV **production**, not PV **surplus**. Substituting the 20:00 timestep from the table above:

```
p_pv = 328,  load = 800,  p_grid_neg = -328
-328 + 328 = 0  >=  0        # constraint satisfied, exactly at the bound
```

The constraint sits precisely on its boundary, so the optimizer pushes export to the maximum it permits — the entire PV output. To make that possible, the battery has to serve the whole load. The constraint never notices, because it only limits the *magnitude* of the export and never the *source* of the energy behind it.

The previous `E <= D` coupling did prevent this, but it was too strict for the reason given in #936: at `grid == 0` (perfect self-consumption) `D == 0`, which forced `E == 0` and forbade the battery from discharging at all.

## Proposed fix

Bound the export by the PV **surplus** rather than by PV production:

```python
# p_pv and the base load are forecast parameters, so max(0, ·) is a per-timestep
# constant -> the constraint stays affine, no binaries or big-M needed
surplus = np.maximum(0.0, p_pv - p_load)
constraints.append(self.vars["p_grid_neg"] + surplus >= 0)
```

Equivalently, and perhaps more explicitly, constrain the battery instead of the grid:

```python
deficit = np.maximum(0.0, p_load - p_pv)
constraints.append(self.vars["p_sto_pos"] <= deficit)
```

With the curtailment path, subtract `p_pv_curtailment` from `p_pv` before taking the maximum, exactly as #981 already does for the current bound.

Checked against the plan above:

- **20:00** — `deficit = 800 − 328 = 472`, so the battery may discharge at most 472 W, giving `p_grid = 800 − 328 − 472 = 0`. No export.
- **PV > load** (a genuine surplus) — `deficit = 0`, so the battery does not discharge, and the surplus is free to be exported.

This satisfies **both** earlier issues at once: the battery may still serve the local load without restriction (#936), while battery-to-grid flow becomes structurally impossible (#795). Bounding `p_grid` alone will always be insufficient, because it leaves the source of the exported energy unconstrained.

## Impact

This is not only a constraint violation but a strictly worse solution. In the plan above, the import price at 21:00 is 0.343 currency/kWh while the export (spot) price at 20:00 is 0.160. The optimizer therefore sells battery energy at 0.160 that could have offset a 0.343 import one hour later — a loss of roughly **0.18 / kWh**, before even accounting for round-trip efficiency.

## Environment

- **EMHASS:** 0.17.9 (Home Assistant add-on)
- **Home Assistant:** core-2026.7.2 on Home Assistant OS 17.3
- **Architecture:** aarch64 (Odroid C4)

```json
"set_use_battery": true,
"set_nodischarge_to_grid": true,
"set_nocharge_from_grid": false,
"inverter_is_hybrid": false,
"costfun": "profit",
"maximum_power_to_grid": 9000,
"battery_charge_power_max": 3680,
"battery_discharge_power_max": 3680,
"battery_nominal_energy_capacity": 15800,
"optimization_time_step": 30,
"delta_forecast_daily": 2
```

## Workaround

Setting `maximum_power_to_grid: 0` blocks the leak, but it also prevents legitimate export of genuine PV surplus, so it is not a real solution.

petrister:
Fixes the `set_nodischarge_to_grid` battery-to-grid loophole reported in #1023 — a regression of #795.

## Problem

The AC-coupled branch introduced in #981 bounds grid export by raw PV:

```python
constraints.append(self.vars["p_grid_neg"] + p_pv >= 0)   # export <= p_pv
```

This caps the *amount* of export but not its *source*. The optimizer discharges the battery to cover the entire local load, which frees the PV for export — functionally battery-to-grid, exactly the loophole #795 described. `p_grid` ends up equal to `-P_PV` even when no surplus exists.

Example from a real `dayahead-optim` (AC-coupled, `inverter_is_hybrid: false`, `set_nodischarge_to_grid: true`) — `load > PV` in every slot, so there is no surplus, yet it exports:

| time  | PV (W) | load (W) | p_batt (W, + = discharge) | p_grid (W, − = export) |
|-------|-------:|---------:|--------------------------:|-----------------------:|
| 20:00 |    327 |      508 |                  **+508** |               **−327** |
| 20:30 |    241 |      508 |                  **+508** |               **−241** |
| 21:00 |     94 |      545 |                  **+545** |                **−94** |

## Fix

Bound export by the PV **surplus**, `max(0, PV − load)`, rather than raw PV. The surplus is held in a dedicated non-negative `Parameter` (`param_export_ceiling`), mirroring the existing `param_load_cost_pos` so it stays correct across warm-started re-solves (a `cp.maximum` of parameters is not DCP-valid in this `>= 0` constraint). The curtailment branch subtracts `p_pv_curtailment` from the same ceiling.

```python
# build time
self.param_export_ceiling = cp.Parameter(self.num_timesteps, nonneg=True, name="export_ceiling")
# per solve
self.param_export_ceiling.value = np.maximum(np.asarray(p_pv, float) - np.asarray(p_load, float), 0.0)
# constraint (AC-coupled)
constraints.append(self.vars["p_grid_neg"] + self.param_export_ceiling >= 0)
```

This keeps battery-to-load fully allowed (the requirement from #936) while making battery-to-grid structurally impossible (the requirement from #795). Bounding `p_grid` alone can never suffice, because the *source* of the exported energy stays unconstrained.

## Verified locally

Applied to a running 0.17.9 add-on. Same evening after the fix — the battery discharges only the deficit and `p_grid == 0`:

| time  | PV (W) | load (W) | p_batt (W) | p_grid (W) |
|-------|-------:|---------:|-----------:|-----------:|
| 20:00 |    327 |      508 |       +181 |          0 |
| 20:30 |    241 |      508 |       +267 |          0 |
| 21:00 |     94 |      545 |       +451 |          0 |

Genuine PV surplus is unaffected: at 19:00 (PV 502 > load 307) the battery stays idle and 195 W is exported as before. `optim_status` remains `Optimal`, no infeasibility.

Closes #1023.

## Summary by Sourcery

Bound grid export by PV surplus rather than raw PV when set_nodischarge_to_grid is enabled for AC-coupled systems, preventing indirect battery-to-grid export while preserving battery-to-load behavior.

Bug Fixes:
- Fix AC-coupled no-discharge-to-grid regression that allowed battery energy to reach the grid via freeing PV for export.
- Correct curtailment export constraint to apply the PV surplus ceiling minus curtailment instead of raw PV.

Enhancements:
- Introduce a dedicated non-negative export ceiling parameter derived from PV minus load and wire it into both initial and warm-started optimization runs.

### DIFF
--- a/src/emhass/optimization.py
+++ b/src/emhass/optimization.py
@@ -196,6 +196,16 @@ class Optimization:
         self.param_load_cost_pos = cp.Parameter(
             self.num_timesteps, nonneg=True, name="load_cost_pos"
         )
+        # Non-negative PV surplus, max(0, PV - load), used as the export ceiling for
+        # set_nodischarge_to_grid on AC-coupled systems. Bounding export by raw PV
+        # (pre-fix behaviour) lets the battery reach the grid indirectly: it covers
+        # the whole load so PV is freed for export (regression of #795, reintroduced
+        # by #981). Bounding by the surplus blocks battery-to-grid while still
+        # allowing battery-to-load. Dedicated Parameter (not max() baked in at build
+        # time) to stay correct across warm-started re-solves.
+        self.param_export_ceiling = cp.Parameter(
+            self.num_timesteps, nonneg=True, name="export_ceiling"
+        )
         self.param_prod_price = cp.Parameter(self.num_timesteps, name="prod_price")
 
         # Per-deferrable-load cost override parameters. When the user supplies a
@@ -1858,17 +1868,23 @@ class Optimization:
         # No discharge to grid: prevent battery energy from reaching the grid. Hybrid inverters
         # prioritise PV to the load, so the battery cannot discharge while PV exports (strict E<=D, #796).
         # AC-coupled systems can, so E<=D wrongly forbids battery-to-load during export and makes the
-        # solve infeasible when a large SoC must be shed (#936). For them bound grid export to the
-        # available PV instead (net of curtailment), blocking battery-to-grid while allowing battery-to-load.
+        # solve infeasible when a large SoC must be shed (#936). For them bound grid export to the PV
+        # *surplus* (max(0, PV - load), param_export_ceiling), not raw PV: bounding by raw PV lets the
+        # battery cover the entire load and free PV for export, i.e. battery-to-grid through a PV detour
+        # (#795, reintroduced by #981). Bounding by the surplus blocks battery-to-grid while still
+        # allowing battery-to-load. Curtailment further lowers the exportable surplus.
         if self.optim_conf["set_nodischarge_to_grid"]:
             if self.plant_conf["inverter_is_hybrid"]:
                 constraints.append(E <= D)
             elif self.plant_conf["compute_curtailment"]:
                 constraints.append(
-                    self.vars["p_grid_neg"] + p_pv - self.vars["p_pv_curtailment"] >= 0
+                    self.vars["p_grid_neg"]
+                    + self.param_export_ceiling
+                    - self.vars["p_pv_curtailment"]
+                    >= 0
                 )
             else:
-                constraints.append(self.vars["p_grid_neg"] + p_pv >= 0)
+                constraints.append(self.vars["p_grid_neg"] + self.param_export_ceiling >= 0)
 
         # Dynamic Power Limits (Ramp Rate)
         if self.optim_conf["set_battery_dynamic"]:
@@ -3783,6 +3799,7 @@ class Optimization:
             self.param_load_forecast = cp.Parameter(current_n, name="load_forecast")
             self.param_load_cost = cp.Parameter(current_n, name="load_cost")
             self.param_load_cost_pos = cp.Parameter(current_n, nonneg=True, name="load_cost_pos")
+            self.param_export_ceiling = cp.Parameter(current_n, nonneg=True, name="export_ceiling")
             self.param_prod_price = cp.Parameter(current_n, name="prod_price")
             self.param_cost_per_load = [
                 cp.Parameter(current_n, name=f"cost_per_load_{k}")
@@ -3958,6 +3975,9 @@ class Optimization:
         self.param_load_forecast.value = p_load
         self.param_load_cost.value = unit_load_cost
         self.param_load_cost_pos.value = np.maximum(np.asarray(unit_load_cost, dtype=float), 0.0)
+        self.param_export_ceiling.value = np.maximum(
+            np.asarray(p_pv, dtype=float) - np.asarray(p_load, dtype=float), 0.0
+        )
         self.param_prod_price.value = unit_prod_price
 
         # Per-load cost forecast overrides. Default each load's per-timestep cost
--- a/src/emhass/optimization.py
+++ b/src/emhass/optimization.py
@@ -59,6 +59,17 @@ CURTAILMENT_TIEBREAK_EPS = 1e-7
 # penalty to genuinely avoidable import, so an aggressive factor is safe.
 BATTERY_FIRST_IMPORT_PENALTY_FACTOR = 100.0
 
+# Soft terminal-SoC target, the same treatment #1002 gave set_battery_first_priority.
+# A hard equality on the horizon's net energy change turns the solve infeasible
+# whenever the requested soc_final simply cannot be reached. That collides with
+# set_nodischarge_to_grid on AC-coupled systems: when PV already covers the load a
+# large SoC shed has no local deficit to discharge into, and export is (correctly)
+# closed off, so no schedule exists (#936 vs #795). Enforcing the target through
+# non-negative slacks priced far above any realistic tariff keeps it met exactly
+# whenever that is possible, while a contradictory target relaxes to the closest
+# reachable SoC instead of returning infeasible.
+SOC_FINAL_DEVIATION_PENALTY_FACTOR = 100.0
+
 
 class Optimization:
     r"""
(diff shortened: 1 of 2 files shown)


petrister:
Heads-up on CI and the interaction with #936, so it isn't a surprise — and because I think it's a design call for you rather than something I should decide unilaterally.

This change will fail `test_nodischarge_to_grid_ac_coupled_feasible_issue936`. That's not an oversight in the test; it's the crux of the conflict.

That test's scenario (PV = 3000 W > load = 600 W every step, forced SoC shed 0.9 → 0.1, `set_nodischarge_to_grid=True`) is only feasible **because the battery is allowed to export**. PV already covers the load, so there is no local deficit to discharge into — the only way to shed SoC is battery → grid. Reproduced in isolation:

```
old constraint (p_grid_neg + p_pv     >= 0):  optimal, export 9.0 kWh, battery discharge 2.7 kWh (all to grid)
this PR        (p_grid_neg + surplus  >= 0):  infeasible
```

So the "feasible" outcome of that test *is* the #795 leak — 2.7 kWh of battery energy leaving through the grid. The #936 fix (export ≤ raw PV) effectively traded #795 correctness for feasibility.

The two requirements are mutually exclusive in that scenario: you can't both forbid battery → grid **and** force a SoC shed that has no local deficit to absorb it. Something has to give.

The clean resolution that satisfies both #936 and #795, I think, is:
- forbid battery → grid (this PR), **and**
- treat `soc_final` as a soft target, so a contradictory shed relaxes instead of going infeasible.

The optimizer would then keep the battery off the grid and still return a plan, getting as close to `soc_final` as the no-export constraint allows. That also lines up with the 0.17.9 direction ("Make set_battery_first_priority a soft penalty so it cannot cause infeasibility").

Happy to take this wherever you prefer:
- update `test_nodischarge_to_grid_ac_coupled_feasible_issue936` to assert the no-leak behaviour, plus add a regression test for the deficit case from #1023 (`load > PV`, battery discharges the full load while PV is exported — the original report), and/or
- extend this PR to soften `soc_final`.

I didn't want to rewrite that test or touch `soc_final` on my own, since it's really your call.

For context (#1023): on an AC-coupled system with `set_nodischarge_to_grid: True` and spot-price feed-in, the current behaviour discharges the battery to cover the whole evening load while exporting the PV — selling battery energy at ~0.15 that would have offset a ~0.33 import a couple of hours later. A guaranteed loss, and not what `set_nodischarge_to_grid` implies.

davidusb-geek:
Hi @petrister, thanks for contributing.
There are currently some failing tests, if you can fix them I should be able to merge you contribution

petrister:
The two failures are fixed; all tests pass locally (686 passed, ruff clean).

**Curtailment (#342 tie-break):** folding `p_pv_curtailment` into the surplus ceiling also capped curtailment itself at the surplus — unrelated, and it removed the placement freedom the tie-break relies on. Split into two independent bounds, giving `export <= min(surplus, PV - curtailment)`.

**#936:** I went with the soft-`soc_final` route we discussed. The equality on the horizon's net energy change is now carried by non-negative slacks priced far above any realistic tariff — the same treatment #1002 gave `set_battery_first_priority`. It's still met exactly whenever a schedule exists for it, but a contradictory target relaxes instead of failing the solve. `test_nodischarge_to_grid_ac_coupled_feasible_issue936` passes **unchanged** as a result.

One test did need changing: `test_nodischarge_to_grid_hybrid_still_blocks_discharge_issue936` asserted infeasibility as a proxy for "E<=D is still applied". That proxy no longer holds, so it now asserts the invariant directly — no battery discharge in any exporting timestep. Happy to revert that if you'd rather keep `soc_final` hard.

Also added a regression test for #1023 as originally reported (load 1000 W > PV 400 W): verified RED on the raw-PV bound and GREEN with the ceiling.

davidusb-geek (approved):

Issue #1020, opened 2026-07-13
[Feature Request] Confirm the public contract for multi-inverter optimization


ffrog8:
## Summary

Before preparing a multi-inverter pull request, I would like maintainer confirmation of the public behavior and compatibility boundaries.

The proposal allows one EMHASS optimization to model multiple battery or hybrid inverters while keeping the existing single-inverter path as the default.

## Proposed public contract

- `multi_inverter_mode: single` remains the default and preserves the existing top-level inverter configuration.
- In `multi` mode, `hybrid_inverter_config_1..N` are the active inverter definitions. Top-level inverter-local values are fallback configuration for single mode only and are not inherited by inverter 1.
- Site-level tariffs, grid limits, topology flags, publishing settings, and secrets remain top-level.
- Public inverter numbering is one-based. Inverter 1 may continue using the legacy internal variables as a compatibility adapter, but its effective values come only from indexed config 1 or documented block defaults.
- Existing result names, signs, and units remain unchanged. In multi mode, unsuffixed inverter quantities are site totals and suffixed quantities such as `P_batt_1` are inverter-local. No public `_0` or `*_total` columns are added.
- `P_batt > 0` remains discharge, `P_grid > 0` remains import, and result/CSV SOC remains a `0..1` fraction while Home Assistant SOC publishing remains percent.
- Supported multi-inverter shapes are battery-only and `split_pv` hybrid fleets. Both require `battery_mode_lock: site`.
- A `split_pv` fleet may mix PV-enabled and PV-disabled hybrid battery blocks. Each PV-enabled block requires its own nested runtime `pv_power_forecast`; a PV-disabled block does not.
- Runtime `multi_inverter_mode: single` intentionally selects the legacy path even when indexed static blocks exist. Nested inverter runtime blocks remain invalid while single mode is active.
- `battery_soc_balance_cost` is a soft site-level objective, defaults to `0.0`, and is inactive when disabled or when fewer than two batteries are active.
- Existing upstream soft-penalty behavior for `set_battery_first_priority` must be preserved; multi-inverter support must not restore the former hard-infeasibility behavior.

## Branch review findings

A review of the current branch found these behaviors that require correction before a PR:

- **R1:** When indexed inverter 1 omits an optional inverter-local value, the legacy top-level value can survive. It should instead use the indexed value or its documented block default.
[...]

PR #1032, opened 2026-07-18
feat: support N stationary batteries via the deferrable-loads array pattern (#610 phase 1)

LesIT1:
## What this does

Implements phase 1 of the multi-battery plan agreed in discussion #610: N stationary batteries, opt-in through a new `number_of_batteries` plant parameter (default 1). The shape follows the greenlight exactly:

- The existing battery parameters (capacity, charge/discharge power limits, efficiencies, SOC min/max/target, stress cost, SOC deficit/surplus thresholds and costs) now accept a list of one entry per battery. Scalars stay accepted and broadcast to every battery, so existing configs need no changes.
- `weight_battery_charge`/`weight_battery_discharge` accept a nested form (one entry per battery, each a scalar or a time series). A flat scalar or series still applies to all batteries, exactly as today.
- Each battery gets its own charge/discharge variables, SOC trajectory and per-timestep direction binary. The power balance sums battery power over batteries, the same way multiple PV plants already sum. No mode, priority or ordering binaries were added: dispatch order emerges from per-battery efficiencies and costs.
- Symmetry breaking: a tiny fixed tilt (`BATTERY_TIEBREAK_EPS`, internal constant, not a config param) prefers lower-index batteries when costs tie exactly, so two identical batteries produce one deterministic plan instead of flapping between equivalent ones. Ties in both directions (charge and discharge) resolve to the lowest index. Any real cost or efficiency difference dominates it; a dedicated test proves a 0.1% round-trip efficiency edge beats the tilt.
- `set_battery_first_priority` carries over by gating on aggregate stored energy above the aggregate minimum.
- `soc_init`/`soc_final` accept per-battery lists at runtime (scalars broadcast), on both the naive-MPC and day-ahead paths.
- Publishing: `sensor.p_batt_forecast` stays the fleet total. With more than one battery, per-battery `sensor.p_batt_forecast_battery<k>` and `sensor.soc_batt_forecast_battery<k>` are added, and no fleet SOC is published since an aggregate SOC across packs of different sizes is not meaningful.
- The web config UI treats the battery parameters the way it already treats the deferrable-loads section: changing `number_of_batteries` grows or shrinks the dependent arrays automatically, and the fields do not get free add/remove buttons that could desync them.

With `number_of_batteries: 1` (the default) the model is mathematically unchanged: regression tests pin the objective value and the full charge/SOC schedule of a fixed scenario to the pre-change optimizer's output, and the published entity set is identical to today's.

## Behaviour notes reviewers should know about

- **Aggregate battery-first semantics.** With mixed fleets, the gate opens on fleet-level stored energy, so a small battery can still hold charge while the fleet as a whole counts as drained. Documented in `docs/config.md`. Per-battery gating would need one binary per battery per timestep, which is exactly the gate machinery the greenlight ruled out.
- **Flat list of length N.** A flat numeric list for `weight_battery_*` whose length happens to equal `number_of_batteries` is read as per-battery values, not as a shared time series of that length. That shape crashed before this PR whenever it did not match the horizon, so no working config changes meaning; the docs call it out and show the nested workaround.
- **`custom_batt_soc_forecast_id` at N>1** is ignored with a logged warning, since the per-battery SOC entities use fixed ids. The fleet power override keeps working at any N.
- **Tie-break determinism vs solver gap.** The epsilon makes the optimum unique, but at the default `lp_solver_mip_rel_gap` (0.01) the solver may return any plan within the gap. Strict run-to-run determinism needs a tight or zero gap, the same knob that already governs schedule variability generally.
- **Day-ahead runtime SOC handling** changed from an unconditional `float(...)` cast to a list-safe passthrough so runtime lists reach the optimizer with a clear length error instead of an opaque crash.
- **Negative prices with zero battery weights.** With more than one battery, default-zero `weight_battery_*` and a tariff that goes negative, the optimizer can profitably churn energy between batteries (one charges while another discharges) to consume negatively priced grid energy, a pattern a single battery cannot produce because of its own charge/discharge exclusivity binary. This is economically rational given a zero cycle cost, but it is hard on hardware, so the docs now recommend a small non-zero `weight_battery_charge`/`weight_battery_discharge` for multi-battery fleets on dynamic tariffs. Suppressing it structurally would mean cross-battery exclusivity constraints, which is the gate machinery this design deliberately avoids.
- **Battery identification** (`set_use_battery_identification`) supports a single battery only and now skips cleanly with a warning at N>1. Its estimator assumes one power meter and one SOC signal, so multi-battery support there is a separate piece of work.
- **Schema-contract test** gained exemptions for the 15 battery params whose declared type is now `array.*` while their defaults stay scalars by design (the broadcast form).
- **`static/configuration_list.html`** got one companion line for the UI array-sync wiring above.

## Out of scope (per the discussion)

Phase 2 (intermittently connected batteries, V2G, SOC re-anchoring) and phase 3 (battery-to-inverter association, #870) are untouched, as agreed.

## Tests

182 new tests across three new files plus targeted additions to the existing suites: broadcast and validation semantics, N=1 no-op pins (objective, schedule, published entities, config values), N=2 model shape and per-battery limits, tie-break determinism and epsilon dominance on scenarios with genuine allocation freedom, aggregate battery-first, per-battery deficit/surplus, the issue-936-style curtailment interaction at N=2, hybrid DC bus, per-battery ramp limits, model size scaling (N=2 adds exactly the replicated per-battery binaries and nothing else), cache behaviour (count change rebuilds, power-limit change stays a warm cache hit), runtime list handling end to end, publish shapes, and stale-frame replay. Full suite green locally.

Beyond the unit tests, I ran a local simulation sweep of 3,685 scenario combinations (battery counts 1 to 3, capacity mixes including a deliberately extreme 100 Wh + 50 kWh fleet, efficiency spreads, weight shapes, battery-first, deficit/surplus, ramp limits, hybrid and curtailment profiles, four price shapes, out-of-band SOC recovery starts) checking solver status and ten invariant classes on every solve: 3,685/3,685 passed.

## Credits

@scruysberghs proved out the constraint replication in his `feature/multiple-batteries` branch and raised the dispatch-prioritisation question this design answers. @ffrog8's findings on #1020 fold into phase 3 as agreed, tagging as promised. Thanks @JoshC1994 for the ping that restarted this and @purcell-lab for the earlier design input on #610.


src/emhass/optimization.py
```diff
@@ -1993,66 +2392,27 @@ def _add_battery_constraints(self, constraints, batt_stress_conf):
                 "maximum_power_from_grid",
                 self.num_timesteps,
             )
-            # 1% SoC tolerance so the gate opens cleanly once the battery has
-            # numerically reached its minimum, avoiding chatter at the floor.
-            soc_tolerance_energy = 0.01 * cap
+            aggregate_stored_energy = sum(current_stored_energy_list)
+            aggregate_min_energy = sum(soc_min_list[k] * cap_list[k] for k in range(self.n_batt))
+            aggregate_cap = sum(cap_list)
```
JoshC1994, line 2466: Minor: for a very lopsided fleet (e.g. 100 Wh + 50 kWh), the 1% aggregate tolerance below can exceed the small battery's usable SoC swing, so the drain-first gate may rarely engage for it. Already documented in `docs/config.md` , a log warning would just surface it at solve time too. Not blocking, feel free to skip.

```suggestion
            aggregate_cap = sum(cap_list)
            if self.n_batt > 1:
                usable_swings = [
                    (soc_max_list[k] - soc_min_list[k]) * cap_list[k] for k in range(self.n_batt)
                ]
                if max(usable_swings) > 10 * min(usable_swings):
                    self.logger.warning(
                        "Batteries are too different in size (%.0fx); the smaller "
                        "battery may not be discharged as expected, since the fleet may "
                        "be treated as already discharged.",
                        max(usable_swings) / min(usable_swings),
                    )
```

src/emhass/optimization.py
```diff
@@ -913,22 +985,182 @@ def _update_def_current_operating_timesteps_params(self, num_def_loads: int) ->
...
+            ),
+            "soc_max": self._batt_list(
+                self.plant_conf, "battery_maximum_state_of_charge", required=True
+            ),
+            "soc_target": self._batt_list(
+                self.plant_conf, "battery_target_state_of_charge", required=True
+            ),
+            "stress_cost": self._batt_list(self.plant_conf, "battery_stress_cost", default=0),
+            "soc_deficit_threshold": self._batt_list(
+                self.optim_conf, "battery_soc_deficit_threshold", default=0.2
```
JoshC1994: `default=0.2` here disagrees with `config_defaults.json's` line 136 `"battery_soc_deficit_threshold": 0.40,` and the new `BATT_ARRAY_PARAMS_OPTIM_CONF` default in `utils.py` line 3523     `"battery_soc_deficit_threshold": 0.4,`.
Pre-existing on master, currently dead code since build_params backfills this key first. But since this file is already being touched, worth aligning to 0.4.

```suggestion
                self.optim_conf, "battery_soc_deficit_threshold", default=0.4
```

JoshC1994 (commented):
I'm still fairly new to reviewing PRs like this, so take my comments with that in mind.
It looks like solid work :)

Left two small suggestions inline, both non-blocking:

-A log warning for when the batteries in a fleet are very different sizes,  just surfaces something that's already documented in docs/config.md.
-A one-line default value that doesn't match between two files (0.2 vs 0.4) pre-existing, not caused by this PR, just noticed it while reading through.

Happy to be corrected on either.

LesIT1:
Thanks @JoshC1994, good catches on both. Landed in 7af02d4.

The 0.4 alignment went in as suggested. You're right that it's dead code today since build_params backfills the key first, but no reason to leave the mismatch sitting there.

Took the warning too, with two small tweaks: guarded the ratio against a zero usable swing (soc_max == soc_min would have divided by zero), and reworded the message so it describes the mechanism the same way docs/config.md does (the gate tracks aggregate SoC, so the small battery may not get drained before import is allowed). Verified it fires for a 100 Wh + 50 kWh fleet and stays quiet for a balanced one.

Don't sell the review short by the way, both comments were on the money.

davidusb-geek:
@LesIT1, another PR merge has caused a conflict in your modifications to optimization.py

LesIT1:
Rebased on master, conflict resolved.

Two parts of the new master needed composing with the multi-battery structure rather than a plain re-apply:

- The #1028 export ceiling carried over unchanged. It only involves grid and PV terms, so it needed no battery-side changes.
- The softened terminal SoC constraint is applied per battery: each battery gets its own under/over slack pair, priced the same way as master. Kept per battery rather than aggregate so one battery overshooting can't cancel another undershooting for free. With number_of_batteries: 1 it reduces to exactly the master formulation.

Full suite green locally.

scruysberghs:
We ran this PR against a real two-battery household to help validation.

**Setup**: a shadow EMHASS container built from this branch, fed the exact runtime payload of our production instance (same prices, PV forecast and load, rendered at the same moment). The site has two plug-in batteries: 5.12 kWh / 2500 W and 5.0 kWh / 2400 W (the second is a hybrid unit with its own PV panels). `number_of_batteries: 2`, per-battery lists for capacity and power, `soc_init` passed as a per-battery list from the live unit SOCs. The load ML model was fitted inside the shadow with the same fit parameters as production, and the deferrable loads were kept in the payload, so both plans run on equivalent inputs (bottom panel shows both load forecasts; they are independently fitted models, hence the small day-2 divergence).

**Results over the 48h MPC horizon**:
- Per-battery outputs (`P_batt_k`, `SOC_opt_k`) behave correctly: independent SOC trajectories from per-battery `soc_init`, individual dispatch (the units even alternate covering the small overnight base load), and when one unit reaches its max SOC the optimizer moves charging to the other one.
- Zero per-battery power limit violations.
- The fleet total tracks our current production single-aggregate plan closely (dotted red), so no regression on the aggregate behaviour, while the per-unit limits and SOC states are now respected individually. Note the aggregate plan occasionally swings at 4200 W, which neither single unit could actually deliver; the per-battery plan never asks a unit for more than it can do, which is exactly the real-world problem this PR solves for us.

**One observation, maybe worth a doc note**: runtime scalars broadcast over all batteries as designed, and they override list values from the config. Our production payload still passed aggregate scalars (e.g. `battery_charge_power_max: 4200`), which silently became `[4200, 4200]` and masked the per-unit limits in a first run. With `number_of_batteries > 1`, runtime battery params should be passed as lists; a warning when a runtime scalar overrides a configured list could save users some head-scratching.

PR1032 comparison on real 2-battery household

Happy to rerun on this site as the PR evolves.

davidusb-geek:
> **Setup**: a shadow EMHASS container built from this branch, fed the exact runtime payload of our production instance (same prices, PV forecast and load, rendered at the same moment). The site has two plug-in batteries: 5.12 kWh / 2500 W and 5.0 kWh / 2400 W (the second is a hybrid unit with its own PV panels). `number_of_batteries: 2`, per-battery lists for capacity and power, `soc_init` passed as a per-battery list from the live unit SOCs. The load ML model was fitted inside the shadow with the same fit parameters as production, and the deferrable loads were kept in the payload, so both plans run on equivalent inputs (bottom panel shows both load forecasts; they are independently fitted models, hence the small day-2 divergence).

Looking at the SOC plot those results seems quite good and coherent.
Even though your ML load predict does not yield exactly the same prediction (mostly for the 2nd day).
I don't think this PR will need more evolution, do you confirm @LesIT1 ?

LesIT1:
Confirmed, nothing more needed from my side. The rebase composed cleanly with the new master (per-battery terminal SoC slacks, export ceiling untouched) and the full suite is green locally.

@scruysberghs thank you for running this on a real two-battery site, that's exactly the validation it needed. The alternating overnight dispatch and the charge handoff at max SOC are the emergent behaviours the design was meant to produce, so it's great to see them show up on real hardware rather than just in sims.

Your runtime scalar observation is a fair catch. The broadcast is by design, but a scalar silently masking a configured per-battery list is a footgun. Rather than grow this PR while it's ready to merge, I'll send a small follow-up adding a warning when a runtime scalar overrides a configured list with number_of_batteries > 1, plus a doc note in the runtime parameters section.

davidusb-geek (approved):

PR #1039, opened 2026-07-23
docs: add the missing heat topology guide

mohammadrezwankhan:
## Problem

The `heat_topology` descriptions in `param_definitions.json` and `openapi.json` point users to `docs/heat_topology.md`, but that page does not exist. This leaves the graph schema, source-type behavior, units, and hybrid-heating setup undiscoverable. The missing page was reported in Discussion #940.

## Changes

- add the referenced heat-topology documentation page
- document sources, storage, consumers, flows, actuator groups, and cost tracks
- include a compiler-verified heat-pump plus gas-boiler DHW example
- explain electric-balance defaults, runtime/static configuration, horizon arrays, validation, and troubleshooting
- link the page from the thermal toctree and the existing thermal-battery guide

## Validation

- documented example executed through `compile_heat_topology`
- `24 passed` for `TestCompileHeatTopology` plus `test_shared_thermal_tank_two_sources`
- Sphinx HTML build succeeded and rendered `heat_topology.html`
- `git diff --check` passed

No runtime behavior is changed.

Discussion: https://github.com/davidusb-geek/emhass/discussions/940

## Summary by Sourcery

Add comprehensive user documentation for the heat topology graph model and link it into the thermal docs navigation and thermal-battery guide.

Documentation:
- Introduce a new heat topology guide describing the graph-based hybrid heating model, configuration structure, and validation/troubleshooting guidance.
- Document sources, storage, consumers, flows, actuator groups, cost tracks, and a worked hybrid DHW example using the heat topology model.
- Link the new heat topology guide from the thermal section toctree and reference it from the thermal-battery documentation for multi-source storage setups.

Issue #1026, opened 2026-07-14
adjust_pv trains on curtailed PV production; ignore_pv_feedback_during_curtailment does not protect the training data


scruysberghs:
`ignore_pv_feedback_during_curtailment` only bypasses the alpha/beta mix of the live sensor value into forecast step 0 (`Forecast.get_mix_forecast`, forecast.py). It has no effect on the adjusted-PV model fit.

`adjust_pv_forecast` retrieves the raw history of `sensor_power_photovoltaics` (measured production) and `sensor_power_photovoltaics_forecast` for `historic_days_to_retrieve` days and fits actual-vs-forecast on all of it (`_retrieve_from_hass` in command_line.py, `adjust_pv_forecast_data_prep` in forecast.py). During curtailment the measured production is deliberately below the achievable PV power, so on systems that curtail regularly (negative injection prices, battery full) the regressor learns a downward bias: the model is taught that "on sunny midday hours, actual < forecast", exactly the hours where curtailment is planned. The adjusted forecast then under-predicts on non-curtailed days, and since the adjusted forecast feeds the optimization, the LP plans with less PV than available.

Suggested fix: exclude timesteps where curtailment was active from the training set. The optimizer already knows its own planned curtailment (`P_PV_curtailment` in `opt_res_latest` / the published `custom_pv_curtailment_id` entity when `compute_curtailment` is true); retrieving that entity's history alongside the PV sensors and dropping rows where curtailment > 0 (or adding the curtailed power back onto the measured production) would keep the training target clean.

Happy to contribute a PR if there is agreement on the approach (drop rows vs. add-back).

scruysberghs:
Follow-up on the two options at the end of the issue (drop rows vs. add-back): I would go for **dropping the curtailed timesteps** from the training set.

Add-back reconstructs a hypothetical measurement from the optimizer's own plan: it assumes the planned curtailment was executed fully and exactly, which is not guaranteed in practice (execution lag between plan and inverter, partial curtailment, export limiters acting independently of the plan). Training on a reconstructed target risks teaching the model a systematic error.

Dropping is conservative: training data for the affected hours is abundant over a multi-day retrieval window, and a somewhat smaller training set is a much smaller risk than a biased target. Concretely: drop every timestep where the curtailment entity (`custom_pv_curtailment_id` history, when `compute_curtailment` is enabled) was > 0, optionally extended by one timestep on either side to absorb execution lag.

PR #1043, opened 2026-07-27
feat: self-tune PV forecast bias via adaptive conformal inference (#841)

Whatsonyourmind:
## Summary

Phase-1-smarter for #841, per your go-ahead. Adds `emhass.pv_bias_calibration`, a **side-effect-free engine** that turns the `weather_forecast_pv_quantile_bias` knob (#961) from a hand-set number into a **self-calibrating** one, using adaptive conformal inference (ACI).

Instead of guessing a blend weight, the user picks an interpretable target — a one-sided **shortfall rate** `α` (how often realised PV may land below the forecast planned against) — and a one-scalar online recursion walks the bias to hold it:

```
planned_t   = bias_t * P10_t + (1 - bias_t) * P50_t
shortfall_t = 1 if realised_t < planned_t else 0
bias_{t+1}  = clip(bias_t + gamma * (shortfall_t - α), 0, 1)
```

The realised shortfall rate converges to `α` (when it's inside the range the P10–P50 blend can express) and keeps tracking it as conditions drift — that's the ACI guarantee. `compute_pv_bias_calibration(...)` returns a recommended `bias` plus diagnostics (achieved rate, feasible range, the static shortfall-vs-bias curve, the full trajectory). It **never touches the live forecast, the LP, or default behaviour** — it reports a value the user (or a later action) can apply.

## Scope — what this PR is and isn't

This is the **algorithm engine** — the part that's fully unit-testable in isolation. It intentionally does **not** wire itself into the Solcast retrieval path: the input it consumes (the per-period P10/P50 you *planned against* vs. realised PV) isn't logged yet, since the #961 blend discards the percentiles after blending. That data-logging + a `forecast-calibration`-style action to feed this engine is the natural **companion step**, and it sits in @LesIT1's Solcast-path territory — happy to coordinate on who takes it so we don't overlap. Until then the engine works on any logged series a caller supplies.

Design notes carried over from the thread:
- **v1 = single *shift* scalar.** The *shrink* degree of freedom on the band width (the Jul-19 over-wide-tail point) is a documented follow-up, not in this PR.
- **Pluggable score.** The default is the shortfall indicator; a `score_fn` hook lets a cost-regret / threshold-weighted-CRPS "merit-function" tie-breaker replace it without changing the recursion — this is where @lutorm's point (forecast shortfall ≠ merit outcome) gets addressed, and where the bias *plateau* @LesIT1 found gets a principled tie-break.
- **γ** is a parameter here; the parameter-free DtACI refinement is cited as a future extension rather than implemented, to keep this PR tight.

## Validation

- `tests/test_pv_bias_calibration.py` — 11 unit tests, all passing: convergence to a feasible target, the exact recursion arithmetic, static-curve monotonicity + feasibility orientation, infeasible-target saturation, the pluggable score hook, and input validation.
- `ruff check` and `ruff format --check` clean on both files.
- @LesIT1's 41-day Perth backtest on the issue is effectively this engine's acceptance test — the γ sweep, cold/warm start behaviour and the plateau finding all match the recursion implemented here.

## Docs & references

`docs/forecasts.md` gains a "Self-tuning the bias (adaptive conformal inference)" subsection under the existing P10-blend docs, plus a CHANGELOG entry. References added as requested:
- Gibbs & Candès (2021), *Adaptive Conformal Inference Under Distribution Shift*, arXiv:2106.00170.
- Gibbs & Candès (2024), *Conformal Inference for Online Prediction with Arbitrary Distribution Shifts*, arXiv:2208.08401 (DtACI).

Credit where due: the auto-tune idea is @lutorm's, the two-phase design @purcell-lab's, and @LesIT1 did the Phase-1 blend (#961), the calibration log and the backtest.

## Summary by Sourcery

Introduce a side-effect-free engine to self-calibrate the conservative PV forecast bias using adaptive conformal inference and document how to use it.

New Features:
- Add pv_bias_calibration module to compute a recommended weather_forecast_pv_quantile_bias from logged P10, P50, and realised PV history based on an adaptive conformal inference recursion.
- Expose diagnostics such as feasible shortfall range, static shortfall-vs-bias curve, and bias trajectory alongside the recommended bias value.

Enhancements:
- Document self-tuning of the PV forecast bias in forecasts.md, explaining the shortfall-rate target, recursion behaviour, and how to apply the recommendation engine.
- Record the new self-tuning PV bias capability and its adaptive conformal inference basis in the changelog.

Documentation:
- Add documentation section describing self-tuning PV forecast bias via adaptive conformal inference, including configuration guidance and references to the underlying methodology.

Tests:
- Add unit test suite for pv_bias_calibration covering convergence behaviour, recursion arithmetic, feasibility diagnostics, pluggable scoring, and input validation.


Whatsonyourmind:
Heads-up on the one red check: `build (ubuntu-latest)` fails at the **Check formatting** step, but it's a pre-existing repo-wide issue, not from this diff. CI runs `uvx ruff format --check --diff .` with an unpinned ruff, and the current ruff formats Python code blocks inside Markdown by default now — it flags `docs/cookbook/tariff_demand_charge.md`, which this PR doesn't touch. The same job is already red on `master` (`aef76a4b`) and on #1039, so it predates this branch. This PR's own files pass both `ruff check` and `ruff format --check`.

Happy to send a separate small PR to unblock CI for the whole repo — either pin ruff in `code-quality.yml`, or reformat that one Markdown code block (it's from my #1031 cookbook doc, so glad to). Whichever you'd prefer; this PR is independent of that fix.

JoshC1994:
@Whatsonyourmind I have been working on a fix for the same issue, just submitted my PR #1044.

Whatsonyourmind:
Thanks @JoshC1994 — good to have that fixed at the source; it unblocks the Code Quality Scan here too. Appreciate you picking it up.

LesIT1:
Ran an acceptance check against our live calibration log today (now 42 days, 2026-06-16 to 2026-07-27): fed it through `compute_pv_bias_calibration` at head a804178 and cross-checked against the backtest replay I posted on #841. The bias trajectory matches exactly on every grid point I tried (alpha 0.10 and 0.25, gamma 0.02 to 0.2, cold and warm starts), including the ordering detail that day t is planned with the bias produced by day t-1. Your test suite passes locally in our venv too. The static curve diagnostic is consistent with the thread: your finer 21-point grid resolves the plateau to 0.45-0.70 at rate 0.143 on our data (I had only posted the coarser 0.5-0.7), and the feasible range comes out (0.024, 0.667).

Two observations from that run, both small:

1. At gamma 0.05 the warm start (bias0 0.5) settles inside the plateau and recommends 0.60, while the cold start recommends 0.40, which sits at static rate 0.167, so it is still mid-climb after 42 days. Both report `converged: True`, the cold one because the 0.05 tolerance is about the same size as the 21-sample tail granularity (1/21 = 0.048) on a history this short. Since you already compute the static curve, one cheap improvement: also return the maximal flat segment of the static curve containing `recommended_bias`. For warm starts that surfaces the "this number isn't unique" warning, and for cold starts the segment comes out narrow, which is a useful "still settling" hint. The pluggable score stays the real fix for picking a point on the plateau.

2. On the score_fn hook: the recursion uses the custom score, but `feasible_shortfall_range`, `static_shortfall_curve` and `target_feasible` still use the default shortfall indicator, while `converged` and the `achieved_shortfall_rate` keys track the custom score. With a merit-function score plugged in, the diagnostics can contradict each other (an all-zero score gives achieved rate 0.0, target_feasible True, converged False). The docstring also says the score receives arrays, but the recursion passes scalars and an array-returning score is silently truncated to its first element. All fine to leave for the v2 cost-regret work, just worth a docstring note so nobody trips on it before then.

On the companion step: happy to take it. Once this lands I'll wire up the logging of the planned-against P10/P50 (the #961 blend currently discards the percentiles after blending) plus a forecast-calibration style action that feeds this engine, and I'll propose the exact shape to David in an issue before building anything.

scruysberghs:
do not forget to take curtailment into the equation when adjusting pv predictions (weather prediction vs real productions vs curtailed pv production)

Whatsonyourmind:
Good call — curtailment is exactly the kind of censoring that would poison this signal. When production is curtailed, realised PV is capped below what the panels could have made, so the forecast looks like it over-called and the tuner reads a spurious shortfall and pushes the bias up for the wrong reason. It's the same contamination EMHASS already guards against on the adjusted-PV side by excluding curtailed timesteps (#1026).

The engine here consumes `(P10, P50, realised)` and stays a pure recommender, so the clean place to handle it is the feed boundary: the history must be built from **uncensored** observations — curtailed steps excluded (one-step margin, as in #1026), and `realised` taken as available/pre-curtailment PV rather than metered export. I've just pushed that as an explicit precondition in the module docstring (`c0b0719c`), and it's a natural thing for the data-logging companion to enforce since it already has to know when `compute_curtailment` was active. Thanks for flagging it.

LesIT1:
Noted on the companion side. Since the logger has to sit next to `compute_curtailment` anyway, I'll state the uncensored-history precondition up front in the companion issue: curtailed steps excluded with the one-step margin per #1026, and realised PV taken from before the curtailment cap rather than metered export. David gets the full shape before anything is built.

Whatsonyourmind:
@scruysberghs implemented in `1103c607` rather than just documented, since you're right that this is the failure mode most likely to bite in practice.

`compute_pv_bias_calibration` now takes an optional `censored` mask marking steps whose realised PV was capped (curtailment, export limiting, inverter clipping), and drops them from the recursion. There's a `censored_margin` too — curtailment ramps in and out, so on raw timesteps the neighbours are often partly capped as well; it defaults to 0 because a "step" here is whatever the caller aggregates, often a whole day, and silently dropping two extra days per curtailed day would be worse than the disease.

One design note worth stating, because the tempting option is wrong. A censored step that *still clears the plan* is genuinely informative — true PV ≥ metered ≥ planned, so it cannot have been a shortfall — while a censored step below the plan is ambiguous. Keeping the informative half and dropping only the ambiguous half looks like free information, but it retains censored non-shortfalls while discarding censored shortfalls, which biases the achieved rate *downward* and would quietly under-protect. So the whole censored set goes.

Dropping isn't free either: curtailment is not missing-at-random, it concentrates on exactly the high-irradiance days you'd most like in the sample. The result now reports `n_censored_excluded` and `censored_fraction`, and warns above 20% that the recommendation describes the non-curtailed regime rather than the site as a whole — better to say so than to hand back a confident-looking number from an unrepresentative history.

Two of the five new tests are the ones I'd point a reviewer at: `test_masking_curtailed_steps_reproduces_the_clean_history` splices curtailed days into a clean history and asserts the masked run recovers the clean bias trajectory *exactly*, and `test_unmasked_curtailment_over_tightens_the_bias` runs the same data unflagged to pin the failure mode you described. Docs updated in `docs/forecasts.md`.

Detecting curtailment is still the caller's job — the recursion can't see it in the numbers alone — so when the logging companion piece lands it should carry the flag through rather than leave users to reconstruct it.

Whatsonyourmind:
Follow-up to my last comment — I went back through how EMHASS already handles this and realigned the API to it in `3298ad13`, since I'd invented vocabulary the codebase doesn't use.

The main change is naming. I had gone with the statistical term (`censored`), but the repo speaks `curtail*`, and the adjusted-PV filter you wrote in #1027 already fixes the shape of it: `curtailment_series`, `curtailed`, `n_curtailed`. So the parameters are now `curtailed` / `curtailment_margin`, and the result keys are `n_curtailed_excluded` / `curtailed_fraction`. "Censoring" survives only in the prose explaining *why* the steps are dropped, where it's the accurate word. I'd also been citing #1026 for the code — that's the issue; the merged implementation is #1027, and the docstring now says so.

Three more things came out of reading `forecast.py` properly:

**It now accepts the power series, not just a boolean mask.** The source of truth in EMHASS is `custom_pv_curtailment_id` (`sensor.p_pv_curtailment`) in watts, which you threshold at `> 0`. Expecting callers to pre-convert that to a bool was an unnecessary seam, so `curtailed=` takes either, thresholds the same way, and treats a non-finite sensor reading as "not curtailed" rather than propagating it.

**It logs the exclusion count at INFO on every filtered run**, mirroring the message shape at `forecast.py:1357`. Previously it only spoke up above the 20% warning threshold, which meant the two curtailment filters looked different in a user's log for no good reason.

**The margin default stays 0, deliberately, and I should have justified it out loud rather than in a param docstring.** Yours is hardcoded to 1 because a step there is one optimization timestep — 30 minutes by default — so a margin either side is cheap and absorbs execution lag. A step in this history is whatever the caller aggregates for one update, which for daily whole-day totals means `curtailment_margin=1` would discard two good days for every curtailed one. So it defaults to 0 and the docs say to set it to 1 when calibrating on raw timesteps, which reproduces your behaviour exactly.

I also added a warning I think matters more than the feature itself: **the flag must never be derived from the forecast residual.** The tempting shortcut — "realised came in far below forecast, so that must have been curtailment" — deletes precisely the shortfall observations the recursion counts, drives the bias to zero and removes all protection. It's the worst possible failure mode here and it's an easy mistake to make, so it's now called out in both the module notes and the docs.

Two smaller notes for the record: the margin widens by array position rather than timestamp, so on a history with gaps the neighbours dropped are adjacent rows rather than adjacent moments (the same latent behaviour as the `.shift()` upstream, now documented); and the not-missing-at-random direction I claimed reverses if curtailment is *scheduled from the forecast* rather than triggered by output, because then the dropped steps are selected on a high planned value — which is the same quantity the shortfall test compares against. The docstring says both directions now instead of asserting the convenient one.

18 tests pass, ruff clean. One thing I left out on purpose: since a curtailed step below the plan is genuinely unresolvable, the honest summary of the achieved rate is an interval rather than a point — scoring the ambiguous steps as 0 and as 1 gives its two ends. That's a real improvement but it grows the result contract, so I'd rather offer it than assume it. Happy to add it if you want it.

davidusb-geek (approved):

Issue #1045, opened 2026-07-27
set_zero_min destroys the signed battery power signal before battery identification sees it


LesIT1:
While building the per-battery identification PR for #1042 I found a data-path problem that affects battery identification (#963 / #1007) at any number of batteries, on default config.

The identification retrieval path goes through `retrieve_home_assistant_data`, which always calls `rh.prepare_data(..., set_zero_min=retrieve_hass_conf["set_zero_min"])`. With `set_zero_min: true`, the default, `prepare_data` clips the whole retrieved frame to >= 0 and then maps exact zeros to NaN:

```python
# retrieve_hass.py
if set_zero_min:
    self.df_final.clip(lower=0.0, inplace=True, axis=1)
    self.df_final.replace(to_replace=0.0, value=np.nan, inplace=True)
```

That clip is column-blind, so it also hits `sensor_power_battery`, which is documented as a signed sensor (charge one sign, discharge the other, direction auto-detected). Quick repro through `prepare_data` itself:

```
before: [1200.0, -900.0, -1500.0, 800.0, -2000.0, 0.0, 1500.0, -300.0]
after : [1200.0, nan,    nan,     800.0, nan,     nan, 1500.0, nan]
```

So on a default-config deployment the estimator never sees one whole direction of battery flow. Depending on the site's sign convention that is either every discharge or every charge sample. The estimator then finds no segments in that direction and reports insufficient data, keeping the configured values. It fails safe, no wrong numbers get published, but the feature silently never fits on exactly the config most people run.

The unit and orchestrator tests inject the DataFrame directly past `prepare_data`, which is why this never showed up in CI.

Suggested fix shape: exempt the battery identification sensors from the zero clip in `prepare_data` (a small opt-out list argument, only passed on the `battery_id` retrieval path), or skip the clip entirely for that retrieval path since none of its consumers want it. Happy to send either as a small PR if you have a preference.

Found while verifying the data path for the #1042 build, where the same applies per battery.

LesIT1:
Closing this one: @volschin filed #1041 the day before this, covering the same root cause, and the report comes with an in-container repro table plus a wider look at the knock-on effects. I missed it when I filed this out of the #1046 build, sorry for the split thread. The two fix shapes proposed here map onto the options listed there, so nothing is lost by consolidating. Follow #1041.

Issue #1042, opened 2026-07-27
[Feature request] Per-battery identification at number_of_batteries > 1 (per-battery sensor pairs)


LesIT1:
## The gap

#1032 deliberately left battery self-identification (#963, shipped in #1007) single-battery: at `number_of_batteries > 1` it skips with a logged warning, because the estimator assumes one power meter and one SOC signal. @JoshC1994 hit this on a real fleet (three batteries, each behind its own GoodWe GW5048D-ES, per-battery signed power and SOC sensors already in HA) and raised it on #610. His packs behave genuinely differently overnight, which is exactly the situation where per-pack constants are worth measuring rather than guessing: #1032 lets you configure capacity and efficiency per battery, but every one of those numbers is currently hand-entered.

There's also an upside a fleet-level signal can't reach at all: one AC meter can't attribute anything to individual packs, which is why identification skips at N>1 in the first place. Per-battery sensor pairs make each pack's capacity and round-trip efficiency genuinely identifiable, one estimator run per pack.

## Proposed shape

An increment over what already exists, following the #1032 index-matching convention:

- `sensor_power_battery` and `sensor_battery_state_of_charge` accept a list of exactly `number_of_batteries` entries, index-matched to the battery lists. A single string stays valid and means what it means today. At N>1 without both lists, identification keeps skipping, with the warning saying what's missing. Deliberately no scalar broadcast for these two: one meter can't be two batteries, so it's exact-length or skip.
- The estimator core is untouched. It's already a pure function of one (power, SOC) column pair, so it runs once per battery over that battery's retrieved pair. The orchestrator grows a per-battery loop around it: config reads, logging, publishing and persistence all become per-k.
- Persistence stays one JSON with a schema version bump, keyed per battery, and the freshness gate moves from file mtime to a per-battery `fitted_at`, so one pack passing a fit can't suppress another pack's retry. N=1 keeps the current shape and behaviour.
- Guardrails apply per battery under the one global trust tier (observe/suggest, auto_apply still deferred), so one pack can pass while another hasn't enough cycles yet.
- Publishing at suggest follows the #1032 naming: N=1 keeps `sensor.battery_identified_capacity` / `sensor.battery_identified_round_trip_efficiency`, N>1 publishes `..._battery<k>` variants.
- Config surface: this touches the schema and web UI array wiring the same way #1032 did for the numeric battery params. If you'd rather keep v1 config-file and runtime only, with no UI array editing for the sensor lists, that cut works too.
[...]

PR #1046, opened 2026-07-27
feat: per-battery identification at number_of_batteries > 1

LesIT1:
Implements #1042 as discussed: battery self-identification now runs once per battery at `number_of_batteries > 1`, so each pack gets its own capacity and round-trip efficiency estimate instead of the feature skipping entirely.

## What changed

- `sensor_power_battery` and `sensor_battery_state_of_charge` accept a list of exactly `number_of_batteries` entity ids, index-matched to the battery config lists. A single string keeps meaning what it means today. There is deliberately no scalar broadcast for these two: one meter cannot tell two batteries apart, so at N>1 it is exact-length lists or identification skips with a warning that names which key is wrong and what it expected. The resolver also rejects duplicate ids within a list (a duplicated fleet meter would otherwise be a de-facto broadcast and could publish the fleet total as one pack's capacity) and any overlap between the power and SOC lists.
- The estimator core is untouched. The orchestrator grows a per-battery loop: retrieval columns, config reads, logging, publishing and persistence are all per-battery, and each battery is independent. One pack can pass while another does not have enough cycles yet; a failed fit for one pack never blocks or discards another pack's result.
- Persistence stays one JSON with a schema version bump at N>1: entries keyed per battery, freshness tracked from each battery's own `fitted_at` instead of the file mtime, so one pack's fresh fit cannot suppress another pack's retry. Each entry also records the exact sensor pair it was fitted from, and a mismatch counts as stale, so editing or reordering the sensor lists re-fits instead of serving a cached result against the wrong pack. A failed fit still never touches previously persisted good data. At N=1 the file keeps today's flat shape byte-identically, and stale files of either shape found under the other battery count are treated as absent and re-fitted, never crashed on.
- Publishing at `suggest` follows the #1032 naming: N=1 keeps `sensor.battery_identified_capacity` / `sensor.battery_identified_round_trip_efficiency` untouched, N>1 publishes `..._battery<k>` variants with per-battery attributes.
- Guardrails apply per battery under the one global trust tier. Observe/suggest only, auto_apply is still deferred, same as before.

## Config surface

The two sensor keys are declared `array.string` in the parameter definitions and stay editable in the config file and via runtime parameters. I took the lighter cut we discussed in the issue for v1: no `number_of_batteries`-driven grow/shrink wiring in the web UI for these two keys (the generic array editing works, exact length is enforced at run time with a precise warning). Two side effects worth knowing about, both verified harmless: a single-battery user who saves from the web UI will have the scalar rewritten as a one-element list (same observable behaviour, pinned by a dedicated equivalence test; the known divergences are misconfigured sensor collisions, such as a battery sensor equal to the load sensor, where the list path dedupes the retrieval column and the bare string keeps master's duplicate), and the regenerated OpenAPI schema now declares these keys as arrays (the server never validates against it; only external schema-validating clients would notice).

## Behaviour notes, disclosed

- N=1 behaviour with a bare-string config is pinned byte-identical by regression tests (publish entities, file shape, log wording, cache handling). Because the keys are now `array.string`, the config UI saves a one-element list for single-battery users; that shape takes the list path with the same observable behaviour, and a dedicated test pins the equivalence side by side. Two misconfiguration corners at N=1 also change, both for the better: a malformed list value now gets a precise skip-warning on the re-fit path instead of a confusing retrieval failure, and identification-off runs are untouched either way.
- At N>1, history retrieval is one batched call per cycle covering only the batteries currently due a re-fit; already-fresh batteries are not part of the batch, so a problem with one of their sensors cannot affect a re-fit. If a sensor in the stale batch is unreachable, that cycle's re-fits defer to a later run. Documented in the feature doc.
- Sensor list validation warns and skips rather than raising, deliberately diverging from `check_batt_params`' hard ValueError for the numeric per-battery lists: identification is an advisory feature and a config mistake in it should never take down the optimization run it rides along with.
- Because the two keys live in the Local section (they are `retrieve_hass_conf`, not `plant_conf`), they get the generic array editor's add/remove buttons rather than the Battery section's `number_of_batteries`-locked length management. A desynced length cannot mis-fit anything: it is caught at run time with the exact-length warning. Wiring them into the count-driven UI machinery is the piece I would leave for a follow-up if you want it at all.
- `_identify_battery_impl` keeps the single-battery pipeline as a verbatim branch rather than sharing a helper with the per-battery loop. That is deliberate: it keeps the N=1 byte-identity auditable line by line against master. Happy to fold them together in a follow-up once the multi-battery path has some real-world mileage.
- While verifying the data path I found a pre-existing issue that affects identification at any battery count on default config: `set_zero_min` clips the signed battery power sensor before the estimator sees it, so one whole flow direction is lost and the fit reports insufficient data. That predates this PR (it is in the #1007-era retrieval path) and is out of scope here; filed separately as #1045 with a repro and a proposed fix shape.

## Tests

- Full suite: 912 passed, 36 xfailed, 0 failed.
- New/rewritten coverage: exact-list happy path at N=2 (per-battery fits, publishes and persistence), wrong-length/scalar/duplicate/non-string/overlapping sensor configs each skipping with the precise warning, partial-failure independence, per-battery freshness (only the stale pack re-fits), sensor-pair cache binding (swapping the lists re-fits and re-attributes), corrupt persistence entries contained to the affected battery, both file-shape transitions, and N=1 byte-identity pins.
- The old N>1 skip test asserted the single-battery-only warning; it now asserts the new behaviour and its replacement warning.

Docs updated: `docs/config.md` for both keys and a new "Multiple batteries" section in `docs/battery_identification.md`.

@JoshC1994 this is the build we discussed on #610 - your three-battery site with its per-battery sensor pairs is exactly the layout the lists are for, and a real-world run would be a great check on it (with one pack offline you would just run `number_of_batteries: 2` and two-entry lists for now). Note the per-battery SOC start question you raised is #585 territory and stays out of scope here, and be aware of the #1045 issue above if your power sensors are signed.

Happy to adjust any of it.

## Summary by Sourcery

Add per-battery orchestration, configuration, and persistence for battery self-identification, enabling multi-battery systems to obtain independent capacity and efficiency estimates while preserving single-battery behaviour.

New Features:
- Support per-battery battery identification for configurations with multiple batteries, with independent fits, caching, and Home Assistant sensor publishing per pack.

Enhancements:
- Extend Home Assistant retrieval and configuration handling so battery power and state-of-charge sensors can be provided as per-battery lists instead of only scalars.
- Introduce a schema-versioned on-disk container format for battery identification results to store separate entries per battery and handle corrupted or stale data robustly.
- Refine logging and sensor naming to distinguish per-battery identification results and to keep single-battery logs and entities backward compatible.

Documentation:
- Document multi-battery behaviour and configuration for battery identification and update configuration docs to describe list-based battery sensor keys.

Tests:
- Add extensive tests covering multi-battery configurations, sensor list validation edge cases, per-battery freshness and caching, corrupted persistence handling, and equivalence between scalar and single-element-list configs at N=1.


LesIT1:
Thanks Sourcery. Taking the three in order:

**Splitting `_identify_battery_impl`:** deliberate, and disclosed in the PR body. The N=1 branch is kept verbatim from master so the byte-identity claim is auditable line by line against the diff, and the regression pins lean on that property. Extracting shared orchestration now would blur exactly what the pins verify. I'd rather fold the two paths together in a follow-up once the multi-battery path has real-world mileage behind it.

**Mutating `retrieve_hass_conf` around the retrieval:** fair point, and it is the one piece of this diff I'd call subtle, which is why it's documented in `identify_battery`'s docstring. It is restore-in-finally and exception-safe, and within a request there is no concurrent observer (the server re-parses params into fresh dicts per request; `identify_battery` is awaited serially). Passing the subset explicitly means changing the signature of `retrieve_home_assistant_data` and `_retrieve_from_hass`, which are shared with the other set_types this PR deliberately doesn't touch. If @davidusb-geek prefers the explicit parameter I'm happy to thread it through, here or as a follow-up.

**Moving the persistence helpers into a `battery_identification_store` module:** the sibling feature keeps its persistence in `command_line.py` too (the adjusted-PV pickle path and the shared `is_model_outdated`), so a new module for four private helpers would be a departure from how the file is organised today. If the store grows in v2 (auto_apply is deferred but planned), that's the point where a module split earns its keep.

JoshC1994:
You mention the v1 lighter cut, yet I found I needed to read the PR/docs and it wasnt intuitive.

Suggest `sensor_power_battery` and `sensor_battery_state_of_charge` grow/shrink to match `number_of_batteries` when `set_use_battery_identification` is enabled?

Seems like a good fit since those fields are only needed for identification, and it would avoid manual YAML edits while keeping the normal battery UI unchanged and be in keeping with the N>1 batteries grow/shrink.

Setting a default into these so no breakages and would also make the transition smoother for existing single-battery configs.

FYI - I have set it up to run using a PR on my fork from your branch, i'll check back in once I get some data. I did have a boo boo today where app startup on reboot cased me to revert back to the master emhass copy causing some bad data.   

Good find and call regarding #1045, as it applys to my `sensor.inv001_battery_power` and `inv002`, where my signed values `negative = charge` & `positive = discharge`.

One other note is my `sensor.inv001_battery_power` is my DC bus power not the AC. Im yet to investigate if its exposed or how I can measure this DC-AC-DC efficiency loss. Something for tomorrow...

Also I resolved the pass `soc_init` via rest command. Would be good to somehow tie `sensor_battery_state_of_charge` to `soc_init` via the UI but probably out of scope for this PR. As you say need to make sure is parsed in the correct format.

LesIT1:
Fair feedback on the v1 cut, and that's the v2 wiring I had in mind, gating it on `set_use_battery_identification` is a good call. The reason I'd still keep it out of this PR is scope: the header-driven sync machinery is same-section only today, and these two fields live in a different section from `number_of_batteries`, so it needs new cross-section plumbing rather than a copy of the deferrable case. The contract this PR ships is settled and pinned by tests, so I'd rather send the UI wiring as a focused follow-up. If @davidusb-geek prefers it in here I'll take that direction.

On defaults, existing single-battery configs carry over as-is: a bare string still works, and a config saved through the UI becomes a one-element list that behaves identically (there's a test pinning that). For new slots at N>1 the deferrable pattern copies the definitions default into each new input, but that would put the same placeholder entity id against two packs and the resolver rejects duplicates at run time, so here I'd leave new slots empty instead. There's no safe entity id to invent for a second battery.

Really glad you're running the branch live, that's the test this needs. On the reboot mishap: the identification cache is safe across that swap in both directions. Master at N>1 skips identification before it ever touches the file, and this branch treats any unrecognised file shape as absent and just re-fits, so there's nothing to clean up there. The bad data will have been the plan output from running master against your multi-battery config.

The DC bus point is worth holding onto. The learned round-trip efficiency describes whatever boundary the sensor measures, so from DC bus power you'll get the round trip at that boundary, without the DC-AC leg (though still including anything between the bus and the cells). Worth saying the field is documented as AC-side, since the optimiser applies `battery_charge_efficiency` and `battery_discharge_efficiency` at the AC balance, so a DC-side feed will understate losses in the plan. Interested in what you find when you dig into it.

Agreed the `soc_init` tie-in is out of scope here, but it's a good idea, worth its own issue.

And yes, good to have your inverters as a second real-world data point for #1045.

davidusb-geek:
@LesIT1 and @JoshC1994 thanks for your contributions and discussions.
I think that the latest proposal by JoshC1994 can be deferred to another PR as this one is already quite big (>1000 lines change).
Lets keep the keep the N=1 (single battery) and N>1 (multi-battery) execution paths separate for now.
I support keeping the In-Place Mutation of `retrieve_hass_conf` as-is using the exception-safe `finally `block.

Great work on this!

davidusb-geek:
You tell me when do you think that this is good to go

LesIT1:
From my side this is good to go. The test suite is green and the only red check is the Code Quality Scan that's currently failing on master too, so that clears when it's fixed at the source. Josh is running the branch live on his GoodWe setup and anything that surfaces there can land as a follow-up along with the v2 UI wiring. Merge whenever suits.

davidusb-geek (approved):

PR #1054, opened 2026-08-02
fix: re-normalise per-load deferrable arrays after runtime overrides (#1040)

### LINKED ISSUE #1040: IndexError: list index out of range in _add_deferrable_load_constraints when using 3 deferrable loads (works fine with 2)
TITLE:
IndexError: list index out of range in _add_deferrable_load_constraints when using 3 deferrable loads (works fine with 2)

---

**Describe the bug**
Optimization (`dayahead-optim`, also reproduced on `naive-mpc-optim`) crashes reliably with an `IndexError` as soon as `number_of_deferrable_loads` is raised from 2 to 3, even though every per-load config array is correctly sized to match (verified extensively, see Additional context). The same config works perfectly with `number_of_deferrable_loads: 2`.

**To Reproduce**
1. Start from a working config with 2 deferrable loads (a single-block load + a flexible load), `dayahead-optim` runs fine, returns `Optimal`.
2. Change `number_of_deferrable_loads` from `2` to `3`, and extend every per-load array to 3 elements (see config diff in Additional context).
3. Restart the add-on (config.json is only read on process start, confirmed via a fresh "Full initialization completed" log line before testing).
4. Call `POST /action/dayahead-optim`.
5. Request fails with a 500 and the traceback in Additional context.

**Expected behavior**
Optimization runs and returns a plan including the 3rd deferrable load, same as it does for loads 0 and 1.

**Screenshots**
Not applicable (server-side crash, no UI involved).

- Home Assistant OS

**Your hardware**
- OS: Home Assistant OS 18.1 (HA Core 2026.7.4)
- Architecture: amd64 (x86_64, running under KVM virtualization)

**EMHASS installation type**
- Add-on (davidusb-geek/emhass-add-on), EMHASS version 0.17.9

**Additional context**

Config diff (relevant deferrable-load keys only):
```diff
- "number_of_deferrable_loads": 2,
+ "number_of_deferrable_loads": 3,
- "nominal_power_of_deferrable_loads": [1800, 6000],
+ "nominal_power_of_deferrable_loads": [1800, 6000, 2000],
- "minimum_power_of_deferrable_loads": [0, 0],
+ "minimum_power_of_deferrable_loads": [0, 0, 0],
- "cost_forecast_per_deferrable_load": ["null", "null"],
+ "cost_forecast_per_deferrable_load": ["null", "null", "null"],
- "is_electric_load": [true, true],
+ "is_electric_load": [true, true, true],
- "operating_hours_of_each_deferrable_load": [6, 0],
+ "operating_hours_of_each_deferrable_load": [6, 0, 0],
- "start_timesteps_of_each_deferrable_load": [0, 0],
+ "start_timesteps_of_each_deferrable_load": [0, 0, 0],
- "end_timesteps_of_each_deferrable_load": [0, 0],
+ "end_timesteps_of_each_deferrable_load": [0, 0, 0],
- "treat_deferrable_load_as_semi_cont": [true, true],
+ "treat_deferrable_load_as_semi_cont": [true, true, true],
- "set_deferrable_load_single_constant": [true, false],
+ "set_deferrable_load_single_constant": [true, false, false],
- "set_deferrable_startup_penalty": [0, 0],
+ "set_deferrable_startup_penalty": [0, 0, 0],
- "deferrable_load_max_cost": [0, 0],
+ "deferrable_load_max_cost": [0, 0, 0.3],
- "set_deferrable_max_startups": [0, 0],
+ "set_deferrable_max_startups": [0, 0, 0],
- "def_minimum_on_time": [0, 0],
+ "def_minimum_on_time": [0, 0, 4],
- "def_minimum_off_time": [0, 0],
+ "def_minimum_off_time": [0, 0, 4],
```
`deferrable_load_groups` is `[]` in both cases. `set_use_battery_identification: true` is also enabled in my config.

Traceback:
```
[ERROR] Exception on request POST /action/dayahead-optim
  File ".../quart/app.py", line 1464, in handle_request
    return await self.full_dispatch_request(request_context)
  File ".../quart/app.py", line 1502, in full_dispatch_request
    result = await self.handle_user_exception(error)
  File ".../quart/app.py", line 1059, in handle_user_exception
    raise error
  File ".../quart/app.py", line 1500, in full_dispatch_request
    result = await self.dispatch_request(request_context)
  File ".../quart/app.py", line 1597, in dispatch_request
    return await self.ensure_async(handler)(**request_.view_args)
  File "/app/src/emhass/web_server.py", line 814, in action_call
    msg, status = await _handle_action_dispatch(
  File "/app/src/emhass/web_server.py", line 564, in _handle_action_dispatch
    opt_res = await optim_actions[action_name](input_data_dict, logger)
  File "/app/src/emhass/command_line.py", line 1917, in dayahead_forecast_optim
    opt_res_dayahead = input_data_dict["opt"].perform_dayahead_forecast_optim(
  File "/app/src/emhass/optimization.py", line 4975, in perform_dayahead_forecast_optim
    self.opt_res = self.perform_optimization(
  File "/app/src/emhass/optimization.py", line 4595, in perform_optimization
    self._add_deferrable_load_constraints(
  File "/app/src/emhass/optimization.py", line 3025, in _add_deferrable_load_constraints
    has_max_cost = max_cost[k] > 0
                   ~~~~~~~~^^^
IndexError: list index out of range
```

With `logging_level: DEBUG`, the log confirms the loop reaches `k=2` before crashing:
```
DEBUG in optimization: Processing deferrable load 0
DEBUG in optimization: Processing deferrable load 1
DEBUG in optimization: Processing deferrable load 2
<crash happens right after, at the max_cost[k] > 0 line for k=2>
```

What I ruled out before filing: since `k=2` reaching the loop body already proves `optim_conf["number_of_deferrable_loads"] == 3` at runtime, `max_cost` (from `optim_conf.get("deferrable_load_max_cost", ...)`, `optimization.py` ~line 3000) must somehow be shorter than 3 *at that exact point*, despite the config saying otherwise. I checked:
1. Config file on disk, re-read immediately before and after the crash: `deferrable_load_max_cost` is `[0, 0, 0.3]` (3 elements), `number_of_deferrable_loads` is `3`. Confirmed via `python3 -c "json.load(...)"`, not just visual inspection.
2. `params.pkl` (the pickled parsed-params dict EMHASS writes for the request): `optim_conf["deferrable_load_max_cost"]` is also `[0, 0, 0.3]`, `optim_conf["number_of_deferrable_loads"]` is `3`.
3. Read `utils.py`'s `build_config()` (uses plain `dict.update()` — no element-wise merging that could truncate an array) and `check_def_loads()` (only pads short lists to match `number_of_deferrable_loads`, never shortens) directly from the `v0.17.9` tag — neither of these code paths should be able to produce a 2-element `deferrable_load_max_cost` given a 3-element input and `number_of_deferrable_loads: 3`.

I was not able to find the actual root cause in the time I had — reverted to `number_of_deferrable_loads: 2` to get a working system back, and I'm filing this rather than continuing to guess blindly. Happy to provide the full `config.json`/`params.pkl` privately if useful, or run further diagnostics on request.

Related issue: I searched first and found #929 (closed via #953), which is a false-positive-warning bug with a very similar config (`number_of_deferrable_loads: 3`, same add-on install). That report's optimization completed fine ("Optimal") — only the warning message was wrong, not the actual run. That makes me think this new `IndexError` is a separate, still-open problem rather than "3 loads just doesn't work" — something more specific in my config (maybe `def_minimum_on_time`/`def_minimum_off_time` being non-zero, `set_use_battery_identification: true`, or the `set_deferrable_load_single_constant`/`treat_deferrable_load_as_semi_cont` combination) may be the actual trigger, but I haven't isolated which one yet.

LesIT1:
## The bug

A runtime payload carrying a per-load array with fewer entries than
`number_of_deferrable_loads` crashes the optimization with
`IndexError: list index out of range`. @kristianbruun-ctrl hit this with a
stale 2-element `deferrable_load_max_cost` against 3 loads, and showed the
crash just moves to whichever short array gets read first (his second repro
used a short `set_deferrable_load_single_constant`). The earliest unguarded
read is in `Optimization.__init__` (`_initialize_decision_variables`,
optimization.py:1577), so any of the nine per-load arrays being short takes
the whole run down.

Root cause: the association loop in `treat_runtimeparams` copies raw runtime
values into `optim_conf` with no length check. `build_params` normalises all
of these arrays via `check_def_loads`, but that runs before the runtime
override, and nothing re-normalises after it. The per-battery arrays had the
same failure mode and got a post-loop re-normalisation in #610/#1032; this PR
extends the same pattern to the deferrable-load arrays.

## The fix

- New module-level table `DEF_LOAD_ARRAY_PARAMS` holding the nine per-load
  arrays and their defaults (values moved verbatim from the existing
  `check_def_loads` call sites in `build_params`; those nine call sites now
  loop over the table, no semantic change).
- A re-normalisation block at the end of `treat_runtimeparams` re-runs
  `check_def_loads` over the table against the final
  `number_of_deferrable_loads`. It sits last deliberately: both the runtime
  `def_load_config` handling and the `heat_topology` compile step can change
  the load count after the association loop, so normalising any earlier can
  be undone.
- A warning fires when a runtime-provided array (modern or legacy name) was
  actually changed by the normalisation, so a stale payload is visible in the
  log instead of silent.
- The block pads a copy, never the caller's list, and a castable
  `number_of_deferrable_loads` string from the payload is written back as an
  int.

Per-load arrays not in the table (`is_electric_load`,
`minimum_power_of_deferrable_loads`, `cost_forecast_per_deferrable_load`,
`def_load_config`) were each checked: every read path is length-guarded or
padded downstream, so they cannot produce this crash and were left alone.

## Behaviour changes

- A short runtime per-load array now pads with the param default and warns,
  instead of crashing. Same semantics `check_def_loads` has always applied to
  config-sourced arrays (#929).
- A `None` element in a runtime array is now replaced with the param default,
  again matching what `build_params` already does to config arrays. Four
  `none_element` cases in `test_config_value_resilience.py` that documented
  the old crash as expected-fail now pass and moved to that file's exclude
  list.
- A runtime scalar for one of these params (e.g.
  `"set_deferrable_load_single_constant": true`) is now broadcast to all
  loads, the same way the per-battery params treat a scalar. Previously a
  scalar was broadcast to the old array length, which crashed when the load
  count grew, and padding it with defaults instead would have given the value
  to some loads but not others.
- The block also runs when no runtime params are passed at all. That is
  required because `heat_topology` can resize the arrays without any runtime
  payload; for a config that came through `build_params` it is a no-op.

## Tests

New file `tests/test_deferrable_runtime_renormalise.py` covering the
reporter's two repro shapes, correct-size passthrough (byte-identical, no
warning), legacy param names, the `def_load_config` count reset, the
`heat_topology` interaction, `runtimeparams=None`, oversize arrays (not
truncated), `None` healing, string counts, and no mutation of the caller's
dict. Full suite green locally.

Fixes #1040

## Summary by Sourcery

Re-normalise deferrable per-load array parameters after runtime overrides to prevent index errors and align their sizing with the final number_of_deferrable_loads.

Bug Fixes:
- Prevent crashes when runtime-provided deferrable-load arrays are shorter than number_of_deferrable_loads by padding them instead of leaving them undersized.

Enhancements:
- Introduce shared DEF_LOAD_ARRAY_PARAMS and legacy-name mapping to centralise per-load deferrable array defaults and reuse them across config and runtime normalisation.
- Emit warnings when runtime-provided deferrable-load arrays are padded or None elements are healed so stale or malformed runtime payloads are visible in logs.

Documentation:
- Document that per-load deferrable array parameters are padded to number_of_deferrable_loads and that runtime short arrays produce warnings in the passing_data guide.

Tests:
- Add a comprehensive test suite for deferrable-load runtime re-normalisation, covering short and correctly sized arrays, legacy parameter names, heat_topology interactions, scalar broadcasts, None healing, oversize arrays, string load counts, and aliasing/mutation behaviour.
- Update resilience tests to mark previously expected-failing None-element deferrable-load cases as now passing due to the new re-normalisation behaviour.

### DIFF
--- a/docs/passing_data.md
+++ b/docs/passing_data.md
@@ -95,6 +95,8 @@ Here is the list of the other additional dictionary keys that can be passed at r
 
 - `set_deferrable_load_single_constant` to define if we should set each deferrable load as a constant fixed value variable with just one startup for each optimization task.
 
+- Note: for the per-load array parameters above (and also `deferrable_load_max_cost`), a value shorter than `number_of_deferrable_loads` is padded with that parameter's default, whether it came from your configuration or from this runtime payload. A runtime-supplied short array also logs a warning naming the parameter and how many entries were added; a config-sourced one stays silent. Pass a full-length array to avoid the warning.
+
 - `solcast_api_key` for the SolCast API key if you want to use this service for PV power production forecast.
 
 - `solcast_rooftop_id` for the ID of your rooftop for the SolCast service implementation.
--- a/src/emhass/utils.py
+++ b/src/emhass/utils.py
@@ -2359,6 +2359,79 @@ async def treat_runtimeparams(
             heat_topology,
         )
 
+    # Re-normalise per-load deferrable array params against the FINAL
+    # number_of_deferrable_loads (#1040). This has to run last: the
+    # association loop above, the def_load_config handling, and the
+    # heat_topology compile step can each change number_of_deferrable_loads
+    # or one of these arrays, and heat_topology runs even when runtimeparams
+    # is None. Mirrors the per-battery re-normalisation above (#610), but
+    # pads short arrays rather than truncating or erroring, matching
+    # check_def_loads's existing pad-to-fit semantics.
+    raw_num_def_loads = optim_conf.get("number_of_deferrable_loads", None)
+    if raw_num_def_loads is not None:
+        try:
+            final_num_def_loads = int(raw_num_def_loads)
+        except (TypeError, ValueError):
+            logger.warning(
+                "number_of_deferrable_loads is not a valid integer "
+                f"({raw_num_def_loads!r}); skipping deferrable load array "
+                "re-normalisation"
+            )
+            final_num_def_loads = None
+        if final_num_def_loads is not None:
+            # Write the cast int back: downstream readers of
+            # optim_conf["number_of_deferrable_loads"] need an int, not
+            # whatever raw value (e.g. a numeric string) the association
+            # loop copied in.
+            optim_conf["number_of_deferrable_loads"] = final_num_def_loads
+            runtime_source = runtimeparams or {}
+            for def_array_name, def_array_default in DEF_LOAD_ARRAY_PARAMS.items():
+                legacy_name = DEF_LOAD_ARRAY_LEGACY_NAMES.get(def_array_name)
+                runtime_value = runtime_source.get(def_array_name)
+                if runtime_value is None and legacy_name is not None:
+                    runtime_value = runtime_source.get(legacy_name)
+                # A JSON null is treated as "not provided", matching the
+                # association loop's own null-skipping - a padded config
+                # array should never be reported as a runtime override.
+                was_provided = runtime_value is not None
+
+                if was_provided and not isinstance(runtime_value, list):
+                    # A runtime scalar means every load, not "pad with the
+                    # default" - mirrors check_batt_params's silent
+                    # broadcast, and overrides any earlier partial handling
+                    # (e.g. set_deferrable_load_single_constant's own scalar
+                    # broadcast above, which doesn't know the final count).
+                    optim_conf[def_array_name] = [runtime_value] * final_num_def_loads
+                    continue
+
+                current_value = optim_conf.get(def_array_name)
+                if isinstance(current_value, list):
+                    # Copy before padding: check_def_loads mutates in place,
+                    # and this list may be the same object as
+                    # runtimeparams[name] or a passed_data alias set earlier
+                    # in this function - only optim_conf's own value should
+                    # change.
+                    before_value = list(current_value)
+                    optim_conf[def_array_name] = list(current_value)
+                else:
+                    before_value = current_value
+                optim_conf[def_array_name] = check_def_loads(
+                    final_num_def_loads,
+                    optim_conf,
+                    def_array_default,
+                    def_array_name,
+                    logger,
+                )
+                _warn_if_runtime_def_array_too_short(
+                    was_provided,
+                    def_array_name,
+                    before_value,
+                    optim_conf[def_array_name],
+                    final_num_def_loads,
+                    logger,
+                )
+            params["optim_conf"] = optim_conf
+
     # Serialize the final params
     params = orjson.dumps(params, default=str).decode()
     return params, retrieve_hass_conf, optim_conf, plant_conf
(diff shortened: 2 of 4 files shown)


LesIT1:
Thanks Sourcery, taking these in turn.

**Extracting the re-normalisation block into a helper.** Declining for this PR on consistency grounds: the block deliberately mirrors the per-battery re-normalisation pass that already lives inline in `treat_runtimeparams` (the #610 block above it, `check_batt_params` plus `_warn_if_runtime_scalar_masks_batt_list`). Extracting only the deferrable half would leave the two sibling passes organised differently for no behaviour change. If the inline weight bothers anyone I'd rather do a follow-up that extracts both passes symmetrically, happy to open that if @davidusb-geek wants it.

**Deriving the test's `DEF_ARRAY_NAMES` from `DEF_LOAD_ARRAY_PARAMS`.** Agreed in principle, and deliberately not done yet for the reason your comment hints at: the table doesn't exist on the base branch, and the tests are written to fail on base at their behavioural assertions rather than at an ImportError, so the red-on-base evidence stays meaningful (the test module docstring documents this). Once this merges that concern disappears and deriving the list is a two-line change, I'll fold it into the next PR that touches these tests.

**Tightening the classification in `_warn_if_runtime_def_array_too_short`.** Fair observation, but the two branches are exhaustive over what `check_def_loads` can currently do: it only ever pads a short list or replaces None elements, it never truncates or rewrites non-None values, so a same-length change is a None heal by elimination. Both branch messages are pinned by regression tests, and the docstring spells out the assumption so anyone expanding `check_def_loads` later has the pointer to revisit the wording alongside it. Keeping it as-is rather than adding a speculative element-wise check now.

davidusb-geek (approved):

Issue #1052, opened 2026-08-02
Failed battery identification fits are not persisted, so the model-age cadence gate never engages and every run re-queries the recorder


volschin:
Split out from #1041 (operational note) at the request of @LesIT1. Independent of the `set_zero_min` bug in #1041 — this makes *any* persistently failing fit expensive, whatever the reason for the failure.

### Symptom

When a battery identification fit fails, nothing is written. `is_model_outdated()` then sees a missing model file on the next call and reports "outdated", so the `battery_identification_model_max_age` cadence gate never engages. Every subsequent optimisation run repeats the full `historic_days_to_retrieve`-day, three-sensor recorder query.

With a 30-minute `naive-mpc-optim` loop and `historic_days_to_retrieve = 7`, runtime went from **3.9 s to 19.3 s** per run here, and the recorder gets a 7-day three-sensor query 48 times a day that is guaranteed to produce the same failure each time.

This is fail-safe in the sense that configured battery values are kept, but it turns a soft failure into a permanent load on the recorder, and it is invisible unless you look at run timings.

### Suggested fix

Persist failed attempts too, purely for the mtime and the status — a small record containing the failure `status` (`insufficient_data`, guardrail rejection, …) and the timestamp, without any fitted parameters. `is_model_outdated()` then sees a recent file and the cadence gate works as designed, so a permanently failing setup retries at `battery_identification_model_max_age` instead of on every run.

Storing the status alongside is useful in its own right: it makes "why is identification not doing anything" answerable from the persisted state rather than from log archaeology, and a status transition (e.g. `insufficient_data` → guardrail rejection) tells the user their config change had an effect even though the outcome is still a rejection.

### Environment

EMHASS 0.17.9, `ghcr.io/davidusb-geek/emhass:latest`, Home Assistant 2026.7, `optimization_time_step = 30`, `historic_days_to_retrieve = 7`, `battery_identification_trust_tier = observe`.

JoshC1994:
Can confirm this bug with real world tests with some of my batteries currently being rejected by the guardrails, every run Battery 1 with `status=rejected_sanity_check`, Battery 2 with `status=insufficient_data`. 
None of these rejected fits are being persisted, the model-age cadence gate never engages, and I can see the exact symptom described here firing on every single run:

```
Battery 1 identification model has no recorded fit, will train new model
Battery 2 identification model has no recorded fit, will train new model
Battery identification: retrieving history for a fresh fit
```

[...]

Issue #1051, opened 2026-08-02
Battery identification over-counts throughput: trapezoid integral bridges recorder gaps from change-only sensors


volschin:
Split out from #1041 (secondary observation) at the request of @LesIT1. #1041 stays scoped to the `set_zero_min` clip; this is the estimator weakness that shows up once the clip is worked around.

### Symptom

With `set_zero_min = False` (so discharge samples survive), `BatteryIdentification.identify()` does run but is rejected by its own guardrails:

```
Capacity 8.83 kWh outside [1.54, 7.68] kWh sanity band
RTE 0.38
TLS vs Theil-Sen capacity diverge by 34.1%
```

The battery is a Marstek Venus E, 5.12 kWh nameplate. One charge segment was measured as 87.5 SoC points / 19.4 kWh throughput — roughly 4× the physical value. Independently measured DC capacity for this unit is 5.37 kWh (geometric mean of a full charge/discharge cycle), so the estimator is out by a factor, not by a margin.

### Cause

`_process_history_dataframe()` resamples per day with `resample(freq).mean()`. A battery power sensor recorded **only on change** (the common case for MQTT/local-API integrations) leaves empty bins, which come out as `NaN`. `_segment()` then does a `dropna()` at the top, so those rows disappear entirely rather than being marked as gaps.

The trapezoid integral in `_build_segment()` afterwards sees two surviving samples that are hours apart and integrates the last known power across the whole gap. A battery that sat idle at 100 % SoC for 5.5 hours after a 2 kW charge contributes ~11 kWh of phantom throughput. That inflates the numerator of the capacity estimate while the SoC delta (the denominator) is unaffected, which is exactly the signature seen: capacity high, RTE low, TLS and Theil-Sen disagreeing because the two estimators weight the corrupted long segments differently.

### Suggested fix

Either would work, and they compose:

* Forward-fill the power column over empty bins before segmentation, so a change-only recorder is reconstructed into the step function it actually represents. This is the physically correct interpretation for a sensor whose contract is "value holds until next report".
* Bound the per-step `dt` used in the trapezoid integral (e.g. drop or cap any step longer than a few multiples of `optimization_time_step`), so a gap can never silently contribute throughput regardless of how the frame was built.

The second is the cheaper guardrail and protects against gaps from recorder purges and restarts too, not just change-only sensors. The first is what makes the estimate right rather than merely not-wrong.

### Environment

[...]

PR #1056, opened 2026-08-02
fix: exempt battery identification sensors from the set_zero_min clip

### LINKED ISSUE #1041: Battery self-identification never sees discharge segments: prepare_data(set_zero_min=True) clips sensor_power_battery
Battery self-identification (#963) can never produce a discharge segment when `set_zero_min` is enabled, which is the default. Every negative sample of `sensor_power_battery` is turned into `NaN` before the identifier sees it, so `identify()` always bails out with `insufficient_data`.

### Root cause

`retrieve_home_assistant_data()` runs `prepare_data()` on the retrieved frame for **every** `set_type`, including `"battery_id"` (`command_line.py:593`):

```python
    rh.prepare_data(
        retrieve_hass_conf["sensor_power_load_no_var_loads"],
        load_negative=retrieve_hass_conf["load_negative"],
        set_zero_min=retrieve_hass_conf["set_zero_min"],
        var_replace_zero=retrieve_hass_conf["sensor_replace_zero"],
        var_interp=retrieve_hass_conf["sensor_linear_interp"],
    )
```

and `prepare_data()` clips the whole DataFrame, not just the load column (`retrieve_hass.py:1269`):

```python
        # Apply Zero Saturation (Min value clipping)
        if set_zero_min:
            self.df_final.clip(lower=0.0, inplace=True, axis=1)
            self.df_final.replace(to_replace=0.0, value=np.nan, inplace=True)
```

So the `sensor_power_battery` column gets `clip(lower=0)` → all discharge samples become `0` → `replace(0, NaN)` → they are dropped by the `dropna()` at the top of `BatteryIdentification._segment()`. Only charge runs survive, `n_dis` is always `0`, and `identify()` returns `insufficient_data`.

This contradicts the comment at the call site that builds the var list (`command_line.py:551-554`):

```python
    if set_type == "battery_id":
        # Battery identification needs signed battery power and measured SoC.
        # The load sensor stays at var_list[0] so prepare_data's load handling
        # is unchanged; the battery columns pass through untouched.
```

The battery columns do not pass through untouched.

### Reproduction

EMHASS 0.17.9, `ghcr.io/davidusb-geek/emhass:latest`, Home Assistant 2026.7, `optimization_time_step = 30`, `set_use_battery = true`, `set_use_battery_identification = true`, `battery_identification_trust_tier = observe`, `historic_days_to_retrieve = 7`, a signed AC battery power sensor (positive = charge, negative = discharge) and a measured SoC sensor. The battery did several full 11 % → 100 % → 11 % cycles in the window.

Every `naive-mpc-optim` run logs:

```
INFO in command_line: Battery identification: retrieving history for a fresh fit
INFO in command_line: Battery identification: Not enough deep segments to fit (charge=3, discharge=0, require >= 3 each). Keeping configured battery values.
WARNING in command_line: Battery identification did not pass guardrails (status=insufficient_data); keeping configured battery values.
```

Running the identical retrieval in-container and only toggling `set_zero_min` on the `prepare_data()` call:

```python
rh.get_data(utils.get_days_list(rhc["historic_days_to_retrieve"]), var_list, ...)
neg_before = int((rh.df_final[power_col] < 0).sum())
rh.prepare_data(var_list[0], load_negative=..., set_zero_min=szm, ...)
neg_after = int((rh.df_final[power_col] < 0).sum())
segs = BatteryIdentification(log)._segment(rh.df_final, power_col, soc_col)
```

| `set_zero_min` | negative power samples before → after | segments |
|---|---|---|
| `True` (default) | 110 → **0** | charge 3, discharge **0** |
| `False` | 110 → 110 | charge 7, discharge 5 |

So the feature is unreachable on a default configuration.

### Why `set_zero_min = False` is not a workaround

`set_zero_min` exists to sanitise the load/PV columns, and it does real work there — my household load sensor legitimately reports occasional negative values (−31 W, −100 W). Turning it off to make battery identification work would feed those straight into the optimisation.

### Suggested fix

Keep `prepare_data()` as-is for the load/PV columns but exempt the two battery columns when `set_type == "battery_id"`. Either

* skip `prepare_data()` entirely for `"battery_id"` (the identifier only needs the raw signed power and SoC, and does its own deadband/hysteresis filtering), or
* snapshot `df[[sensor_power_battery, sensor_battery_state_of_charge]]` before the call and restore it afterwards, or
* add an explicit `protected_columns` argument to `prepare_data()` that is excluded from the clip/replace step.

The same consideration applies to the SoC column: a measured SoC of exactly `0 %` would be turned into `NaN` by the `replace(0.0, NaN)` line.

### Secondary observation (separate from the bug)

With `set_zero_min = False` the fit does run but is rejected by the guardrails — `Capacity 8.83 kWh outside [1.54, 7.68] kWh sanity band`, `RTE 0.38`, `TLS vs Theil-Sen capacity diverge by 34.1%`. The cause looks like the per-day `resample(freq).mean()` in `_process_history_dataframe()`: a battery power sensor recorded only on change leaves empty bins as `NaN`, `_segment()` drops those rows, and the trapezoid integral in `_build_segment()` then bridges hours of missing time at the last known power. One of my charge segments came out as 87.5 SoC points / 19.4 kWh throughput — roughly four times the physical value. Forward-filling the power column over empty bins (or bounding the per-step `dt` used in the integral) would make the estimator robust to change-only recorders. Happy to open a separate issue for this if you prefer.

### Operational note

When the fit fails, nothing is persisted, so `is_model_outdated()` sees a missing file on the next call and the cadence gate of `battery_identification_model_max_age` never engages. With a 30-minute optimisation loop this means a full `historic_days_to_retrieve`-day, three-sensor recorder query on *every* run: `naive-mpc-optim` went from 3.9 s to 19.3 s here. It might be worth persisting failed attempts too (with their status) purely for the mtime, so a permanently failing setup does not hammer the recorder.

LesIT1:
Fixes #1041.

## The bug

`retrieve_home_assistant_data()` runs `prepare_data()` on the retrieved frame for every `set_type`, and `prepare_data()`'s `set_zero_min` treatment (on by default) clips the whole DataFrame to `>= 0` and then replaces `0.0` with NaN. On the `battery_id` path this destroys exactly the data battery self-identification exists to consume: every discharge sample of the signed `sensor_power_battery` becomes NaN, so `_segment()` can never produce a discharge run and `identify()` always exits with `insufficient_data`. On a default configuration the feature is structurally unreachable, as @volschin's repro table in #1041 shows (110 negative samples before, 0 after). The same `replace(0.0, NaN)` line also eats a legitimately measured 0 % SoC sample.

## The fix

The shape agreed in the issue thread: an explicit `protected_columns` argument on `prepare_data()`, excluded from the clip and zero-to-NaN replacement only. Everything else `prepare_data()` does still applies to the protected columns (timezone conversion, duplicate-index cleanup, and the `sensor_replace_zero` / `sensor_linear_interp` treatment if a user configures them there). Skipping `prepare_data()` wholesale for `battery_id` was rejected in-thread because it would drag those unrelated behaviours along.

`retrieve_home_assistant_data()` passes the battery power and SoC entity ids as `protected_columns` when `set_type == "battery_id"`, reusing `_append_entity_ids` so the bare-string (N=1) and list (N>1, #1046) config forms are handled identically to retrieval. Every other `set_type` passes None.

When no protected column is present in the frame, the original two clip lines run verbatim, so the default path is the exact code that runs today (same pattern as the N=1 branch in #1046: the no-op claim stays auditable by inspection). The load column cannot be captured by protection even if a battery sensor were configured equal to the load sensor: the load column is renamed to `<sensor>_positive` before the clip, and protected names not present in the frame are ignored.

- `battery_id` retrieval only: the identifier now sees signed battery power (both flow directions) and exact 0 % SoC samples. This is the fix; it makes identification reachable on default config. Fits that previously exited `insufficient_data` will now actually run and be judged by the existing guardrails.
- Every other path: no change. `protected_columns` defaults to None and the guard branch keeps the existing whole-frame clip byte-for-byte. A no-op test pins that an omitted argument, an explicit None, and a list naming no present column all produce identical frames.

## Tests

- `test_prepare_data_protected_columns`: a battery_id-shaped frame keeps its negative power samples and a measured 0.0 SoC under protection, while an unprotected negative probe column in the same frame is still clipped to NaN (counterfactual) and load handling is unchanged. Fails on current master on the behavioural assertions.
- `test_prepare_data_protected_columns_default_noop`: omitted kwarg, explicit None, and an absent-column list produce frame-identical results, with the clip proven to have run.
- `test_retrieve_from_hass_battery_id_protected_columns`: the battery_id call site threads the right columns for both the N=1 string and N>1 list config forms; the naive-mpc test now also pins that non-battery paths pass None. Fails on current master.

Docs: one sentence in `docs/battery_identification.md` stating the signed sensors are preserved regardless of `set_zero_min`.

The two secondary observations from the issue are deliberately out of scope here, they have their own issues now: #1051 (trapezoid bridging over recorder gaps) and #1052 (failed fits not persisted).

## Summary by Sourcery

Preserve signed battery power and 0% state-of-charge samples during Home Assistant data preparation for battery self-identification while keeping existing clipping behavior unchanged elsewhere.

Bug Fixes:
- Prevent battery identification frames from having negative battery power samples and 0% SoC readings clipped to NaN by the default set_zero_min data cleaning.

Enhancements:
- Introduce a protected_columns parameter to prepare_data that exempts selected columns from the set_zero_min clip and zero-to-NaN replacement while still applying other preprocessing.
- Thread battery power and SoC sensor ids through battery_id retrieval and into prepare_data as protected_columns, covering both single-sensor and multi-sensor configurations.

Documentation:
- Clarify in battery_identification documentation that signed battery sensors are preserved on battery_id retrieval regardless of set_zero_min, while load data remains sanitized.

Tests:
- Add tests verifying that protected_columns preserves negative battery power and 0% SoC samples for battery_id while unprotected data is still clipped.
- Add tests ensuring that omitting protected_columns, passing None, or naming absent columns all reproduce the current clipping behavior exactly.
- Add a battery_id retrieval test asserting that protected_columns receives the correct battery power and SoC sensors and that non-battery_id paths do not set it.

### DIFF
--- a/docs/battery_identification.md
+++ b/docs/battery_identification.md
@@ -11,7 +11,7 @@ This is opt-in and default off. In this first version it is advisory only: it re
 - The measured battery state of charge, set as `sensor_battery_state_of_charge` (in percent).
 - Enough history. The fit needs weeks of signed power and state of charge with enough reasonably deep charge and discharge cycles. If the data is too shallow, or the fit fails an internal sanity check, it publishes nothing and keeps your configured values.
 
-The two sensors are only retrieved when battery self-identification is enabled, so they cost nothing on a normal run.
+The two sensors are only retrieved when battery self-identification is enabled, so they cost nothing on a normal run. Their signed values are kept intact on this retrieval regardless of the `set_zero_min` data cleaning setting, which continues to sanitize the load data as usual.
 
 ## Enabling it
 
--- a/src/emhass/command_line.py
+++ b/src/emhass/command_line.py
@@ -551,6 +551,12 @@ def _append_entity_ids(var_list: list, value) -> None:
         var_list.append(value)
 
 
+def _append_battery_id_sensors(var_list: list, retrieve_hass_conf: dict) -> None:
+    """Append the battery power and SoC sensor config values to ``var_list``."""
+    _append_entity_ids(var_list, retrieve_hass_conf["sensor_power_battery"])
+    _append_entity_ids(var_list, retrieve_hass_conf["sensor_battery_state_of_charge"])
+
+
 async def _retrieve_from_hass(
     set_type: str,
     retrieve_hass_conf: dict,
@@ -570,16 +576,13 @@ async def _retrieve_from_hass(
     if set_type == "battery_id":
         # Battery identification needs signed battery power and measured SoC,
         # one pair per battery. The load sensor stays at var_list[0] so
-        # prepare_data's load handling is unchanged (its set_zero_min clip, on
-        # by default, still applies to every retrieved column including
-        # these - a pre-existing limitation tracked upstream separately). A
-        # list value (N>1, already resolved by _identify_battery_impl) is
+        # prepare_data's load handling is unchanged; the battery columns are
+        # passed to prepare_data as protected_columns so its set_zero_min clip
+        # cannot destroy the discharge direction or a measured 0% SoC (#1041).
+        # A list value (N>1, already resolved by _identify_battery_impl) is
         # appended per-id, deduped against var_list; a bare string (N=1) is
         # the plain single-sensor case.
-        power_cfg = retrieve_hass_conf["sensor_power_battery"]
-        soc_cfg = retrieve_hass_conf["sensor_battery_state_of_charge"]
-        _append_entity_ids(var_list, power_cfg)
-        _append_entity_ids(var_list, soc_cfg)
+        _append_battery_id_sensors(var_list, retrieve_hass_conf)
         if logger:
             logger.debug(f"Variable list for battery_id retrieval: {var_list}")
     elif optim_conf.get("set_use_pv", True):
@@ -616,12 +619,21 @@ async def retrieve_home_assistant_data(
         )
     if not success:
         return False, None, days_list
+    protected_columns = None
+    if set_type == "battery_id":
+        # The identifier needs both flow directions of the signed battery
+        # power sensor and any legitimately measured 0% SoC sample, so these
+        # columns are exempt from the set_zero_min clip (#1041). Columns not
+        # present in the retrieved frame are ignored by prepare_data.
+        protected_columns = []
+        _append_battery_id_sensors(protected_columns, retrieve_hass_conf)
     rh.prepare_data(
         retrieve_hass_conf["sensor_power_load_no_var_loads"],
         load_negative=retrieve_hass_conf["load_negative"],
         set_zero_min=retrieve_hass_conf["set_zero_min"],
         var_replace_zero=retrieve_hass_conf["sensor_replace_zero"],
         var_interp=retrieve_hass_conf["sensor_linear_interp"],
+        protected_columns=protected_columns,
     )
     return True, rh.df_final.copy(), days_list
 
--- a/src/emhass/retrieve_hass.py
+++ b/src/emhass/retrieve_hass.py
@@ -1224,6 +1224,7 @@ class RetrieveHass:
         var_replace_zero: list[str],
         var_interp: list[str],
         skip_renaming: bool = False,
+        protected_columns: list[str] | None = None,
     ) -> bool:
         r"""
         Apply some data treatment in preparation for the optimization task.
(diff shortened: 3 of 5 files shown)


LesIT1:
Good catch, taken. The subTest context now carries both `power_cfg` and `soc_cfg`, so a failure in the N>1 case names the exact SoC config involved. In the latest revision.

davidusb-geek (approved):

Issue #770, opened 2026-03-27
UnboundLocalError: inv_stress_conf not assigned when optimization is infeasible (thermal battery, v0.17.1)


OptimalNothing90:
**Describe the bug**
When using a thermal battery load with physics-based building model and `naive-mpc-optim`, if `start_temperature` is below `min_temperatures[0]`, the optimization correctly identifies the problem as infeasible — but instead of returning a graceful error, EMHASS crashes with an `UnboundLocalError` in `optimization.py`:

```
WARNING  - Optimization failed with status: 'infeasible'. Retrying with relaxed constraints (Continuous LP)...
ERROR    - Relaxed optimization also failed with status: infeasible
ERROR    - Exception on request POST /action/naive-mpc-optim
  File "/app/src/emhass/optimization.py", line 2490
    if inv_stress_conf:
       ^^^^^^^^^^^^^^^
UnboundLocalError: cannot access local variable 'inv_stress_conf' where it is not associated with a value
```

**To Reproduce**

Call `POST /action/naive-mpc-optim` with a payload where:
- `def_load_config` contains a `thermal_battery` entry with physics-based building model
- `start_temperature` is below `min_temperatures[0]`

Minimal example (relevant excerpt):
```json
{
  "def_load_config": [{
    "thermal_battery": {
      "start_temperature": 21.4,
      "min_temperatures": [22.0, 22.0, ...],
      "max_temperatures": [24.0, 24.0, ...],
      "supply_temperature": 27,
      "carnot_efficiency": 0.4,
      "u_value": 0.21,
      "envelope_area": 362,
      "ventilation_rate": 0.12,
      "heated_volume": 651,
      "thermal_inertia_time_constant": 1.5
    }
  }],
  "prediction_horizon": 96
}
```

With `start_temperature: 21.4` below `min_temperatures[0]: 22.0`, the building is already below the minimum constraint at t=0. Combined with a cold overnight temperature forecast (~-3°C), both MIP and relaxed LP return infeasible — triggering the crash.

**Expected behavior**
A graceful error response (e.g. HTTP 400/500 with `{"status": "error", "message": "Optimization infeasible"}`) instead of a Python crash that leaves the caller with no actionable information.

**Workaround**
Ensure `start_temperature >= min_temperatures[0]`. In our case: lowering `min_temp_day` from 22°C to 21.5°C kept `start_temperature` above the minimum and resolved both the infeasibility and the crash.

**Screenshots**
N/A

[...]

PR #1059, opened 2026-08-04
fix: keep the relaxed-LP rescue one-shot so it cannot poison the cached problem (#1048)

### LINKED ISSUE #1048: ⁠[Bug / Unexpected Behavior] Unnecessary high charging power spikes & smoothing behavior upon modifying battery stress cost⁠
I am observing recurring, unmotivated high-power charging spikes in the battery schedule. The optimizer schedules charging at very high power over short time intervals instead of spreading out the charging power evenly at lower rates over a longer timeframe or the hole curtailment period.
Additionally, I noticed an unexpected behavior when trying to fix this:
Whenever I tweak the battery_stress_cost-parameter and re-run the MPC optimization, the resulting charging profile becomes significantly smoother 99% of the time exactly like I want to have it. Strangely, this smoothing happens both when increasing and when decreasing the battery stress penalty cost parameter.
It feels as though simply modifying any parameter forces bigger recalculation that solves the local solver trap, resulting in an optimal/smoothed curve, whereas standard consecutive runs seem to get stuck or produce erratic power spikes.

My battery stress cost setting was around 0-0.15 

Before (spike on the left side):
image

After:
image

2 seperate spike examples. 
image image

`{
  "adjusted_pv_model_max_age": 24,
  "adjusted_pv_regression_model": "LassoRegression",
  "adjusted_pv_solar_elevation_threshold": 10,
  "battery_charge_efficiency": 0.95,
  "battery_charge_power_max": 7000,
  "battery_discharge_efficiency": 0.95,
  "battery_discharge_power_max": 15000,
  "battery_dynamic_max": 0.9,
  "battery_dynamic_min": -0.9,
  "battery_identification_model_max_age": 24,
  "battery_identification_trust_tier": "observe",
  "battery_maximum_state_of_charge": 1,
  "battery_minimum_state_of_charge": 0,
  "battery_nominal_energy_capacity": 16000,
  "battery_soc_deficit_cost": 0,
  "battery_soc_deficit_threshold": 0.4,
  "battery_soc_surplus_cost": 0,
  "battery_soc_surplus_threshold": 0.9,
  "battery_stress_cost": 0.05,
  "battery_stress_segments": 48,
  "battery_target_state_of_charge": 1,
  "capacity_cost_per_kw": 0,
  "compute_curtailment": true,
  "continual_publish": true,
  "cost_forecast_per_deferrable_load": [
    "null"
  ],
  "costfun": "profit",
  "def_minimum_off_time": [
    0
  ],
  "def_minimum_on_time": [
    0
  ],
  "deferrable_load_groups": [],
  "deferrable_load_max_cost": [
    0
  ],
  "delta_forecast_daily": 1,
  "end_timesteps_of_each_deferrable_load": [
    0
  ],
  "historic_days_to_retrieve": 9,
  "influxdb_database": "home_assistant",
  "influxdb_host": "a0d7b954-influxdb",
  "influxdb_measurement": "W",
  "influxdb_port": 8086,
  "influxdb_retention_policy": "autogen",
  "influxdb_use_ssl": false,
  "influxdb_verify_ssl": false,
  "inverter_ac_input_max": 24000,
  "inverter_ac_output_max": 15000,
  "inverter_efficiency_ac_dc": 0.95,
  "inverter_efficiency_dc_ac": 0.95,
  "inverter_is_hybrid": true,
  "inverter_stress_cost": 0,
  "inverter_stress_segments": 10,
  "is_electric_load": [
    true
  ],
  "load_cost_forecast_method": "hp_hc_periods",
  "load_forecast_method": "mlforecaster",
  "load_negative": false,
  "load_offpeak_hours_cost": 0.24,
  "load_peak_hour_periods": {
    "period_hp_1": [
      {
        "start": "17:00"
      },
      {
        "end": "21:00"
      }
    ]
  },
  "load_peak_hours_cost": 0.24,
  "logging_level": "DEBUG",
  "lp_solver_mip_rel_gap": 0,
  "lp_solver_timeout": 45,
  "maximum_power_from_grid": 9000,
  "maximum_power_to_grid": 7800,
  "method_ts_round": "nearest",
  "minimum_power_of_deferrable_loads": [
    0
  ],
  "mlforecaster_weather_features": [
    "temp_air",
    "cloud_cover",
    "wind_speed",
    "heating_degree",
    "cooling_degree"
  ],
  "model_type": "load_forecast",
  "modules_per_string": [
    30
  ],
  "nominal_power_of_deferrable_loads": [
    0
  ],
  "num_lags": 48,
  "num_threads": 0,
  "number_of_deferrable_loads": 0,
  "open_meteo_cache_max_age": 60,
  "operating_hours_of_each_deferrable_load": [
    4
  ],
  "optimization_time_step": 30,
  "perform_backtest": false,
  "photovoltaic_production_sell_price": 0.068,
  "production_price_forecast_method": "constant",
  "publish_deferrable_load_states": false,
  "pv_inverter_model": [
    "Fronius_International_GmbH__Fronius_Primo_5_0_1_208_240__240V_"
  ],
  "pv_module_model": [
    "CSUN_Eurasia_Energy_Systems_Industry_and_Trade_CSUN295_60M"
  ],
  "sensor_battery_state_of_charge": "sensor.battery_state_of_charge",
  "sensor_linear_interp": [
    "sensor.emhass_invariable_lasten",
    "sensor.sigen_plant_consumed_power",
    "sensor.hausverbrauch_gesamt"
  ],
  "sensor_power_battery": "sensor.power_battery",
  "sensor_power_load_no_var_loads": "sensor.hausverbrauch_gesamt",
  "sensor_power_photovoltaics": "sensor.sigen_plant_pv_power",
  "sensor_power_photovoltaics_forecast": "",
  "sensor_replace_zero": [
    "sensor.power_photovoltaics",
    "sensor.p_pv_forecast",
    "sensor.hausverbrauch_gesamt"
  ],
  "set_battery_dynamic": false,
  "set_battery_first_priority": false,
  "set_deferrable_load_single_constant": [
    false
  ],
  "set_deferrable_max_startups": [
    0
  ],
  "set_deferrable_startup_penalty": [
    0
  ],
  "set_nocharge_from_grid": true,
  "set_nodischarge_to_grid": true,
  "set_total_pv_sell": false,
  "set_use_adjusted_pv": false,
  "set_use_battery": true,
  "set_use_battery_identification": false,
  "set_use_pv": true,
  "set_zero_min": true,
  "sklearn_model": "KNeighborsRegressor",
  "split_date_delta": "48h",
  "ssl_no_verify": false,
  "start_timesteps_of_each_deferrable_load": [
    0
  ],
  "strings_per_inverter": [
    1
  ],
  "surface_azimuth": [
    183
  ],
  "surface_tilt": [
    30
  ],
  "treat_deferrable_load_as_semi_cont": [
    false
  ],
  "use_influxdb": true,
  "use_websocket": false,
  "var_model": "sensor.power_load_no_var_loads",
  "weather_forecast_method": "csv",
  "weather_forecast_pv_quantile_bias": 0,
  "weight_battery_charge": 0.15,
  "weight_battery_discharge": 0
} `

LesIT1:
Fixes #1048.

## The bug

I dug into this in [the issue thread](https://github.com/davidusb-geek/emhass/issues/1048#issuecomment-5148868324): the relaxed-LP retry path permanently poisons the cached optimization problem.

The chain, on current master:

1. On a cache-hit run (`self.prob` already built), `batt_stress_conf` and `inv_stress_conf` are `None`. Only the `self.prob is None` build block populates them.
2. If that solve fails (`infeasible` / `unbounded` / `user_limit`, which is how a HiGHS timeout comes back), the relaxed retry rebuilds constraints and objective with those `None` confs. The stress epigraph constraints get skipped and the stress terms drop out of the objective.
3. `self.prob = prob_relaxed` then replaces the cached problem permanently. Every later cache-hit run re-solves the stress-free, binary-relaxed LP and reports plain "Optimal". Only the poisoning run itself ever shows "Optimal (Relaxed)".

That also explains the strangest symptom in the report: nudging `battery_stress_cost` in either direction changes `plant_conf_hash` in the cache key, forces a MISS, and rebuilds a clean problem. Any structural config change "fixes" it until the next failed solve. Runtime-excluded keys don't, which is a nice confirmation of the mechanism: tweaking `lp_solver_timeout` or `lp_solver_mip_rel_gap` themselves would not have helped, because those are excluded from the cache key.

The reporter's setup made this easy to hit: `lp_solver_mip_rel_gap: 0` (exact optimal instead of the shipped 1%) plus a 45s timeout on a true MILP (`inverter_is_hybrid: true`), so `user_limit` timeouts were routine.

## The fix

Three parts, all in `optimization.py` (the new attributes are initialised in `__init__`, everything else is in `perform_optimization`):

1. The stress configs built alongside the problem are now kept on the instance (`self._batt_stress_conf` / `self._inv_stress_conf`) and read back on cache-hit runs, so a retry rebuild gets the same configs (same CVXPY variables) the cached problem was built with. The existing #770 fix at the retry site already handled the constraint gating half of this gap; this closes it for the constraint and objective content too.
2. The rescue result is routed through a local `solved_prob`. `self.prob` is never replaced by `prob_relaxed`: the relaxed solve is a one-shot rescue for that run only, and the next run re-attempts the clean cached MILP.
3. Because the rescue problem is now discarded, everything the retry rebuild produces is kept run-local. The rebuilt thermal artifacts (predicted temperatures, heating demands, q_input) go into locals used only by that run's extraction, and the instance-held references the builders replace (the hybrid inverter transfer variables and the thermal `q_input_var` persistence hooks) are snapshotted and restored around the rescue. Without this, a rescued run would leave those references pointing at variables of the discarded problem, and later runs would extract thermal columns frozen at the rescue values.

The binary relaxation itself (forcing `treat_deferrable_load_as_semi_cont` and `set_deferrable_load_single_constant` off for the rescue solve) is clearly intentional, so I have not touched it. The bug was that it became permanent.

## Behaviour changes

- A rescued cache-hit run now honours the battery/inverter stress constraints and costs instead of silently dropping them.
- Poisoned follow-up runs are gone. Every run after a rescue re-attempts the real MILP. This also fixes the latent second bug where the semi-continuous / single-constant binary logic stayed permanently dropped after one rescue, which bites even with all stress costs at zero.
- Status reporting is now honest: each rescued run reports "Optimal (Relaxed)". Previously the poisoned follow-up runs reported plain "Optimal" while solving the degraded problem.
- A config that fails every run now pays the relaxed rebuild on each failing run instead of cheaply re-solving the retained relaxed LP. I think that cost is correct: the fast path was fast because it had silently stopped being the configured problem.
- One consequence of honest statuses: the `/api/v1/plan` store only refreshes on plain "Optimal" (command_line.py), so a config that fails every single run keeps serving the last clean plan instead of refreshing with rescue results labelled "Optimal". Previously it refreshed because the poisoned runs were mislabelled. Rescued runs still return and publish their own results through the normal response path, as they do today; only the plan store gate is stricter about them, and it always has been.
- For thermal loads with inertia, the q_input carried into the run after a rescue now comes from the cached problem's own last solve (the failed MILP's incumbent) rather than the discarded rescue problem. If that solve produced no values, the existing heating-demand fallback and forced rebuild take over, so there is no stuck state.

## Tests

- `test_relaxed_rescue_keeps_cached_problem`: the cached problem object survives a rescue (identity), and the following clean run reports plain "Optimal" again.
- `test_relaxed_rescue_does_not_poison_followup_runs`: behavioural check with a single-constant load and a price shape where the MILP must pick one contiguous block but the relaxed LP strictly prefers a split dispatch. On master the run after a rescue produces the split; with the fix it is back to the contiguous MILP dispatch.
- `test_cached_retry_rebuilds_stress_configs`: on a cache-hit retry, the objective rebuild receives the cached non-None stress config.
- `test_relaxed_rescue_keeps_thermal_artifacts_fresh`: regression guard for the run-local rescue. The thermal persistence hooks survive a rescue by identity, and the thermal columns of the run after a rescue track a changed outdoor forecast instead of freezing at the rescue values. This one is green on master too (the retained-problem behaviour masked it); it pins the property that keeps the one-shot rescue safe.

The first three fail on master at the behavioural assertion and pass with the fix. Full suite green locally.

## Summary by Sourcery

Prevent relaxed-LP rescue on cache-hit runs from permanently replacing the cached MILP problem and associated thermal artifacts, ensuring the rescue remains a one-shot fallback tied to the failing run only.

Bug Fixes:
- Ensure cache-hit retries reuse the original battery and inverter stress configurations so stress constraints and costs are preserved during relaxed-LP rescue solves.
- Avoid poisoning follow-up runs by keeping the cached MILP problem intact and routing relaxed-LP rescue results through a separate solved problem reference.
- Prevent thermal persistence hooks and artifacts from being overwritten by variables belonging to the transient relaxed problem, avoiding frozen or stale thermal outputs after a rescue.

Tests:
- Add regression tests that simulate solver failures to verify the cached problem survives relaxed-LP rescue and that subsequent runs revert to normal MILP behaviour and dispatch.
- Add tests to confirm cache-hit retries rebuild stress configs correctly and pass non-None cached stress configurations into the objective rebuild.
- Add tests that validate thermal artifacts remain fresh across rescues, with predicted temperatures and q_input values responding correctly to changing outdoor forecasts.

### DIFF
--- a/src/emhass/optimization.py
+++ b/src/emhass/optimization.py
@@ -323,6 +323,14 @@ class Optimization:
         # Note: The self.prob object will be constructed in a subsequent step
         self.prob = None
 
+        # Stress configs built alongside self.prob, kept for the relaxed-retry
+        # path: a solve failure on a cached problem rebuilds constraints and
+        # objective, and must reuse the same stress configs (same CVXPY
+        # variables) the cached problem was built with (issue #1048). Refreshed
+        # unconditionally whenever the build block runs.
+        self._batt_stress_conf = None
+        self._inv_stress_conf = None
+
     def _init_soc_recovery_params(self) -> None:
         """Initialize CVXPY parameters used for out-of-band SOC recovery.
 
@@ -5059,14 +5067,19 @@ class Optimization:
                         f"Deferrable load {k}: deactivated (no operating timesteps, not thermal)"
                     )
 
-        # Initialize stress config variables (needed by retry path even when
-        # self.prob is cached from a previous call, see #770)
-        inv_stress_conf = None
-        batt_stress_conf = None
+        # Stress configs are needed by the retry path even when self.prob is
+        # cached from a previous call (see #770). Start from the configs built
+        # with the cached problem so a cached-retry rebuild keeps the stress
+        # constraints and objective terms instead of silently dropping them
+        # (issue #1048); they reference the same CVXPY variables as self.vars.
+        inv_stress_conf = self._inv_stress_conf
+        batt_stress_conf = self._batt_stress_conf
 
         # Build Problem (Lazy Construction)
         if self.prob is None:
             self.logger.info("Building CVXPY problem structure...")
+            inv_stress_conf = None
+            batt_stress_conf = None
 
             # Start with bound constraints
             constraints = self.constraints[:]
@@ -5149,6 +5162,18 @@ class Optimization:
                 objective_expr.args[0] += penalty_terms_total
 
             self.prob = cp.Problem(objective_expr, constraints)
+            # Keep the stress configs paired with the problem they were built
+            # for, so cached-retry rebuilds see them (issue #1048).
+            self._batt_stress_conf = batt_stress_conf
+            self._inv_stress_conf = inv_stress_conf
+
+        # Thermal artifacts read by the extraction below. Locals so a relaxed
+        # retry can swap in its own rebuilt artifacts for this run only, while
+        # the instance attributes stay paired with the cached problem
+        # (issue #1048).
+        predicted_temps = self.predicted_temps
+        heating_demands = self.heating_demands
+        q_inputs = self.q_inputs
 
         # Solver Configuration
         solver_opts = {"verbose": False}
@@ -5234,6 +5259,13 @@ class Optimization:
                 f"Solver {selected_solver} failed: {e}. Checking status for fallback..."
             )
 
+        # The problem whose status/value the extraction below reads. Stays
+        # self.prob on a clean solve; points at the relaxed problem after a
+        # retry WITHOUT replacing self.prob, so the cached problem survives
+        # intact for the next run (issue #1048: caching prob_relaxed made the
+        # stress-free, binary-relaxed rescue permanent).
+        solved_prob = self.prob
+
         # Check for failure or "bad" status
         # Note: "user_limit" often means timeout. "infeasible" means configuration conflict.
         fail_statuses = ["infeasible", "unbounded", "user_limit", None]
@@ -5251,6 +5283,23 @@ class Optimization:
                 self.optim_conf.get("set_deferrable_load_single_constant", [])
             )
 
+            # The rebuild below replaces a few instance-held references with new
+            # objects that only live in the relaxed problem (the hybrid inverter
+            # transfer variables and the thermal q_input persistence hooks).
+            # Snapshot them so they can be restored after the rescue: instance
+            # state must keep pointing at the cached problem's objects, or later
+            # cache-hit runs would read variables the solver no longer touches
+            # (issue #1048).
+            hybrid_var_keys = ("p_dc_ac", "p_ac_dc", "is_dc_sourcing")
+            original_hybrid_vars = {
+                key: self.vars[key] for key in hybrid_var_keys if key in self.vars
+            }
+            original_q_input_vars = {
+                k: params["q_input_var"]
+                for k, params in self.param_thermal.items()
+                if "q_input_var" in params
+            }
+
             # Relax Configuration: Disable Binary Logic
             n_def = self.optim_conf["number_of_deferrable_loads"]
             self.optim_conf["treat_deferrable_load_as_semi_cont"] = [False] * n_def
@@ -5261,11 +5310,12 @@ class Optimization:
 
             # Re-apply main constraints
             self._add_main_power_balance_constraints(constraints_relaxed)
-            # (Note: We reuse previous stress configs as they don't change with relaxation)
-            # Guard on feature flags, not on the stress conf objects: stress conf is None
-            # on cached-problem retry paths (self.prob is not None), so guarding on the
-            # object itself would silently skip these constraints on every retry after the
-            # first call, leaving the relaxed problem without battery/inverter constraints.
+            # (Note: We reuse previous stress configs as they don't change with relaxation.
+            # On a cached-problem retry they come from self._*_stress_conf, the configs
+            # the cached problem was built with, see issue #1048.)
+            # Guard on feature flags, matching the build block's own gating: these
+            # builders also add the non-stress battery/inverter constraints, so they
+            # must run whenever the feature is on, stress cost or not.
             if self.plant_conf.get("inverter_is_hybrid", False):
                 self._add_hybrid_inverter_constraints(constraints_relaxed, inv_stress_conf)
             if self.optim_conf.get("set_use_battery", False):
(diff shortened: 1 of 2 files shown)


LesIT1:
The encapsulation this asks for is in place already: the `_status`/`_value` writes live only inside `_rescued_solve_patch`, and every rescue test goes through that one helper. No test touches the private attributes directly.

On brittleness, the production retry path already depends on the same internal: `perform_optimization` marks a rescued solve with `prob_relaxed._status = "Optimal (Relaxed)"`. So the helper mirrors an existing in-repo dependency rather than adding a new one. If a future cvxpy release changed how status and value are stored, the injection would stop tripping the retry gate and the first assertion on "Optimal (Relaxed)" would fail loudly, pointing straight at the helper. A version guard would add machinery without changing that outcome.

Happy to revisit if cvxpy ever exposes a public way to set these. That would be the cleaner seam for both the helper and the production status override.

davidusb-geek (approved):

PR #1063, opened 2026-08-05
fix: battery identification no longer counts throughput across recorder gaps (#1051)

### LINKED ISSUE #1051: Battery identification over-counts throughput: trapezoid integral bridges recorder gaps from change-only sensors

LesIT1:
Fixes #1051.

## The bug

Battery identification retrieves history resampled to `optimization_time_step` and the segmentation drops the NaN buckets a recorder gap leaves behind. The two samples on either side of the gap then sit next to each other, and the trapezoidal integral bridges them with a single step spanning the whole gap. With active power on both sides that fabricates `p_avg * gap_hours` of throughput out of nothing. That's the 89,809 Wh "discharge segment" from @JoshC1994's 79 h outage on a 10 kWh pack, and the milder version of it taxes every gap in @volschin's change-only profile (69 gaps over 30 min in his 14 day census, 61% of wall-clock inside gaps).

## The fix

A step between consecutive observed samples longer than 3x the expected sample step is treated as a recorder gap. The expected step is `optimization_time_step`, now passed into `BatteryIdentification` at both call sites; without it (standalone use) the estimator falls back to the median spacing of the data. Gap handling is an internal constant, not a new config param.

A gap ends the current charge/discharge run at the last observed sample, mirroring the existing end-of-data flush, and any new run starts at the first sample after the gap. So a gap contributes zero throughput and zero dSoC. Gap steps are also excluded from the sign-convention vote, which otherwise counts one stale pre-gap power sample against the real steps.

Dropouts up to two consecutive missing resample buckets (dt = 3x step) are still bridged, so ordinary recorder jitter doesn't split runs.

## Design choice, disclosed

There were two ways to make a gap contribute zero throughput:

1. Drop only the oversized trapezoid step but keep the run alive across the gap, keeping its full dSoC span.
2. End the run at the gap (what this PR does).

Option 1 looks gentler on segment counts, but it credits the run with SoC change that happened while nobody was watching. On a multi-day outage the SoC drifts, so throughput excludes the gap while dSoC includes it, and the capacity estimate biases low. It can also stitch two shallow cycles on either side of a gap into one segment that clears the 10-point depth floor on unobserved state. Ending the run keeps every segment's throughput and dSoC internally consistent.

The honest cost: on gap-heavy history, runs that used to be bridged now split, and each piece has to clear the depth and duration floors on its own. A profile like volschin's can land on `insufficient_data` instead of a fabricated fit. That's the intended behaviour change: no estimate beats a wrong estimate, and his cycles are densely reported while active, so real segments survive.

Forward-filling change-only sensors at retrieval (so idle periods come back as real flat data instead of gaps) is a separate improvement and stays a follow-up, as discussed in the issue.

## Behaviour changes

- Gap-free history is byte-identical to before (strict `>` threshold, a regular grid never trips it).
- Gapped history: fewer, cleaner segments; fits that only passed because of fabricated throughput now fail honestly (`insufficient_data` or unchanged guardrail rejections).
- No new config params, no schema or persistence changes.

## Tests

Eleven new tests in `tests/test_battery_identification.py` (`TestGapHandling`). Four are proven red on master before the fix, on the behavioural asserts:

- a 79 h gap adds zero throughput and zero dSoC (red on master: 88,000 Wh counted vs 9,000 Wh actually measured)
- end to end, a 79 h mid-run gap no longer corrupts the fit (red on master with exactly the reported failure signature, `rejected_sanity_check`)
- gap steps don't vote on the sign convention (red on master)

The rest pin the fix's own edges: the change-only idle-gap profile keeps its segments and stays accurate, a two-bucket dropout still bridges (with segment depth checked, on 30 min and on odd grid steps where a float-hours comparison would misclassify), the same guarantees hold for discharge runs and for lone samples at the series edges, explicit `time_step` overrides the median inference, and gap detection survives a microsecond-resolution index (pandas 3 default, where a naive epoch-integer comparison against nanoseconds would silently disable it).

Full suite green locally on Windows (970 passed, 32 xfailed), pandas 3.0.3.

@volschin you offered to re-run your gap census against a patched build, that would be the perfect acceptance check for scenario 2 here. @JoshC1994 your 79 h case is scenario 1 and is what the end-to-end test reproduces.

## Summary by Sourcery

Prevent battery identification from counting energy throughput and state-of-charge changes across recorder gaps, using the configured optimization time step to detect gaps and ending runs at gap boundaries.

Bug Fixes:
- Stop trapezoidal integration from fabricating throughput over long recorder outages or change-only sensor gaps by treating oversized steps as gaps with zero contribution.
- Exclude gap steps from sign-convention detection so stale pre-gap power cannot flip charge/discharge labeling.

Enhancements:
- Add optional time_step parameter to BatteryIdentification to control gap detection and keep standalone usage functional.
- Make gap detection robust to different timestamp resolutions and odd sampling grids, preserving segments through short dropouts.

Documentation:
- Document that long gaps in battery history are excluded from identification, may lead to insufficient-data outcomes, and that short dropouts remain bridged.

Tests:
- Add a comprehensive TestGapHandling suite covering gap throughput/dSoC behavior, end-to-end fit integrity, short-dropout bridging, discharge direction gaps, edge samples, explicit time_step behavior, and sign-convention voting.

### DIFF
--- a/docs/battery_identification.md
+++ b/docs/battery_identification.md
@@ -13,6 +13,8 @@ This is opt-in and default off. In this first version it is advisory only: it re
 
 The two sensors are only retrieved when battery self-identification is enabled, so they cost nothing on a normal run. Their signed values are kept intact on this retrieval regardless of the `set_zero_min` data cleaning setting, which continues to sanitize the load data as usual.
 
+Stretches of history with no recorded data are excluded from the fit. Any step between two samples longer than three times the `optimization_time_step` counts as a recorder gap: the battery state across it is unobserved, so it contributes no energy throughput and no state of charge change, and any charge or discharge run in progress ends at the last sample before the gap. This means history riddled with long gaps (a recorder outage, or a power sensor that only reports on changes and stays silent for hours) can honestly come back as "not enough data" rather than producing an estimate from invented energy. Shorter dropouts, up to two consecutive missing steps, are still bridged.
+
 ## Enabling it
 
 Set these in your configuration:
--- a/src/emhass/battery_identification.py
+++ b/src/emhass/battery_identification.py
@@ -58,6 +58,13 @@ POWER_DEADBAND_W = 25.0  # |power| below this is treated as idle (sign hysteresi
 SOC_REVERSAL_HYSTERESIS = 1.0  # SoC points of counter-move tolerated before a run splits
 MIN_SEGMENT_DURATION = pd.Timedelta(minutes=20)
 MIN_SOC_SWING_PER_SEGMENT = 10.0  # a segment must move at least this many SoC points to be "deep"
+# A step between consecutive observed samples longer than this many multiples of
+# the expected sample step is a recorder GAP (#1051): the battery state across it
+# is unobserved, so the step must contribute neither throughput nor dSoC. Gaps
+# end the current run at the last observed sample. Up to two consecutive missing
+# resample buckets (dt == 3x step) are tolerated as benign recorder jitter and
+# still bridged by the trapezoid.
+MAX_GAP_STEP_MULTIPLE = 3.0
 
 # Sufficiency floors
 MIN_DEEP_SEGMENTS = 3  # per direction, before a capacity fit is attempted
@@ -224,8 +231,18 @@ def _intercept_diag(x: np.ndarray, y: np.ndarray) -> float:
 class BatteryIdentification:
     """Pure estimator: history in, :class:`BatteryIdentificationResult` out."""
 
-    def __init__(self, logger: logging.Logger):
+    def __init__(self, logger: logging.Logger, time_step: pd.Timedelta | None = None):
+        """
+        :param logger: Logger for diagnostics.
+        :param time_step: The expected spacing of the history samples (the
+            retrieval resample step, ``optimization_time_step`` in production).
+            Used to tell recorder gaps apart from ordinary steps. When ``None``
+            the step is inferred as the median spacing of the data itself,
+            which keeps the estimator usable standalone but is only reliable
+            while most consecutive samples are actually adjacent.
+        """
         self.logger = logger
+        self.time_step = time_step
 
     # -- segmentation ---------------------------------------------------------
     def _segment(self, df: pd.DataFrame, power_col: str, soc_col: str) -> list[_Segment]:
@@ -246,15 +263,49 @@ class BatteryIdentification:
         # Float hours since the first sample (tz-safe, resolution-independent).
         hours_axis = (times - times[0]).total_seconds().to_numpy() / 3600.0
 
+        # Recorder gaps (#1051): after the retrieval resample, missing history
+        # shows up as NaN buckets which the dropna above removes, leaving one
+        # oversized step between the samples on either side of the gap. The
+        # battery state across that step is UNOBSERVED, so it must contribute
+        # neither throughput (the trapezoid would fabricate p_avg * gap_hours)
+        # nor dSoC. Oversized steps are treated as run boundaries below and are
+        # excluded from the sign vote here.
+        # Compare spacings in integer nanoseconds, not float hours: a dropout of
+        # exactly MAX_GAP_STEP_MULTIPLE missing-plus-one buckets must stay
+        # bridged on EVERY grid, and float division (1 min = 1/60 h) turns that
+        # equality into a spurious ">" on steps that are not a binary fraction
+        # of an hour. The index is normalized to nanoseconds first: asi8 counts
+        # in the index's own resolution (microseconds by default on pandas 3),
+        # while Timedelta.value is always nanoseconds.
+        dt_ns = np.diff(times.as_unit("ns").asi8)
+        if self.time_step is not None:
+            base_step_ns = float(pd.Timedelta(self.time_step).as_unit("ns").value)
+        else:
+            base_step_ns = float(np.median(dt_ns)) if len(dt_ns) else 0.0
+        if base_step_ns > 0:
+            gap_mask = dt_ns > MAX_GAP_STEP_MULTIPLE * base_step_ns
+        else:
+            # Degenerate spacing (e.g. duplicate timestamps dominating the
+            # median): no reliable notion of a "normal" step, so no gap logic.
+            gap_mask = np.zeros(len(dt_ns), dtype=bool)
+        n_gaps = int(gap_mask.sum())
+        if n_gaps:
+            self.logger.debug(
+                "battery_identification: %d recorder gap(s) longer than %.2g h "
+                "excluded from segmentation",
+                n_gaps,
+                MAX_GAP_STEP_MULTIPLE * base_step_ns / 3.6e12,  # ns -> h (3600 s/h * 1e9 ns/s)
+            )
+
         # Auto-detect the sign convention from the data itself. Each rising-SoC
         # interval [k, k+1] is CAUSED by the power held over it, power[k] (the
         # same convention the trapezoid integral uses), NOT power[k+1] - pairing
         # with the end sample mislabels pulsed charge-then-discharge data. Over
         # charging intervals the causing power must be positive; if the net vote
         # is negative the meter reports charge as negative and we flip so that
-        # positive = charge.
+        # positive = charge. Gap steps are unobserved and do not vote.
         d_soc_step = np.diff(soc)
-        rising = d_soc_step > 0
+        rising = (d_soc_step > 0) & ~gap_mask
         if rising.any():
             charge_power_vote = float(np.sum(power[:-1][rising]))
             self.logger.debug(
@@ -284,6 +335,19 @@ class BatteryIdentification:
             step_dir = 0
             if abs(power[i]) >= POWER_DEADBAND_W:
                 step_dir = 1 if power[i] > 0 else -1
+            if gap_mask[i - 1]:
+                # The step into sample i spans a recorder gap. The current run
+                # (if any) ends at the last observed sample, mirroring the
+                # end-of-data flush. Any new run starts AT the first post-gap
+                # sample - never at i-1, which would bridge the gap back into
+                # a segment's dSoC and duration.
+                _flush(i - 1)
+                run_dir = step_dir
+                if run_dir != 0:
+                    run_start = i
+                    run_soc_extreme = soc[i]
+                    run_extreme_idx = i
+                continue
             if run_dir == 0:
                 if step_dir != 0:
                     run_dir = step_dir
--- a/src/emhass/command_line.py
+++ b/src/emhass/command_line.py
@@ -1448,9 +1448,9 @@ async def _identify_battery_impl(
             )
             return
         configured_capacity_wh = float(plant_conf.get("battery_nominal_energy_capacity", 0) or 0)
-        result = BatteryIdentification(logger).identify(
-            df, power_col, soc_col, configured_capacity_wh
-        )
+        result = BatteryIdentification(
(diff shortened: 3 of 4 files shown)


LesIT1:
Fair nits, all three taken and folded into the amended head, with straight quotes rather than the curly ones from the suggestion block.

volschin:
Acceptance run for scenario 2 (the change-only / gap-heavy profile) is done, on live history, not synthetic data.

**Setup.** `master` (ad9b919) vs this branch (ac23236), same pickled frame, retrieved through the real `battery_id` path (`retrieve_home_assistant_data` → `rh.get_data` + `prepare_data(protected_columns=...)`) against my HA, `optimization_time_step=30`, `sensor.marstek_venuse_power` / `sensor.marstek_venuse_state_of_charge`, Marstek Venus E, nameplate 5.12 kWh. Python 3.12, pandas 3.0.5, numpy 2.2.6 on Linux.

### Gap census, re-run as promised

14 days (2026-07-22 → 2026-08-05), 684 buckets:

| | |
|---|---|
| observed samples | 274 / 684 buckets (40.1 %) |
| steps > 3 × 30 min | 35 |
| wall clock inside those gaps | 8 d 06:30 (59.1 %) |
| largest gap | 20 h |
| raw trapezoid over gap steps | **94.9 kWh of 180.4 kWh total** |

So on my profile a bit over half of the step integral was being invented. After segmentation the damage is smaller but still large: master builds 22 segments totalling **97.3 kWh** of throughput, this branch builds 21 totalling **77.2 kWh** — 20.1 kWh (21 %) of segment throughput was fabricated across gaps.

The single worst artefact is visible directly: master has a "charge segment" 2026-07-31 08:00 → 2026-08-01 16:00, 32 h long, dSoC +51, throughput 9581 Wh, slope 187.9 Wh/SoC-point against a cohort where every other segment sits between 22 and 92. It bridged a 20 h gap. On this branch it is gone.

**No-regression check:** the longest strictly regular (no missing bucket at all) slice in my data, 11 h / 23 rows, segments byte-identical on both revisions (same sha1 over direction/start/end/dSoC/throughput). The `>` threshold does not trip on a regular grid, confirmed on real data.

### Fit outcome, 14 days

| | master | this PR |
|---|---|---|
| status | `low_confidence` | `rejected_sanity_check` |
| capacity (TLS) | 6.68 kWh, CI [5.59, 7.97] | 5.95 kWh |
| Theil-Sen cross-check | 6.02 kWh | 6.11 kWh |
| RTE | 0.774 | 0.9999 → sqrt = 1.000, rejected |
| equal-SoC cycles | 10 | 0 |
| segments (chg/dis) | 12 / 10 | 10 / 11 |

Neither publishes, so nothing regressed in practice — but the intercept diagnostic is the interesting part: **-299 210** on master versus **+84** on this branch. The gap-bridged segments were what pulled the fit off the origin.

### Fit outcome, 26 days — this is the real result

Same config with `historic_days_to_retrieve=30` (5 days had no recorder data, 24 d 16 h span, 1212 buckets, 47.4 % observed, 58 gaps, 51.6 % of wall clock inside them, 122.2 kWh of 296.2 kWh raw trapezoid over gap steps):

| | master | this PR |
|---|---|---|
| status | `rejected_sanity_check` | **`ok`** |
| capacity (TLS) | 7.39 kWh | **5.65 kWh**, CI [5.49, 5.85] |
| Theil-Sen | 5.47 kWh (diverges 25.9 %) | 5.41 kWh (diverges 4.3 %) |
| RTE | 0.575 | 0.991 (energy balance 0.989) |
| equal-SoC cycles | 0 | 21 |
| segments (chg/dis) | 24 / 20 | 22 / 21 |

Independent ground truth for this pack, measured by my own HA automations from cumulative energy counters (anchor at SoC > 94 %, read out at SoC < 15 %, both directions), not from resampled power:

* usable AC capacity, discharge side: 4.864 kWh per 100 SoC points
* DC capacity, `sqrt(charge × discharge)`: **5.371 kWh** (nameplate 5.120)
* cycle efficiency `discharge / charge`: **83.2 %**; monthly meter efficiency sensor: 82.3 %

The identifier's `C = 100 * sqrt(S_chg * S_dis)` is the same geometric-mean construction, so 5.65 kWh is directly comparable to my 5.371 kWh: **+5.2 %** on this branch, against **+37.6 %** on master. That is the fix working, on a profile with 52 % of wall clock missing.

### One thing worth knowing before merge

On the 26 day window this branch does not just fail more honestly, it now **passes** the guardrails and would publish under `trust_tier: suggest` — with RTE 0.991 (one-way 0.995, just inside `SQRT_RTE_HIGH = 0.999`) against an independently metered 83.2 %. The energy-balance cross-check agrees at 0.989, but it is computed from the same resampled samples, so it is not an independent witness here.

That bias is orthogonal to this PR — it is the change-only sampling itself, i.e. exactly the forward-fill follow-up you deferred — but the effect is that the gap fix removes what was masking it. Before this PR my profile produced a wrong-and-rejected fit; after it, a good capacity and a too-high RTE that clears every gate. If `suggest` writes `battery_round_trip_efficiency ≈ 0.99` into an optimizer that should be seeing 0.83, the optimizer will over-cycle.

Two possible mitigations, both cheap and neither blocking this PR:

1. tighten `SQRT_RTE_HIGH` (0.995 rather than 0.999 — a real AC-coupled pack that lossless does not exist), or
2. make the energy-balance cross-check gating rather than informational when it lands within a hair of the slope-ratio value, since agreement between two statistics drawn from the same throughput series is not corroboration.

FWIW I also tried the deferred forward-fill as a side experiment: last-value hold on the raw states, time-weighted mean into the 30 min grid. It removes every gap (0 gaps, 659 rows) but the fit gets *worse*, not better — RTE 1.76, TLS 8.01 kWh vs Theil-Sen 5.35 kWh, `rejected_sanity_check`. Holding the last value across `unavailable` stretches keeps a stale non-zero power alive, so the follow-up will need explicit staleness handling; it is not a drop-in.

### Tests

`tests/test_battery_identification.py` on this branch: **46 passed, 4 subtests passed** on Linux / pandas 3.0.5 (your run was Windows, so that is a second platform).

On master the new suite is red for the right reason, not only on the missing kwarg: with a shim that accepts and ignores `time_step`, **8 of the `TestGapHandling` tests still fail on their behavioural asserts** (`test_gap_contributes_zero_throughput`, `..._zero_dsoc`, `test_identify_recovers_truth_despite_gap`, `test_gap_steps_do_not_vote_on_sign_convention`, `test_discharge_gap_contributes_nothing`, `test_gap_at_series_edges_is_harmless`, `test_gap_detected_on_microsecond_index`, `test_explicit_time_step_overrides_inference`).

**Verdict from my side: acceptance passed for scenario 2.** Capacity error drops from +37.6 % to +5.2 % against an independently metered value, gap-free history is provably untouched, and the one remaining problem is the follow-up you already scoped — plus the guardrail note above, which I would rather see handled before `suggest` becomes anyone's default.

LesIT1:
That's the acceptance evidence this fix needed, thank you for putting a metered pack behind it. +5.2% against your counter-derived 5.371 kWh, with the same geometric-mean construction on both sides, is about as direct a validation as it can get. The byte-identical check on your regular 11h slice matters just as much, since "gap-free history untouched" was a claim I could only prove synthetically before. And the shim run on master is appreciated: 8 behavioural failures means the tests are pinning the semantics, not just the signature. One housekeeping note: the head you tested (ac23236) and the current head (c27b71be) are tree-identical, the re-push was only to re-roll a CI flake on the Windows job.

On the guardrail point, I agree, and I'd rather it not stay buried in a PR thread. You're right that this fix removes the masking: before, your profile produced a fit wrong enough to get rejected; now it produces a good capacity plus an RTE inherited straight from the change-only sampling bias, and 0.995 one-way sails through a 0.999 threshold. Of your two options I'd take the threshold tightening first, since it's one constant and defensible on physics alone (no real AC-coupled pack sits at 99.9%). On the second, if I'm reading you right, you'd have suspiciously close agreement trip the gate: two statistics drawn from the same throughput series landing within a hair of each other tells you nothing, so it shouldn't be allowed to read as confirmation. The catch is that closeness is also what a healthy fit looks like, since on dense clean data both estimators converge on the true RTE, so a closeness trigger would need conditioning on something more (proximity to the guardrail edge, say) before it stops rejecting exactly the fits we want. That needs more thought than a PR comment, so I'd put both mitigations in the same place.

Unless David prefers it handled in this PR, I'll open a small separate issue for the RTE guardrail so it gets its own decision, with your 0.991-vs-0.832 case as the motivating data. Your forward-fill experiment goes straight into the follow-up scoping too: last-value hold fabricating energy across unobserved stretches is the trap the issue thread called when we agreed the dt bound had to sit underneath any fill, and RTE 1.76 turns that from an argument into a measurement.

davidusb-geek:
> Unless David prefers it handled in this PR, I'll open a small separate issue for the RTE guardrail so it gets its own decision, with your 0.991-vs-0.832 case as the motivating data. Your forward-fill experiment goes straight into the follow-up scoping too: last-value hold fabricating energy across unobserved stretches is the trap the issue thread called when we agreed the dt bound had to sit underneath any fill, and RTE 1.76 turns that from an argument into a measurement.

Lets merge this one and you can open a separate issue for the remaining improvement (unless is already done, haven't read the latest issues and PR)

davidusb-geek (approved):

PR #1064, opened 2026-08-05
Fix: duplicate forecast_dates when Forecast is constructed on an exact half-interval tie second

LesIT1:
Fixes the long-standing intermittent CI failure in tests/test_optimization.py:

```
ValueError: Length mismatch: Expected axis has 24 elements, new values have 48 elements
```

(sometimes 25 instead of 24). This has been failing random CI runs on otherwise green PRs for months, most recently twice in one day on #1063. It is not a test problem: the same crash can hit any real run on the default config.

## Root cause

With `method_ts_round: "nearest"` (the shipped default), `Forecast.__init__` keeps the raw wall time as `start_forecast`, only stripping microseconds. The `forecast_dates` builder then rounds the whole `date_range` in UTC, stamp by stamp.

If the constructor happens to run at exactly HH:15:00 or HH:45:00 (seconds zero), every stamp in the 30 minute range sits exactly halfway between two grid points. Pandas rounds half-to-even, so neighbouring stamps collapse into duplicates: a :45 start gives 24 unique values out of 48, a :15 start gives 25. Those are exactly the two numbers CI has been reporting. The first downstream index assignment then crashes, for example `forecast_out.index = self.forecast_dates` in `_get_load_forecast_typical`.

Deterministic repro (pandas 2.2.3, 3.0.3 and 3.0.5):

```python
import pandas as pd
freq = pd.Timedelta("30min"); tz = "Europe/Paris"
start = pd.Timestamp("2026-08-05 04:45:00", tz=tz)
end = start + pd.DateOffset(days=1)
fd = (pd.date_range(start, end - freq, freq=freq, tz=tz).tz_convert("utc")
      .round(freq, ambiguous="infer", nonexistent="shift_forward").tz_convert(tz))
print(len(fd), len(fd.unique()))   # 48, 24
```

So every `Forecast` construction is a 2-seconds-per-hour lottery, on any OS and any pandas since 2.2. That is also why the flake never correlated with anything: it looked Windows-only and time-of-day dependent for a while, but that was sampling noise over which runners happened to construct a `Forecast` in a tie second.

## The fix

Round the start once instead of rounding the built index. The "nearest" branch now aligns `start_forecast` with a single-stamp round through the same UTC route the index round used (`tz_convert("utc").round(freq).tz_convert(tz)`), and the index-wide `.round()` chain is deleted from the three date_range builders (`__init__`, `get_forecast_days_csv`, `set_cached_forecast_data`). A range built from an aligned start at a fixed freq is aligned by construction, so the per-stamp tie ambiguity disappears.

Rounding the scalar in UTC keeps the DST safety the old chain had: UTC has no ambiguous or nonexistent times, which is also why the old `ambiguous="infer"` kwarg was inert once f754cf99 moved the round to UTC (scalar `Timestamp.round` rejects `"infer"` anyway).

The "first" and "last" branches already floor/ceil the start, so for them deleting the round changes nothing in timezones whose UTC offset is a multiple of the step. In zones where it is not (Chatham, Kathmandu or Eucla with a 30 minute step, half-hour-offset zones with a 60 minute step), the floored local start put every stamp exactly halfway between UTC grid points, so the old round collapsed the index on every single run; those zones were unusable and this fixes them too. I swept over 2000 start/mode combinations through the old and new pipelines, including DST-crossing days in both directions and offset-timezone cases: byte-identical output everywhere the old code did not crash, outside the narrow DST windows disclosed below.

## Behaviour change, disclosed

A "nearest" run whose constructor lands in a tie second now gets a deterministic aligned index instead of a crash, and under "first" and "last", timezones whose UTC offset is not a multiple of the optimization step go from crashing on every run to working.

Every start that did not crash before produces byte-identical `forecast_dates`, with one narrow exception: "nearest" starts inside a window up to half a step wide right before a DST transition (or before the same wall-clock time on its eve). There the shorter index is an exact prefix of the longer and the length differs by exactly the DST hour, because aligning the start moves one end of the one-day span across the transition. The new length is always consistent with the span the index actually covers (a 23-hour day gets 46 half-hour slots). That is a few windows per year, each under half a step wide.

## Tests

- Tie-second tests pin `pd.Timestamp.now` (a small module-attribute shim, no new dependencies) to :15 and :45 starts and assert the 48-stamp index stays unique. On the unfixed builder they fail with 24/25 unique stamps.
- An end-to-end test reproduces the exact CI ValueError through `get_load_forecast(method="typical")` on the unfixed code.
- DST tie cases: a spring-forward day (46 stamps) and a fall-back day (50 stamps).
- A no-op guard recomputes the old build-then-round pipeline inline at two non-tie seconds (one rounding down, one up) and asserts stamp-for-stamp equality for all three `method_ts_round` modes, pinning the aligned-start construction as byte-identical away from ties.
- An Australia/Adelaide (UTC+9:30) case at a 60 minute step pins the start alignment to the UTC grid, the same route the old index-wide round used, rather than the local wall-time grid.

## Why no issue first

The bug is fully characterized by the repro above, the fix is small and mechanical, and there is no design choice to discuss, so an issue would just restate this PR.

One out-of-scope observation while in here: `utils.get_forecast_dates` floors unconditionally regardless of `method_ts_round`, so the two date builders disagree about "nearest". Pre-existing, unchanged by this PR, happy to look at it separately if useful.

## Summary by Sourcery

Align forecast start timestamps in UTC to prevent duplicate forecast_dates and length mismatches, and simplify date_range construction to rely on the aligned start instead of rounding the full index.

Bug Fixes:
- Prevent intermittent ValueError length mismatches caused by duplicate forecast_dates when Forecast is constructed on half-interval tie seconds.
- Fix forecast date construction for timezones whose UTC offset is not a multiple of the optimization step, ensuring unique indices across DST transitions and non-integer offsets.

Enhancements:
- Unify forecast date construction across Forecast initialization, CSV export, and cached forecast data to use aligned starts without per-stamp rounding.
- Add tests covering tie-second behavior, DST transitions, non-tie equivalence with the previous pipeline, and non-integer offset timezones to pin the new behavior.

Tests:
- Introduce async and synchronous tests that pin pd.Timestamp.now to tie and non-tie instants to verify unique forecast_dates, end-to-end load forecast stability, DST behavior, and UTC-based rounding in offset timezones.

### DIFF
--- a/src/emhass/forecast.py
+++ b/src/emhass/forecast.py
@@ -215,7 +215,18 @@ class Forecast:
             self.params = orjson.loads(params)
 
         if self.method_ts_round == "nearest":
-            self.start_forecast = pd.Timestamp.now(tz=self.time_zone).replace(microsecond=0)
+            # Round the start once, in UTC (no DST ambiguity there). A range built
+            # from an aligned start at a fixed freq is aligned by construction;
+            # rounding the built index stamp-by-stamp instead lets round-half-to-even
+            # collapse neighbouring stamps into duplicates whenever now() falls
+            # exactly on a half-interval boundary (HH:15:00 / HH:45:00 at 30 min).
+            self.start_forecast = (
+                pd.Timestamp.now(tz=self.time_zone)
+                .replace(microsecond=0)
+                .tz_convert("utc")
+                .round(self.freq)
+                .tz_convert(self.time_zone)
+            )
         elif self.method_ts_round == "first":
             self.start_forecast = (
                 pd.Timestamp.now(tz=self.time_zone).replace(microsecond=0).floor(freq=self.freq)
@@ -242,16 +253,11 @@ class Forecast:
             self.end_forecast = (self.start_forecast + pd.DateOffset(days=_delta_days)).replace(
                 microsecond=0
             )
-        self.forecast_dates = (
-            pd.date_range(
-                start=self.start_forecast,
-                end=self.end_forecast - self.freq,
-                freq=self.freq,
-                tz=self.time_zone,
-            )
-            .tz_convert("utc")
-            .round(self.freq, ambiguous="infer", nonexistent="shift_forward")
-            .tz_convert(self.time_zone)
+        self.forecast_dates = pd.date_range(
+            start=self.start_forecast,
+            end=self.end_forecast - self.freq,
+            freq=self.freq,
+            tz=self.time_zone,
         )
         if (
             params is not None
@@ -1541,7 +1547,15 @@ class Forecast:
         """
         start_forecast_csv = pd.Timestamp.now(tz=self.time_zone).replace(microsecond=0)
         if self.method_ts_round == "nearest":
-            start_forecast_csv = pd.Timestamp.now(tz=self.time_zone).replace(microsecond=0)
+            # Same start alignment as __init__: round once in UTC so the built
+            # range needs no stamp-by-stamp round (see the comment there).
+            start_forecast_csv = (
+                pd.Timestamp.now(tz=self.time_zone)
+                .replace(microsecond=0)
+                .tz_convert("utc")
+                .round(self.freq)
+                .tz_convert(self.time_zone)
+            )
         elif self.method_ts_round == "first":
             start_forecast_csv = (
                 pd.Timestamp.now(tz=self.time_zone).replace(microsecond=0).floor(freq=self.freq)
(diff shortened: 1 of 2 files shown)


LesIT1:
Point 3 taken, and it taught me something. I had assumed `pd.Timestamp` was a C extension type that rejects class attribute assignment, which is why the test swapped the whole module attribute for a shim. Probed it on both ends of the supported range (pandas 2.2.3 and 3.0.3) and class-level assignment of `now` works on both, so the test now patches only `Timestamp.now`, scoped to the constructor call, and the shim class is gone. Amended in a49054d9.

On the two extraction suggestions, I want to keep this PR surgical:

- The duplicated "nearest" alignment mirrors the existing structure: the "first" and "last" floor/ceil expressions are already duplicated verbatim between `__init__` and `get_forecast_days_csv`. Pulling just the new branch into a helper would leave the other two duplicated and make the two sites read differently. Extracting all three is a real cleanup, but a separate one.
- Same for centralizing the three `date_range` builds: they look alike but take different inputs (`get_forecast_days_csv` adds `timedelta_days` to its end, `set_cached_forecast_data` recomputes its end from `delta_forecast_daily`), and they were just as parallel before this change; the diff only removed the shared round chain.

Happy to do that consolidation as a follow-up PR if there is appetite for it.

davidusb-geek (approved):

PR #1070, opened 2026-08-11
fix: tighten the battery identification RTE guardrail to reject near-lossless fits

### LINKED ISSUE #1069: Battery identification: RTE guardrail admits a near-lossless fit that change-only sampling produces on real data (follow-up from #1063)
Follow-up promised in #1063 (see the merge comment there). This is scoped to the RTE guardrail only.

## Problem

The gap-throughput fix in #1063 removed what was masking a change-only sampling bias in the RTE estimate. @volschin's acceptance run on that PR showed it directly (all numbers from [his comment](https://github.com/davidusb-geek/emhass/pull/1063#issuecomment-5188064123), on live history from his Marstek Venus E):

- On his ~26 day window (24d16h span, 47.4% of buckets observed, 51.6% of wall clock inside recorder gaps) the fit now passes every guardrail: status `ok`, capacity 5.65 kWh with a tight CI, RTE **0.991** (one-way sqrt 0.995, just inside `SQRT_RTE_HIGH = 0.999`).
- His independently metered ground truth for the same pack, from cumulative energy counters rather than resampled power, is a cycle efficiency of **83.2%** (his monthly meter sensor says 82.3%).
- The energy-balance cross-check agrees with the slope-ratio RTE (0.989 vs 0.991), but both are computed from the same resampled series, so the agreement is shared source, not corroboration. The cross-check is currently informational anyway: `identify()` records `rte_crosscheck_energy_balance` but nothing gates on it.

So the capacity estimate is now good (+5.2% vs his metered value, against +37.6% on master before #1063), but the RTE clears every gate at a value no real AC-coupled pack can have.

At `trust_tier: suggest` that fit produces a logged recommendation to set `battery_charge_efficiency` / `battery_discharge_efficiency` to ~0.995 each way, plus the advisory RTE sensor at 0.991. A user who follows that recommendation hands the optimizer an efficiency of ~0.99 where the true value is ~0.83, and it will over-cycle the battery because cycling looks nearly free.

## Two candidate mitigations, both proposed by @volschin in the #1063 thread (neither implemented)

**1. Tighten `SQRT_RTE_HIGH` from 0.999 to 0.995** (battery_identification.py:91). One constant, defensible on physics alone: no real AC-coupled pack is that close to lossless. It rejects volschin's case (sqrt(0.991) = 0.9955 > 0.995), and the cost on legitimate fits should be nil for AC coupling. The only case it could touch is a DC-side power sensor on a pack genuinely above 0.99 round trip, and I can't construct a plausible example of one. This is the one I'd do first.

**2. Make suspiciously close slope-ratio vs energy-balance agreement count against the fit rather than for it**, since two statistics drawn from the same throughput series agreeing is not independent confirmation. The honest complication, worked out in the PR thread: closeness is also what a healthy fit looks like. On dense clean data both estimators converge on the true RTE, so a bare closeness trigger would start rejecting exactly the fits we want. It needs conditioning on something more, for example closeness combined with proximity to the guardrail edge, before it can gate without false positives. More design work than option 1, so I'd treat it as a possible second step rather than the opener.

## Non-goals

The change-only sampling bias itself is the forward-fill follow-up already deferred in #1051 / #1063, and it stays separate. @volschin's side experiment there is worth keeping in view for whoever picks it up (a plain last-value hold fabricated energy through a different mechanism: RTE 1.76, TLS 8.01 kWh vs Theil-Sen 5.35 kWh, rejected), so any change-only fill needs explicit staleness handling with the dt bound from #1063 underneath it. This issue is only about the guardrail that sits in front of publishing.

History: #1051 (parent issue for the gap mechanism, closed by #1063), #1063 (merged, acceptance data and mitigation discussion in-thread).

Happy to build whichever shape you prefer. Option 1 is a one-line PR plus a test; option 2 I'd want to sketch in-thread first.

LesIT1:
Closes #1069. This is option 1 from that issue, the one greenlit as the first step.

## What changed

`SQRT_RTE_HIGH` goes from 0.999 to 0.995 in `battery_identification.py`. That's the upper bound on the one-way efficiency sqrt(RTE) a fit is allowed to publish with. No real AC-coupled pack sits above 0.995 one-way (RTE > 0.99), so anything up there is a sampling artifact, not physics.

The motivating case is @volschin's acceptance run on #1063: with the gap fix in place his 26-day fit passed every guardrail at RTE 0.991 (one-way 0.9955), against an independently metered cycle efficiency of 83.2% for the same pack. Under `trust_tier: suggest` that fit produces a recommendation to set both efficiency params to ~0.995, and an optimizer fed ~0.99 where the truth is ~0.83 will over-cycle the battery because cycling looks nearly free. With this change that fit lands `rejected_sanity_check` and the configured values stay in place.

One side tweak in the same block: the rejection message now prints the efficiency with 4 decimals instead of 3. At the new bound a rejected 0.9955 would otherwise print as "0.995 outside [0.8, 0.995]", which reads like a contradiction.

## Behaviour change, disclosed

Fits with one-way efficiency in (0.995, 0.999] flipped from publishable to rejected. That's the entire point of the issue, but it does mean a setup that was getting a (wrong) published RTE in that band will now keep its configured values and log the rejection. The only setup I could imagine legitimately living in that band is a DC-side power sensor on a genuinely high-RTE pack, and I couldn't construct a plausible example, same as in the issue discussion.

Everything else is untouched: the lower bound, the capacity sanity bands, the CI and divergence gates, and the energy-balance cross-check (still informational).

## Tests

- `test_near_lossless_rte_rejected`: volschin's exact numbers (RTE 0.991, one-way 0.9955) must land `rejected_sanity_check`, and the message assert pins the rejection to the efficiency band rather than the separate RTE > 1 gate. Fails on master, which publishes `ok` for that fit.
- `test_plausible_high_rte_still_passes`: RTE 0.988 (one-way 0.994), just below the new bound, must still publish. Pins the bound location from below.

Full suite green locally.

Option 2 from the issue (making suspiciously close slope-ratio vs energy-balance agreement count against the fit) stays open for a later step, per the discussion there.

## Summary by Sourcery

Tighten the battery identification efficiency guardrail to reject near-lossless round-trip efficiency fits and ensure unrealistic results do not get published.

Bug Fixes:
- Prevent publishing of battery fits with implausibly high one-way efficiency by lowering the upper sqrt(RTE) sanity bound.
- Clarify rejection logs by increasing the printed precision of the one-way efficiency value in sanity check messages.

Tests:
- Add regression tests ensuring near-lossless efficiency fits are rejected and plausible high-efficiency fits still pass under the new bound.

### DIFF
--- a/src/emhass/battery_identification.py
+++ b/src/emhass/battery_identification.py
@@ -88,7 +88,9 @@ CAPACITY_SANITY_HIGH = 1.5
 ABS_CAPACITY_LOW_WH = 200.0
 ABS_CAPACITY_HIGH_WH = 500000.0
 SQRT_RTE_LOW = 0.80  # one-way efficiency = sqrt(RTE) plausible range
-SQRT_RTE_HIGH = 0.999
+# No real AC-coupled pack is near-lossless: a one-way efficiency above 0.995
+# (RTE > 0.99) is a sampling artifact, not physics (#1069).
+SQRT_RTE_HIGH = 0.995
 
 # Energy-balance cross-check is only meaningful over an approximately CLOSED
 # window: if the net SoC drift is large relative to the total SoC swung, the
@@ -595,7 +597,7 @@ class BatteryIdentification:
         if not (SQRT_RTE_LOW <= eta <= SQRT_RTE_HIGH):
             result.status = "rejected_sanity_check"
             result.messages.append(
-                f"One-way efficiency sqrt(RTE)={eta:.3f} outside "
+                f"One-way efficiency sqrt(RTE)={eta:.4f} outside "
                 f"[{SQRT_RTE_LOW}, {SQRT_RTE_HIGH}]; keeping configured values."
             )
             return result
--- a/tests/test_battery_identification.py
+++ b/tests/test_battery_identification.py
@@ -182,6 +182,27 @@ class TestBatteryIdentification(unittest.TestCase):
         self.assertEqual(res.status, "rejected_sanity_check")
         self.assertFalse(res.is_ok)
 
+    def test_near_lossless_rte_rejected(self):
+        # The #1069 case: change-only sampling produced a fit with RTE 0.991
+        # (one-way sqrt 0.9955) on a pack independently metered at 0.83
+        # round-trip efficiency. No real AC-coupled pack is that close to
+        # lossless, so the efficiency guardrail must reject it rather than
+        # publish. Asserting on the message pins the rejection to the
+        # efficiency band, not the separate RTE > 1 gate.
+        df = _make_history(10000.0, 0.991, n_cycles=6)
+        res = self._identify(df)
+        self.assertEqual(res.status, "rejected_sanity_check", msg=str(res.messages))
+        self.assertFalse(res.is_ok)
+        self.assertTrue(any("sqrt(RTE)" in m for m in res.messages), msg=str(res.messages))
+
+    def test_plausible_high_rte_still_passes(self):
+        # Counterfactual pin for the bound location: RTE 0.988 (one-way sqrt
+        # 0.994) sits just below SQRT_RTE_HIGH and must still publish.
+        df = _make_history(10000.0, 0.988, n_cycles=6)
+        res = self._identify(df)
+        self.assertEqual(res.status, "ok", msg=str(res.messages))
+        self.assertTrue(res.is_ok)
+
     # -- default-no-op sanity: to_dict is JSON-serialisable scalars -----------
     def test_result_to_dict_is_scalar_json(self):
         import json


LesIT1:
Thanks Sourcery. Passing on both, with reasons.

On making the bound configurable or topology-tied: the constants block this sits in is deliberately not user configuration. The header comment there spells it out, these are safety rails and a user shouldn't be able to widen them from config. That was a design decision when the feature first landed, and it matches the project's general preference for internal constants over new params. Topology isn't modelled anywhere in the identification feature either, so a topology-keyed bound would be new machinery for a case that so far only exists hypothetically. If someone shows up with a DC-side sensor on a genuinely >0.99 RTE pack, that's the moment to revisit, and the discussion in #1069 already flags exactly that case.

On the comment strength: the #1069 reference is doing the work there. It holds the empirical case (a 0.9955 one-way fit on a pack independently metered at 83.2%) and the reasoning, so a future maintainer landing on that line has the full rationale one click away. The claim is also scoped to AC-coupled packs on purpose, which is the softening you're after.

davidusb-geek (approved):

PR #1071, opened 2026-08-11
fix: size the mlforecaster load forecast history retrieval from the trained model's lags (#1067)

### LINKED ISSUE #1067: mlforecaster: naive-mpc-optim fails with skforecast last_window ValueError when num_lags is large, regardless of retrieval backend or historic_days_to_retrieve
**EMHASS version: 0.18.0** (add-on, updated within the last few days) Installation: Home Assistant OS add-on (5b918bf2_emhass) 
**Load forecast method**: mlforecaster, sklearn_model: KNeighborsRegressor 
**Data retrieval**: InfluxDB (for fit/tune), REST API (for the live naive-mpc-optim loop)

**Summary**

After forecast-model-tune (Bayesian/Optuna hyperparameter search) selected num_lags: 144 as optimal for our load data, every subsequent naive-mpc-optim call (triggered every 5 minutes by a Home Assistant automation) fails with:
ERROR in machine_learning_forecaster: Error during prediction: `last_window` must have as many values as needed to generate the predictors. For this forecaster it is 144.
ERROR in app: Exception on request POST /action/naive-mpc-optim
  File "/app/.venv/lib/python3.12/site-packages/quart/app.py", line 1465, in handle_request
    return await self.full_dispatch_request(request_context)
  ...
  File "/app/src/emhass/web_server.py", line 794, in action_call
    input_data_dict = await set_input_data_dict(
  File "/app/src/emhass/command_line.py", line 2090, in set_input_data_dict
    result = await _prepare_naive_mpc_optim(ctx, stage_times=stage_times)
  File "/app/src/emhass/command_line.py", line 1802, in _prepare_naive_mpc_optim
    p_load_forecast = await ctx.fcst.get_load_forecast(
  File "/app/src/emhass/forecast.py", line 2115, in get_load_forecast
    forecast_out = await self._get_load_forecast_ml(df, use_last_window, mlf, debug)
  File "/app/src/emhass/forecast.py", line 2004, in _get_load_forecast_ml
    forecast_out = await mlf.predict(data_last_window, weather_future=weather_future)
  File "/app/src/emhass/machine_learning_forecaster.py", line 456, in predict
    predictions = await asyncio.to_thread(
  ...
  File "/app/.venv/lib/python3.12/site-packages/skforecast/recursive/_forecaster_recursive.py", line 1222, in _create_predict_inputs
    check_predict_input(
  File "/app/.venv/lib/python3.12/site-packages/skforecast/utils/utils.py", line 984, in check_predict_input
    raise ValueError(
ValueError: `last_window` must have as many values as needed to generate the predictors. For this forecaster it is 144.

The model itself is genuinely good — a full backtest (forecast-model-fit with perform_backtest: true, 445 days of InfluxDB history) confirms num_lags: 144 clearly outperforms num_lags: 48 on every metric:
___________________________________________
**num_lags	R²	MAE	RMSE	MAPE**
48	0.326	218.3 W	323.8 W	33.0%
144	**0.478	188.3 W	284.8 W	28.7%**
___________________________________________

So this isn't a bad hyperparameter choice by Optuna — the tuned model is simply unusable in the live/real-time prediction path.

**What I ruled out**
**Not a data-availability problem.** 
Queried the Home Assistant recorder SQLite DB directly for the underlying sensor:

SELECT datetime(MIN(last_updated_ts),'unixepoch'), datetime(MAX(last_updated_ts),'unixepoch'), COUNT(*)
  FROM states WHERE metadata_id = <id>;
  -- oldest: 2026-07-30 02:12:23, newest: 2026-08-09 20:13:00, rows: 31913

That's ~10 days of continuous history (default purge_keep_days), far more than the 3 days (144 × 30 min) the model needs.
**Not InfluxDB-specific.** Passing "use_influxdb": true as a runtime parameter directly to naive-mpc-optim did not change the outcome — same error. The startup log shows:
_INFO in web_server: WebSocket and InfluxDB modes disabled - using REST API for data retrieval_

This appears to be decided once at process startup from the permanent config, not overridable per-request for this action — unlike forecast-model-fit / forecast-model-tune / forecast-calibration, where the same use_influxdb runtime parameter is honored correctly and logged as InfluxDB integration enabled: ....
**Not fixable via config** num_lags. Setting the permanent config.json num_lags to 144 (matching the saved model) and restarting the add-on made no difference — identical error.

**Reproduction steps**
1. Set load_forecast_method: mlforecaster, sklearn_model: KNeighborsRegressor.
2. Run forecast-model-tune with use_influxdb: true and a large historic_days_to_retrieve (in our case ~445 days), without pinning num_lags — let Optuna search freely. It selects a large lag count (144 in our case).
3. Let the normal naive-mpc-optim automation run (every 5 minutes in our setup, via REST API, no InfluxDB).
4. Observe the ValueError above on every run.
5. Confirm the model is otherwise valid: forecast-model-fit with perform_backtest: true and the same num_lags completes successfully and produces good backtest metrics.

**Related issue**
#726 reports a similar-looking symptom (Unable to obtain: N lags_opt values from sensor...) using WebSocket retrieval rather than InfluxDB/REST, and also found that raising historic_days_to_retrieve (to 2–3 days) for the live call did not resolve it. That issue uses a different, EMHASS-wrapped error message than the raw skforecast ValueError we see, so it's possible EMHASS catches this condition in some code paths but not others (or the wrapping has regressed in 0.18.0).

**Question / request**

Is the live naive-mpc-optim / MPC prediction path expected to size its last_window fetch based on the loaded model's actual num_lags, or does it use a fixed/independent window? If the latter, could this be made dynamic (read the required lag count from the loaded model object), or at minimum documented as a constraint so users know not to let forecast-model-tune search unboundedly over num_lags when mlforecaster is used for live/frequent predictions?

Happy to provide the full add-on log output or config.json (redacted) if useful.

LesIT1:
Fixes #1067.

## The bug

When `load_forecast_method: mlforecaster` runs inside an optimization (`naive-mpc-optim` or `dayahead-optim`), the load forecast fetches its own history window in `Forecast._prepare_hass_load_data`, and both callers size that window as `delta_forecast_daily.days` (default 1). The trained model is only unpickled after the fetch, so the retrieval knows nothing about how many past samples the auto-regressive model actually needs for its skforecast `last_window`.

With a tuned model at `lags_opt: 144` and a 30 min time step, the model needs 144 samples (3 days) but the fetch returns one day plus part of today, so every optimization run dies inside skforecast:

```
ValueError: `last_window` must have as many values as needed to generate the predictors. For this forecaster it is 144.
```

No config knob can fix it from the outside: `historic_days_to_retrieve` is not used on this path, and tune is free to pick lags (up to 72h of history at the configured step) that the live path can never feed. This is not a 0.18.0 regression. The sizing has been that way for years.

## The fix

All inside `forecast.py`, no new config parameters:

- `get_load_forecast` now loads the mlf pickle before the history retrieval (new `_load_mlf_model` helper, same missing-file error as before) and, when `use_last_window` is set, enlarges the retrieval to `max(days_min_load_forecast, ceil(window_size / samples_per_day) + 1)`. The extra day covers DST-short days and small recorder gaps.
- The window size comes from `forecaster.window_size` (authoritative in skforecast: it equals the largest lag even for non-contiguous lags), falling back to `lags_opt` for tuned models and `num_lags` otherwise.
- `_get_load_forecast_ml` now checks the retrieved history against the model's window size before calling predict. If the recorder holds less than the model needs, the run aborts with a clear error naming the numbers and the sensor instead of a raw skforecast traceback.

Verified against skforecast 0.19.1 that a longer `last_window` produces predictions identical to an exact-size window (only the tail is used), so runs that already had enough history are unaffected.

## Behaviour changes

- The mlforecaster fetch can now retrieve more days than `delta_forecast_daily` (that is the fix). Recorder queries on this path get proportionally longer for large-lag models. This also applies to the default untuned config: `num_lags: 48` at 30 min with `delta_forecast_daily: 1` now fetches 2 days instead of 1, which is what makes runs just after midnight (almost no samples from today yet) reliable.
- A missing model pickle now fails before the history fetch instead of after it (same error message, cheaper failure).
- Insufficient recorder history now returns a logged error and aborts the optimization cleanly instead of raising the skforecast ValueError.

The `naive`, `typical`, `csv` and `list` methods are untouched, as is the `forecast-model-predict` endpoint (that path is sized by `historic_days_to_retrieve`, which the user controls).

## Tests

New `TestMlforecasterLastWindowSizing` in `tests/test_forecast.py`, running a real fitted `MLForecaster` against a fake recorder whose returned history is a deterministic function of the days requested:

- a 144-lag model with `delta_forecast_daily: 1` now forecasts (fails on master with the exact `last_window` ValueError above, proven red on base)
- a recorder retaining less than the model needs aborts with False, not a raw ValueError
- a small model keeps `days_min_load_forecast` as the floor (no fetch enlargement)
- a missing pickle fails before any retrieval happens
- unit coverage for the window-size source order and the sizing math at 30 min and 1 h steps

Full suite: 987 passed / 32 xfailed locally (980 baseline + 7 new).

Docs: added a note to `docs/mlforecaster.md` about the recorder retention this path needs.

@Lumme78 you mentioned you could test on your InfluxDB setup with the 450 days of history. That would be a great real-world check of this exact path if you're still up for it: check out this branch, keep your tuned `num_lags: 144` model, and run `naive-mpc-optim` with your normal `delta_forecast_daily`. It should now retrieve 4 days and produce a forecast.

## Summary by Sourcery

Align mlforecaster load forecast history retrieval with the trained model’s lag requirements to avoid skforecast last_window errors during optimizations.

New Features:
- Automatically size mlforecaster load history retrieval based on the model’s effective window size and the optimization time step when use_last_window is enabled.

Bug Fixes:
- Prevent mlforecaster-based optimizations from failing with skforecast last_window ValueError by ensuring sufficient historical samples are retrieved for large-lag models.
- Ensure missing ML forecaster pickle files cause an immediate, clean failure before any recorder access.
- Handle cases where the recorder retains too little history by aborting the forecast cleanly with a clear error message instead of propagating skforecast exceptions.

Enhancements:
- Introduce helpers to derive the ML forecaster’s required window size and corresponding history duration from the fitted model and configuration.
- Add validation that the retrieved history length meets the model’s last-window requirement before running predictions.
- Clarify in get_load_forecast docs that days_min_load_forecast is a lower bound for mlforecaster and may be increased automatically to satisfy model lag needs.

Documentation:
- Document recorder retention requirements for using mlforecaster within optimization runs, including guidance on necessary history duration for large-lag models.

Tests:
- Add TestMlforecasterLastWindowSizing suite to validate history sizing logic, model window-size derivation, clean failure modes, and edge cases across different lag counts and time steps.

### DIFF
--- a/docs/mlforecaster.md
+++ b/docs/mlforecaster.md
@@ -245,6 +245,11 @@ The hyperparameter tuning using Bayesian optimization improves the bare KNN regr
 The tuning routine can be computing intense. If you have problems with computation times, try to reduce the `historic_days_to_retrieve` parameter. In the example shown, for a 240-day train period, the optimization routine took almost 20 min to finish on an amd64 Linux architecture machine with an i5 processor and 8 GB of RAM. This is a task that should be performed once in a while, for example, every week.
 ```
 
+```{note}
+
+When the model is used inside an optimization run (`load_forecast_method: mlforecaster` with `dayahead-optim` or `naive-mpc-optim`), EMHASS automatically retrieves enough recent load history to fill the model's last window: at least as many past samples as the model has lags. For example, a tuned model with 144 lags at a 30 min time step needs 3 days of history. Make sure your Home Assistant recorder (`purge_keep_days`) retains at least that much history for the load sensor, otherwise the forecast will abort with an explicit error.
+```
+
 ## How does this work? 
 This machine learning forecast class is based on the `skforecast` module. 
 We use the recursive autoregressive forecaster with added features. 
--- a/src/emhass/forecast.py
+++ b/src/emhass/forecast.py
@@ -20,6 +20,7 @@ except ImportError:  # Windows
 
 
 import logging
+import math
 import os
 import pickle
 import pickle as cPickle
@@ -1986,27 +1987,79 @@ class Forecast:
         )
         return await self.get_weather_covariates(future_index, weather_features)
 
-    async def _get_load_forecast_ml(
-        self, df: pd.DataFrame, use_last_window: bool, mlf, debug: bool
-    ) -> pd.DataFrame | bool:
-        """Helper for ML forecast."""
+    async def _load_mlf_model(self, mlf, debug: bool) -> MLForecaster | bool:
+        """Resolve the MLForecaster instance to use for the load forecast.
+
+        In production (``debug=False``) the model is read from the saved pickle
+        file; in debug/unit-test mode the passed ``mlf`` object is used as-is.
+        Returns False when the pickle file is missing.
+        """
+        if debug:
+            return mlf
         model_type = self.params["passed_data"]["model_type"]
         filename = model_type + "_mlf.pkl"
         filename_path = self.emhass_conf["data_path"] / filename
-        if not debug:
-            if filename_path.is_file():
-                async with aiofiles.open(filename_path, "rb") as inp:
-                    content = await inp.read()
-                    mlf = pickle.loads(content)
-            else:
-                self.logger.error(
-                    "The ML forecaster file was not found, please run a model fit method before this predict method"
-                )
-                return False
+        if not filename_path.is_file():
+            self.logger.error(
+                "The ML forecaster file was not found, please run a model fit method before this predict method"
+            )
+            return False
+        async with aiofiles.open(filename_path, "rb") as inp:
+            content = await inp.read()
+            return pickle.loads(content)
+
+    @staticmethod
+    def _mlf_window_size(mlf) -> int:
+        """Return the number of observed samples the fitted model needs as its last window.
+
+        skforecast's ``forecaster.window_size`` is authoritative: it equals the
+        largest lag, which can exceed ``len(lags)`` when the lags are not
+        contiguous. Fall back to ``lags_opt``/``num_lags`` for model objects
+        without a fitted forecaster attached.
+        """
+        window_size = getattr(getattr(mlf, "forecaster", None), "window_size", None)
+        if window_size is None and getattr(mlf, "is_tuned", False):
+            window_size = getattr(mlf, "lags_opt", None)
+        if window_size is None:
+            window_size = mlf.num_lags
+        return int(window_size)
+
+    def _mlf_required_history_days(self, mlf, days_min_load_forecast: int) -> int:
+        """Days of history to retrieve so the model's last window can be filled.
+
+        Sized from the model's real lag requirement at the optimization time
+        step, never below ``days_min_load_forecast``. One extra day covers
+        DST-short days and minor recorder gaps.
+        """
+        window_size = self._mlf_window_size(mlf)
+        samples_per_day = max(1, int(pd.Timedelta(days=1) / self.freq))
+        needed_days = math.ceil(window_size / samples_per_day) + 1
+        return max(days_min_load_forecast, needed_days)
+
+    async def _get_load_forecast_ml(
+        self, df: pd.DataFrame, use_last_window: bool, mlf, debug: bool
+    ) -> pd.DataFrame | bool:
+        """Helper for ML forecast.
+
+        ``mlf`` is the already-loaded model: ``get_load_forecast`` reads the
+        pickle before the history retrieval (see ``_load_mlf_model``) so the
+        retrieval window can be sized from the model's lag requirement.
+        """
         data_last_window = None
         if use_last_window:
             data_last_window = copy.deepcopy(df)
             data_last_window = data_last_window.rename(columns={self.var_load_new: self.var_load})
+            window_size = self._mlf_window_size(mlf)
+            if len(data_last_window) < window_size:
+                self.logger.error(
+                    "Not enough load history to fill the ML forecaster's last window: "
+                    f"the model needs {window_size} samples at {self.freq} "
+                    f"but only {len(data_last_window)} were retrieved. Check that the "
+                    "Home Assistant recorder (or database) retains at least "
+                    f"{self._mlf_required_history_days(mlf, 0)} days of history for "
+                    f"{self.retrieve_hass_conf['sensor_power_load_no_var_loads']}."
+                )
+                return False
         # When the model was trained with weather covariates, supply the future weather over the
         # forecast horizon so the recursive predict has the exog columns it expects.
         weather_future = await self._build_weather_future(data_last_window, mlf)
@@ -2077,7 +2130,10 @@ class Forecast:
         Get and generate the load forecast data.
 
         :param days_min_load_forecast: The number of last days to retrieve that \
-            will be used to generate a naive forecast, defaults to 3
+            will be used to generate a naive forecast, defaults to 3. For the \
+            'mlforecaster' method this is a lower bound: the retrieval window is \
+            automatically enlarged to cover the trained model's last-window \
+            requirement (its lags) when needed.
         :type days_min_load_forecast: int, optional
         :param method: The method to be used to generate load forecast, the options \
             are 'typical' for a typical household load consumption curve, \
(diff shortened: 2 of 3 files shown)


LesIT1:
Thanks for the review. Not taking these four, reasoning below.

**Fixed timestamp in the fake recorder**: the sample count is already fully deterministic. The fake builds its index with `periods=n_days * samples_per_day`, not from a time span, so the wall clock never changes how many rows come back. The only real clock dependence in those tests is `Forecast.__init__` reading now() for `forecast_dates`, which a fixed anchor inside the fake would not remove. The one place that mattered (a 50-step horizon on the 25-hour day before a fall-back DST transition) is already handled by sizing the small-model test to 52 lags, with a comment saying exactly why.

**ceil on `samples_per_day`**: this one points the wrong way. Flooring `samples_per_day` can only overestimate `needed_days`, which costs at most one extra fetched day. Switching to ceil can only underestimate, which eats into the +1 buffer that exists for DST-short days and recorder gaps. Over-fetching is cheap and under-fetching is the bug this PR fixes, so floor is intentional. I checked the sizing at 7, 15, 30, 60, 61 and 90 minute steps during the build.

**"computing intense"**: agreed it reads oddly, but that sentence predates this PR and my diff never touches it. I'd rather keep this one scoped to #1067; happy to fix it in a follow-up docs pass.

**Centralizing window size and validation into one helper**: the split is load-bearing. The sizing has to run before the retrieval, when no data exists yet (that is the whole bug), and the length check can only run after it, so a single helper taking `data_last_window` cannot replace both. The getattr chain in `_mlf_window_size` is there for pickles saved by older EMHASS versions that predate `is_tuned`/`lags_opt`, and a raise-and-catch flow would cut against this method's existing return-False error idiom (`df is False`, `forecast_out is False`). Moving the policy onto `MLForecaster` as a property would also work, but it widens a targeted fix; I can do that in a follow-up if David prefers the logic to live there.

Lumme78:
Tested this against our real setup as offered — great news, it works.

What I did: patched only forecast.py from this branch into our running EMHASS container (no full rebuild), keeping everything else on the released 0.18.0 image; re-ran forecast-model-fit with num_lags: 144, use_influxdb: true, historic_days_to_retrieve: 448 to restore the tuned model; then let our normal 5-minute automation fire naive-mpc-optim with its full production payload (InfluxDB-independent, REST-based, delta_forecast_daily: 1 as before).

Result: complete success, no last_window ValueError. Log excerpt from the automated run:

```
INFO in forecast: Retrieving data from hass for load forecast using method = mlforecaster
INFO in retrieve_hass: Using REST API for data retrieval
...
INFO in optimization: Total value of the Cost function = 20.31
INFO in command_line: Publishing data to HASS instance
INFO in command_line: Optimization completed in 3.8s
```

I also ran a manual naive-mpc-optim call with a minimal payload (just prediction_horizon) to poke at it further — that one failed at the LP stage with 'infeasible', but only because I hadn't supplied load_cost_forecast/prod_price_forecast/etc. The forecast step itself succeeded there too, so the fix held even under an intentionally incomplete request.

Afterward I restored the original forecast.py and refit back to num_lags: 48, so we're back on the stable released code while this is under review — didn't want to run unmerged code in production long-term.

This is a solid real-world confirmation on top of your own test suite. Thanks again for digging into this and building the fix so quickly — let us know if there's anything else useful we can check.

LesIT1:
That's exactly the confirmation this PR needed, thanks for setting it up so carefully. Patching just forecast.py onto the released image is the right isolation for testing this change, and refitting to 144 lags first made it a true repro of the original failure. Good to see your production automation go straight through with delta_forecast_daily: 1 unchanged.

On the minimal manual call: the infeasible comes from the solver stage, which this PR doesn't touch, so the forecast succeeding there is the part this fix owns. If the infeasible persists with your full payload on released code, that would be worth its own issue, but from your description it was just the stripped-down request.

Sensible to drop back to the released code while it's under review. Nothing else I need checked, this plus the test suite covers it. Over to David now.

davidusb-geek (approved):

PR #1083, opened 2026-08-18
fix: build the csv/list forecast grid from the frozen start instead of re-reading the clock (#1076)

### LINKED ISSUE #1076: Intermittent KeyError in csv/list forecast paths when a run crosses an optimization_time_step boundary
**Describe the bug**

Reported by @RikBast in https://github.com/davidusb-geek/emhass/issues/1047#issuecomment-5308284502 (EMHASS 0.18.0, 15 min steps, Europe/Brussels): a `naive-mpc-optim` run at 17:45:09 crashed with

```
KeyError: "[Timestamp('2026-08-16 17:30:00+0200', tz='Europe/Brussels')] not in index"
```

in `_extract_daily_forecast`, called from `get_load_cost_forecast` on the `list` path. The next run a minute later worked (well, it went infeasible, but that turned out to be a separate problem with his deferrable window, handled back in #1047).

The root cause is that the forecast date grid is read from the wall clock twice, at two different moments:

1. `Forecast.__init__` freezes `self.start_forecast` / `self.forecast_dates` from `pd.Timestamp.now()` when the input data dict is set up ([forecast.py L218-L270](https://github.com/davidusb-geek/emhass/blob/f380180b/src/emhass/forecast.py#L218-L270)). The MPC input DataFrame `df_input_data_dayahead` ends up indexed on these dates.
2. `get_load_cost_forecast(method="list")` then builds a second grid via `get_forecast_days_csv`, which calls `pd.Timestamp.now()` again ([forecast.py L1541-L1567](https://github.com/davidusb-geek/emhass/blob/f380180b/src/emhass/forecast.py#L1541-L1567)) and indexes the passed list on it.

Data prep between those two reads takes seconds (HA history retrieval, weather forecast). If a rounding boundary falls in that gap, the two grids disagree by one step. The one-step skew and the timing point at `method_ts_round: "first"` on the reporter's setup: an init at 17:44:5x floors to 17:30 while the second read at 17:45:09 floors to 17:45 (the shipped default is "nearest", where the same race exists but the boundary sits at the half-interval point instead, and "last" would name the later stamp as missing). `_extract_daily_forecast` then does a strict lookup of the input frame's day stamps against the list frame ([forecast.py L1653](https://github.com/davidusb-geek/emhass/blob/f380180b/src/emhass/forecast.py#L1653)), and the input frame's first stamp (17:30) is not in the list frame's index (which starts at 17:45). That is exactly why the reported error names a single missing timestamp, one step before the crash time.

So this is an intermittent race: it only bites when the request straddles a rounding boundary, which matches the reporter seeing it sporadically "at the 15 min step". Shorter timesteps and slower data prep make it more likely. It is not a 0.18.0 regression, the same strict lookup is present in v0.17.0. All four `get_forecast_days_csv` consumers re-read the clock, but the strict lookup that crashes is specific to the list variants (load cost and production price); the csv variants can only end up with a silently misaligned merge, and only when the CSV has no usable timestamp column. I verified the load-cost list path.

**To Reproduce**

Deterministic repro on current master (f380180b), pinning `pd.Timestamp.now` on each side of a boundary. Raises the exact reported KeyError as written; re-pin both calls to the same step and the forecast comes back clean.

```python
TZ = "Europe/Brussels"
HORIZON = 10

def _pin_now(ts_str):
    fixed_utc = pd.Timestamp(ts_str, tz=TZ).tz_convert("UTC")
    def fake_now(tz=None):
        if tz is None:
            return fixed_utc.tz_localize(None)
        return fixed_utc.tz_convert(tz)
    return mock.patch.object(pd.Timestamp, "now", staticmethod(fake_now))

# config: optimization_time_step 15, method_ts_round "first", time_zone Europe/Brussels
# params["passed_data"]["prediction_horizon"] = HORIZON
# params["passed_data"]["load_cost_forecast"] = [0.25] * 200

with _pin_now("2026-08-16 17:44:55"):   # set-input-data side: floors to 17:30
    fcst = Forecast(retrieve_hass_conf, optim_conf, plant_conf, params_json,
                    emhass_conf, logger, get_data_from_file=True)

# what naive_mpc_optim passes: input data indexed on the frozen dates
df = pd.DataFrame({"p_pv_forecast": 0.0, "p_load_forecast": 500.0},
                  index=fcst.forecast_dates[:HORIZON])

with _pin_now("2026-08-16 17:45:09"):   # optimization side: floors to 17:45
    fcst.get_load_cost_forecast(df, method="list")
# KeyError: "[Timestamp('2026-08-16 17:30:00+0200', tz='Europe/Brussels')] not in index"
```

**Expected behavior**

The csv/list forecast should align to the same date grid as the input data it is being merged into, regardless of when within the step the call happens to run.

**Proposed fix**

`get_forecast_days_csv` re-derives the start stamp from the clock with the same rounding logic `__init__` already ran. Reusing the frozen `self.start_forecast` instead of re-reading the clock keeps the csv/list grid consistent with `df_final` by construction, and consistency is what the strict lookup actually needs (if the frozen dates were stale, the input frame would be equally stale). Happy to put together a PR with a clock-pinned regression test along the lines of the repro above if you want it.

EMHASS 0.18.0 (reporter) and current master, Python 3.12, pandas 3.x.

LesIT1:
Fixes #1076.

## The bug

`Forecast.__init__` freezes `self.start_forecast` and `self.forecast_dates` from `pd.Timestamp.now()` when the input data dict is built. `get_forecast_days_csv` then read the clock a second time and re-applied the `method_ts_round` rounding to build its own grid. Data prep between those two reads (HA history retrieval, weather forecast) takes seconds, so a run that crosses a rounding boundary in that gap ends up with two grids one step apart. `_extract_daily_forecast` does a strict `df_csv.loc[fcst_index]` against the input frame, and the lookup raises

```
KeyError: "[Timestamp('2026-08-16 17:30:00+0200', tz='Europe/Brussels')] not in index"
```

naming the stamp one step behind the crash time. Two reporters hit it from different configs: @RikBast on 15 minute steps with `method_ts_round: "first"` (sporadic, boundary at the step edge), and @JoshC1994 on 30 minute steps with `nearest`, where round-half-to-even sends `:15:00` down and `:45:00` up so only the `:15` run has its two reads disagree. That gave him one crash an hour, always naming the top of the hour, 45 times in 45 hours.

It is not a 0.18.x regression. The same strict lookup and the same double clock read are in v0.17.0.

The crash is the visible half. The quieter half is that the same skew mis-prices the perfect-optim path without raising anything. See the disclosure list below.

## The fix

`get_forecast_days_csv` reuses the already-frozen `self.start_forecast` instead of reading the clock again. The rounding still happens, once, in `__init__`. Consistency with `df_final` is what the strict lookup actually needs, and this gives it by construction: if the frozen stamp were stale, the input frame would be equally stale, so the two cannot drift apart.

The removed block was a verbatim copy of the alignment that `__init__` already runs, so this deletes a duplicate rather than moving logic around.

- A run that straddles a boundary now returns a grid anchored one step earlier than the old builder produced, matching the frame it is being merged into. That is the fix, but it is a real change in what comes back rather than only the removal of an exception.
- Runs that do not straddle a boundary are unaffected. The two grids were already identical there.
- The deleted block carried its own `else: self.logger.error("Wrong method_ts_round passed parameter")`, which was unreachable. An invalid value leaves `self.start_forecast` unset in `__init__`, and the first use of it a few lines later raises `AttributeError`, so construction never gets far enough for this branch to run.
- The perfect-optim list path (`list_and_perfect=True`) slices with `between_time` rather than the strict `.loc`, so a straddling run never raised. It returned the right number of rows with the wrong values: on master the tail repeats, `[0.17, 0.18, 0.18]` where the list says `[0.17, 0.18, 0.19]`. That is silently mis-priced input to a perfect optimization, and it is fixed here too. CSV data with no usable timestamp column has the same exposure through a different branch of the same function, though nothing here tests that case.
- A CSV that carries its own timestamp column is unaffected either way: that path never consults the built grid.

## Alternative considered

Swapping the strict `.loc` for `.reindex(fcst_index)` was raised in the issue. It is worse than it looks: the NaN row survives the merge and reaches `self.param_load_cost.value` in `optimization.py`, where cvxpy rejects it with `ValueError: Parameter value must be real`. That trades a `KeyError` naming the forecast grid for a `ValueError` inside solver setup, so it relocates the failure rather than fixing it.

## Tests

New `TestForecastDaysCsvStartAlignment` in `tests/test_forecast.py`, pinning `pd.Timestamp.now` on either side of a boundary the same way the tie-second tests from #1064 do:

- `test_grid_start_is_the_frozen_start_not_the_clock` - the returned grid starts at `self.start_forecast` however late it is called.
- `test_boundary_straddle_does_not_raise` - RikBast's config and crash instant, asserting the full horizon comes back with no NaN. The constructor time is reconstructed: the KeyError pins it to somewhere in 17:30-17:45, his log does not record it.
- `test_boundary_straddle_does_not_raise_prod_price` - the production-price list path, which shares the builder and the strict lookup.
- `test_nearest_half_hour_tie_boundary` - JoshC1994's config, 30 minutes and `nearest` at `:15`. On master this raises naming the top of the hour, which is the signature from his logs.
- `test_perfect_optim_list_path_prices_are_not_shifted` - the silent case above, asserting the values rather than just that nothing raised.
- `test_same_step_call_is_unchanged` - counterfactual: both reads inside one step. Passes on master too, so the set cannot false-green.

On master three of those raise the reported `KeyError` and two fail on value assertions; the counterfactual passes on both.

The existing `_assert_longer_lists_forecast` helper also gains the two paths it had to leave out. Its comment said `get_forecast_days_csv` read wall-clock time and so could not be pinned, which was true and is the same root cause fixed here. With the frozen start it is pinned by `_pin_forecast_to_date` like everything else in that helper, so the four DST list tests now assert exact slot counts and no NaN for load cost and production price as well. All four fail on master.

Full suite on the branch: 995 passed, 32 xfailed.

## Summary by Sourcery

Use the initialization-time forecast start for all CSV and list forecast grids to keep derived forecasts aligned with their input data.

Bug Fixes:
- Prevent forecast CSV and list grids from drifting across clock-rounding boundaries by reusing the forecast start frozen during initialization.
- Correct silently shifted pricing values in perfect-optimization list forecasts.

Enhancements:
- Expand forecast coverage to validate load-cost and production-price list paths, boundary alignment, nearest-rounding behavior, and unchanged same-step calls.

Tests:
- Add regression tests covering boundary-straddling failures, forecast-grid alignment, production-price forecasts, and perfect-optimization pricing.
- Strengthen DST and longer-list forecast assertions to verify exact slot counts and absence of missing values for list-priced forecasts.

### DIFF
--- a/src/emhass/forecast.py
+++ b/src/emhass/forecast.py
@@ -1546,27 +1546,13 @@ class Forecast:
         :rtype: pd.date_range
 
         """
-        start_forecast_csv = pd.Timestamp.now(tz=self.time_zone).replace(microsecond=0)
-        if self.method_ts_round == "nearest":
-            # Same start alignment as __init__: round once in UTC so the built
-            # range needs no stamp-by-stamp round (see the comment there).
-            start_forecast_csv = (
-                pd.Timestamp.now(tz=self.time_zone)
-                .replace(microsecond=0)
-                .tz_convert("utc")
-                .round(self.freq)
-                .tz_convert(self.time_zone)
-            )
-        elif self.method_ts_round == "first":
-            start_forecast_csv = (
-                pd.Timestamp.now(tz=self.time_zone).replace(microsecond=0).floor(freq=self.freq)
-            )
-        elif self.method_ts_round == "last":
-            start_forecast_csv = (
-                pd.Timestamp.now(tz=self.time_zone).replace(microsecond=0).ceil(freq=self.freq)
-            )
-        else:
-            self.logger.error("Wrong method_ts_round passed parameter")
+        # Reuse the start stamp frozen in __init__ rather than reading the clock
+        # again. Data prep between the two calls takes seconds, so a re-read that
+        # lands the other side of a rounding boundary builds a grid one step off
+        # from the one df_final is indexed on, and the strict lookup in
+        # _extract_daily_forecast then raises KeyError (issue #1076). The
+        # rounding itself still happens exactly once, in __init__.
+        start_forecast_csv = self.start_forecast
         end_forecast_csv = (
             start_forecast_csv + pd.DateOffset(days=self.optim_conf["delta_forecast_daily"].days)
         ).replace(microsecond=0)
--- a/tests/test_forecast.py
+++ b/tests/test_forecast.py
@@ -1490,11 +1490,10 @@ class TestForecast(unittest.IsolatedAsyncioTestCase):
         ``_pin_forecast_to_date``, so exact slot counts are asserted.
 
         ``get_load_cost_forecast`` and ``get_prod_price_forecast`` with
-        method="list" both internally call ``get_forecast_days_csv()`` which
-        uses wall-clock time rather than ``self.start_forecast``.  Pinning
-        those would require mocking ``pd.Timestamp.now`` throughout; instead
-        we just verify they do not raise and produce a DataFrame with the
-        expected columns.
+        method="list" internally call ``get_forecast_days_csv()``.  That used to
+        read the wall clock, so neither could be pinned here and both were left
+        out; since #1076 it builds from ``self.start_forecast``, which
+        ``_pin_forecast_to_date`` sets, so their slot counts are asserted too.
         """
         p_pv_forecast = await fcst.get_weather_forecast(method="list")
         self.assertIsInstance(p_pv_forecast, pd.DataFrame)
@@ -1515,11 +1514,25 @@ class TestForecast(unittest.IsolatedAsyncioTestCase):
         self.assertEqual(p_load_forecast.values[0], 1)
         self.assertEqual(p_load_forecast.values[-1], expected_last)
 
-        # get_load_cost_forecast and get_prod_price_forecast with method="list"
-        # call get_forecast_days_csv() which always uses wall-clock time and
-        # therefore cannot be pinned to a fixed date without mocking
-        # pd.Timestamp.now globally.  Those code paths are not relevant to the
-        # DST list-trimming fix validated here, so they are omitted.
+        # The two list-priced paths, now that the pinned start reaches them.
+        df_input = pd.DataFrame(
+            {
+                "p_pv_forecast": p_pv_forecast.values[:, 0],
+                "p_load_forecast": p_load_forecast.values,
+            },
+            index=p_pv_forecast.index,
+        )
+        df_cost = fcst.get_load_cost_forecast(df_input, method="list")
+        self.assertEqual(len(df_cost), expected_last)
+        self.assertEqual(df_cost[fcst.var_load_cost].isna().sum(), 0)
+        self.assertEqual(df_cost[fcst.var_load_cost].values[0], 1)
+        self.assertEqual(df_cost[fcst.var_load_cost].values[-1], expected_last)
+
+        df_price = fcst.get_prod_price_forecast(df_cost, method="list")
+        self.assertEqual(len(df_price), expected_last)
+        self.assertEqual(df_price[fcst.var_prod_price].isna().sum(), 0)
+        self.assertEqual(df_price[fcst.var_prod_price].values[0], 1)
+        self.assertEqual(df_price[fcst.var_prod_price].values[-1], expected_last)
 
     async def test_get_forecasts_with_longer_lists_summer(self):
         """3-day list forecast in summer (no DST transition): exactly 3×48 slots."""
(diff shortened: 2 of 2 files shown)


LesIT1:
Thanks. Declining this one: the two helpers solve different problems and do not share a signature.

`TestForecastDatesTieAlignment._build_forecast` takes a UTC instant, pins the clock for the duration of construction only, and hands back a built `Forecast`. It is a one-shot construct-under-a-pin helper, because `__init__` is the only place that class needs the clock held.

`_pinned_now` takes a local wall-clock instant plus a zone and returns a reusable context manager. Both of those differences carry weight here. The local form keeps each test reading the same as the report it reproduces, so `17:44:55` and `08:15:44` appear as the reporters wrote them rather than converted to UTC. The context manager is needed because these tests pin the clock twice within one test, at two different instants: once for `__init__`, then again for the later call that exposes the skew. That second pin is the whole point of the bug being tested.

Folding them together would mean either taking UTC-only input, which loses that readability, or carrying both call shapes behind one signature. It would also mean editing `TestForecastDatesTieAlignment`, which this PR does not otherwise touch, so a one-line source fix would grow a diff into unrelated test code.

I have documented the distinction in `_pinned_now`'s docstring instead, so the next person reading them side by side does not have to work out why they differ.

davidusb-geek (approved):

PR #1085, opened 2026-08-18
fix: keep partial days in REST history retrieval for battery identification (#1061)

### LINKED ISSUE #1061: Multi-battery identification: one battery's missing data drops history days for other batteries too
Related to #1046, and to #1060 (the two compound each other).

**Summary**
With `number_of_batteries > 1`, when more than one battery needs a fresh identification fit, `_identify_battery_impl` batches every stale battery's power + SoC sensors (plus the load sensor) into one `retrieve_home_assistant_data` call. Inside `_get_data_rest_api`, if any sensor in that combined list has zero readings for a given day, the whole day is dropped for every sensor in the batch  not just the one that
actually had the gap.

Each battery's fit only ever reads its own two columns (`identify(df, power_col, soc_col, ...)`), so there's no real dependency between them. But because they're fetched together, my one sparsely-reporting battery (inv002, see issue #1060) can cost my healthy battery (inv001) usable days it would otherwise have had. I think this is a plausible reason my healthy battery landed right at the edge of the RTE sanity bound (1.017, just barely rejected) in one run, despite testing fine in isolation.

**Suggested fix**

Instead of aborting the whole day on the first missing var, keep the vars that did return data and only leave gaps for the one that didn't:

```
\`\`\`diff
--- a/src/emhass/retrieve_hass.py
+++ b/src/emhass/retrieve_hass.py
@@ -524,26 +524,30 @@
             for day_idx, day in enumerate(days_list):
-                df_day = pd.DataFrame()
-                skip_day = False
+                day_index = pd.date_range(
+                    start=day, periods=int(24 * 60 / (self.freq.total_seconds() / 60)),
+                    freq=self.freq, tz=day.tzinfo,
+                )
+                df_day = pd.DataFrame(index=day_index)
+                missing_vars_today = []
                 for i, var in enumerate(var_list):
                     data = results.get((day_idx, var))
                     if data is None:
-                        skip_day = True
-                        break
+                        missing_vars_today.append(var)
+                        continue
                     df_resampled = self._process_history_dataframe(
                         data, var, day,
                         is_first_day=(day_idx == 0),
                         is_last_day=(day_idx == len(days_list) - 1),
                     )
                     if df_resampled is None:
-                        skip_day = True
-                        break
-                    if i == 0:
-                        df_day = pd.DataFrame(index=df_resampled.index)
+                        missing_vars_today.append(var)
+                        continue
                     df_day = pd.concat([df_day, df_resampled], axis=1)
-                if skip_day:
-                    days_skipped += 1
-                    continue
+                if missing_vars_today:
+                    days_skipped += 1
                 self.df_final = pd.concat([self.df_final, df_day], axis=0)
\`\`\`
```

I'm less confident in this one, it's the shared retrieval path used by `perfect-optim`/`adjust_pv`/`naive-mpc-optim` too, so the day-index construction (timezone/DST handling especially) needs real test coverage that I haven't done. Posting it as a direction rather than a ready patch; happy to test against a build if useful. No extra network calls either way, the per-entity fetches already happen concurrently, this only changes how the results get merged.

LesIT1:
Fixes #1061.

## The problem

`_get_data_rest_api` retrieves history one day at a time for the whole variable
batch, and the day loop drops a day for every variable as soon as any one
variable has no data for it. Battery identification batches the load sensor plus
every stale battery's power and SoC sensors into a single `var_list`, so one
sparsely-reporting battery costs a healthy battery days of history it did have
data for. Each battery's fit only ever reads its own two columns, so the
coupling is purely an artifact of the shared fetch.

## The fix

A new internal keyword argument `keep_partial_days` on `get_data` and
`_get_data_rest_api`, default False. Off, the day loop behaves exactly as
today: any missing variable skips the day for everyone. On, each variable
keeps whatever it returned for the day, and the day is only skipped when every
variable came back empty. A variable missing on a kept day simply ends up as a
NaN stretch after the per-day frames are joined, the same as a recorder gap,
so the gap handling added in #1063 still sees it as a gap. No
synthetic day index is built anywhere: the kept indices are the real resampled
timestamps, which is what keeps DST days and a partial last day behaving as
they do now.

The flag is set in one place, `_retrieve_from_hass`, and only when
`set_type == "battery_id"`; every other set_type going through that same call
(perfect-optim, adjust_pv, naive-mpc, dayahead) gets the default. The other
five `get_data` call sites are untouched and keep today's semantics: the
curtailment history filter, the ML forecaster data prep in `command_line.py`
and in `forecast.py`, forecast calibration, and the InfluxDB CSV export. The
websocket and InfluxDB retrieval paths are untouched too; if a
websocket-configured setup falls back to REST, the flag rides along into the
fallback.

On naming: per the issue thread this rides as a keyword argument since only
`_retrieve_from_hass` ever sets it. Happy to rename the flag if you prefer
something else.

## Behaviour changes (disclosed)

- battery_id retrieval only: days that used to be dropped because one sensor
  in the batch was silent are now kept for the sensors that did report. A
  sensor with no data at all for the whole range no longer takes the entire
  retrieval down with it; its battery falls back to configured values through
  the existing missing-column guard, and the other batteries fit normally.
- One edge worth stating plainly: that includes the load sensor. If the load
  sensor is silent for the entire range, battery_id retrieval used to fail
  outright; now the batteries still fit, since the fits only read their own
  power and SoC columns. In that state `prepare_data` logs an error about the
  missing load column and returns False, and its return value is ignored at
  this call site, which is pre-existing for every set_type; the frame skips
  the usual load renaming and timezone conversion, which the identification
  path does not depend on. I left that ignored return alone here to keep this
  change scoped; happy to follow up separately if you want it checked.
- A new warning line summarises how many days were kept partial, next to the
  existing skipped-days warning.
- Everything else, including all non-battery_id retrieval, is unchanged; a
  test pins the default path against today's behaviour.

## Tests

[VERIFY counts before submit]
- test_get_data_rest_api_keep_partial_days_keeps_day: the core behaviour; on
  current master this fails because the partial day is dropped.
- test_get_data_rest_api_default_still_skips_partial_days: the default-off
  no-op proof, passes on master and on this branch.
- test_get_data_rest_api_keep_partial_days_all_missing_still_skipped: a day
  with every variable missing is still skipped.
- test_get_data_rest_api_keep_partial_days_partial_last_day_no_fabricated_tail:
  a partial last day keeps its real index, no fabricated tail rows.
- test_retrieve_from_hass_keep_partial_days_plumbing: the flag is set for
  battery_id and only battery_id (this one exercises the command_line call
  site but lives in tests/test_retrieve_hass.py next to the behaviour it
  gates; happy to move it if you'd rather).
- test_get_data_websocket_fallback_threads_keep_partial_days: the websocket
  fallback threads the flag.

Full suite: 995 passed / 32 xfailed locally on Windows.

@JoshC1994 the branch is up, so this is the ping I promised. Absolutely no
action needed and no reply expected. If at some point you happen to be running
your 3-battery setup against a build anyway, this is the branch.

## Summary by Sourcery

Preserve partial history days during battery identification so sparse sensors do not reduce the data available for other batteries.

Bug Fixes:
- Preserve history for sensors that report on partially populated days during battery identification, preventing sparse batteries from discarding data belonging to healthy batteries.

Enhancements:
- Keep the existing all-or-nothing day handling by default while allowing battery identification to retain partial-day data and report partial-day retention.
- Propagate partial-day retrieval through websocket-to-REST fallback and document the updated battery identification history behavior.

Documentation:
- Update battery identification documentation to describe shared retrieval behavior and partial-day handling.

Tests:
- Add coverage for partial-day retention, default behavior, fully missing days, partial final days, battery-identification plumbing, and websocket fallback propagation.


LesIT1:
Thanks, took the first one: `keep_partial_days` is now passed by keyword in both `_get_data_rest_api` calls, and the fallback test now asserts `call_args.kwargs` directly instead of poking at positional args.

Declining the second. The consumers that matter already key off actual columns rather than the list: `prepare_data` filters its fix-up lists with `if c in self.df_final.columns`, and the battery identification path checks `power_col not in df.columns` before every fit. All three retrieval paths (REST, websocket, InfluxDB) set `self.var_list` to the requested list, so filtering it down to materialised columns on just this path would make the attribute mean different things depending on how the data arrived, and a separate `requested_var_list` would be a new attribute with no consumer. The comment above the assignment documents the invariant instead, which keeps the diff on the bug it is fixing.

davidusb-geek (approved):

JoshC1994:
I have be running the commit. Initially expected my sparse batt2 sensor to produce visible partial-day retention warnings. However, the log shows that batt2 generally returned a small number of samples rather than no samples at all.  Looks like those days are being handled later as recorder gaps. To directly exercise PR #1085, I need a day where batt2 has no usable data while batt0 or batt1 does. I have disabled the batt2 intergration (not activley cycling atm anyway) and stopped enties from updating and will report back in 24hrs+ with the log. hopefully i can capture this for some real world evidence.  I do have faith in the fix though ha.

PR #1086, opened 2026-08-18
fix: persist failed battery identification attempts so the model-age backoff engages (#1052)

### LINKED ISSUE #1052: Failed battery identification fits are not persisted, so the model-age cadence gate never engages and every run re-queries the recorder

LesIT1:
Closes #1052.

## The bug

Only a successful battery identification fit is ever persisted. On a non-ok result (`insufficient_data`, `rejected_sanity_check`, `low_confidence`) both writers bail before touching the file, so the next run finds no record, the freshness gates report "outdated", and the full `historic_days_to_retrieve` recorder pull runs again. `battery_identification_model_max_age` never engages for a setup whose fits keep failing. @volschin measured a run going from 3.9s to 19.3s on a 30-minute naive-mpc loop with a 7-day lookback, and @JoshC1994 reported roughly 30s of recorder I/O per 5-minute cycle with two batteries rejecting the same way every time.

## The fix

Option B from the thread, the one you picked on 2026-08-18 ("Agree to sticking with B. It seems like the most robust option"): keep the retained ok record, add a `last_attempt` block alongside it, and have the freshness gates treat a fresh recorded failure the same way they treat a fresh fit.

`last_attempt` holds the failure `status`, an `attempted_at` timestamp and the result `messages`. A failed fit now does a read-modify-write: a retained `status == "ok"` payload keeps every field and gains `last_attempt`; with no usable prior fit, the file (or slot) becomes a failure-only payload of `schema_version` plus `last_attempt`, deliberately with no `status` key, so none of the places that gate on `status == "ok"` can mistake it for a result. A later successful fit rewrites the payload or slot wholesale, which clears `last_attempt` by itself.

N=1 keeps riding the file mtime. The failure write bumps it, so the existing `is_model_outdated` gate engages, and the cache-hit branch returns on a fresh well-formed `last_attempt` instead of falling through to a re-fit.

N>1 is per slot. `last_attempt` lives inside `batteries[str(k)]`, and `_battery_fit_is_stale` treats a fresh one as not-stale for that battery alone. It also records the `(power, soc)` pair the attempt was made against, so editing the sensor lists voids a backoff the same way it already voids a cached ok record. The not-stale branch of the loop now publishes only entries that are `status == "ok"` and bound to the currently resolved pair; anything else logs at info and publishes nothing.

Two guards against the backoff turning into a trap: `max_age <= 0` is tested before either freshness branch, so "attempt every run" can't be defeated by a fresh `last_attempt`, and a malformed `last_attempt` (not a dict, no status, unparsable timestamp) falls through to a re-fit, so a hand-edited or truncated file can't become a permanent no-fit state.

No new config parameters and no `schema_version` bump, the keys are additive. `docs/battery_identification.md` and the `battery_identification_model_max_age` entry in `docs/config.md` are updated to match.

## Behaviour changes

1. **Failed fits now write to the data path**, where they wrote nothing before. At N=1 that mtime bump is the mechanism rather than a side effect: it's what makes the cadence gate engage.

2. **A retained estimate keeps being served indefinitely while attempts fail.** On master the `suggest` sensors went quiet once the record aged past max_age and fits kept failing; now the last good estimate carries on being published. `fitted_at` is already published as a sensor attribute, so the estimate's real age stays visible, and the docs now say this outright and point at the log for the failures themselves.

3. **Fixing a config problem after a rejected fit no longer forces an immediate retry**, the backoff holds until max_age. The docs name the escape hatch: set `battery_identification_model_max_age` to 0 for one run, or delete `battery_identification.json`.

4. **Mixed-version data directories.** An older reader sharing the file changes behaviour only in the retained-ok case: it sees a fresh mtime and `status == "ok"`, so it serves the retained record, which may be older than its own max_age. On a failure-only payload an old reader still falls through and re-fits every run, no worse than today. I checked the v0.18.0 and v0.18.1 tags, both have the `status == "ok"` cache gate.

5. **Two existing tests changed deliberately.** `test_failed_fit_writes_nothing_and_leaves_existing_untouched` is renamed to `test_failed_fit_records_last_attempt_and_preserves_existing_record`, since the "writes nothing" half stopped being true; it still seeds the same marker record and still asserts the marker survives and nothing is published. `test_n2_partial_failure_independence` keeps its name but now asserts the failing battery is present as a failure-only slot instead of absent.

6. **Cross-schema transition guard.** If the file on disk has the other layout and still holds a retained ok record (a v2 container found on an N=1 run, or a flat v1 payload found on an N>1 run), a failed fit skips the write and retries next run, which is master's behaviour, rather than risk discarding an estimate it doesn't know how to retain. A leftover container whose slots are all failure-only isn't protected, because there's nothing to keep and skipping it would leave no self-clearing path.

## Tests

The new tests pin: the backoff engaging at N=1 with no prior fit; the backoff expiring and re-fitting once the mtime is aged past max_age; a retained ok record still being served across failed attempts; a success after failure clearing `last_attempt`; an N>1 failing battery gaining an in-slot record and dropping out of the next retrieval batch; a sensor-pair change on a failure-only entry forcing an immediate retry; the publish gate refusing to serve an old pair's capacity for a newly resolved pair; and both directions of the cross-schema guard.

Two are RED on base. I ran them against master to confirm they fail on behaviour, not on an import or fixture error:

- `test_n1_no_prior_fit_failure_writes_failure_only_payload_and_backs_off` gives `AssertionError: False is not true : a failed fit must record the attempt`
- `test_n2_failing_battery_backs_off_and_is_excluded_from_next_retrieval` gives `AssertionError: '1' not found in {...}` (only battery 0 is present on master)

Full suite on the final tree: `999 passed, 32 xfailed` (the f380180b baseline of 989 plus the 10 new tests here). CI has not run yet as I write this.

## One thing left open

N=1 freshness still comes from the file mtime and N>1 from the stored per-slot timestamp, the same split as on master. Unifying them would mean changing `is_model_outdated`, which the adjusted-PV pickle shares, so I kept it out of this PR. Happy to do it separately if you want the two paths on one mechanism.

## Summary by Sourcery

Persist failed battery identification attempts and use them to back off retries while preserving and safely serving valid retained estimates.

Bug Fixes:
- Persist failed battery identification attempts so model-age backoff prevents repeated history retrieval and fitting.
- Continue serving retained successful battery estimates while subsequent fitting attempts fail.

Enhancements:
- Apply failure backoff independently per battery, including sensor-pair validation and safe handling of malformed records.
- Preserve retained estimates across failures and protect compatible records during transitions between single- and multi-battery storage layouts.

Documentation:
- Update battery identification documentation to describe persisted failures, retry cadence, retained estimates, and how to force an immediate retry.

Tests:
- Add coverage for failure-only persistence, backoff expiry, retained estimates, per-battery behavior, sensor changes, publish safety, and cross-schema protection.


LesIT1:
Took the typo, "so that the pack retries" reads better, pushed. Leaving the other three as they are: "each X and each Y" takes a singular verb, and the singular is doing work there (each record is bound to its own sensor pair, not all of them to one pair). On splitting _battery_fit_is_stale: the sensor-match and last_attempt pieces are already module-level helpers shared with other call sites, so what's left reads top to bottom in the order the checks run, and splitting the rest would scatter the branch logging. The double read in the multi-battery failure path is deliberate: the raw read has to see a v1/foreign shape that the container loader normalises away by design, it's a tiny local file, and the path only runs after a failed fit. Happy to revisit any of these if you'd rather, David.

davidusb-geek (approved):

JoshC1994:
Not sure if you need or wanted the feedback but I am running this commit.

Run 1 - battery_identification_model_max_age: 24 : the backoff works exactly as intended. Every cycle now logs recorded a failed attempt and backs off, with logging showing age climbing consistently between runs (8.3h → 8.4h → 8.5h...), Runs have dropped from 30s/cycle to effectively 0.

Run 2 - battery_identification_model_max_age: 0: recorder history retrieval + refit ran every 5-minute cycle at 28–30s cost as expected. 

Thanks LesIT1 for the solid fix.

PR #1087, opened 2026-08-18
fix: derive the window warning from the sequence length for sequence deferrable loads (#1081)

### LINKED ISSUE #1081: Sequence deferrable loads: no warning when the allowed window is shorter than the sequence
## The gap

A sequence deferrable load has to be placed in one piece: the solver picks a single start position for the whole profile. So it needs at least `len(nominal_power_of_deferrable_loads[k])` usable timesteps inside its `[start_timesteps, end_timesteps)` window. Give it fewer and the problem is infeasible every run, and because the window is a hard constraint the whole optimization dies with it, battery and all other loads included.

Nothing warns about this. Worse, whether you get a warning at all is decided by a parameter that has no effect on sequence loads.

This came out of @RikBast's reports on #1047: a 5-step sequence `[2200, 2200, 2200, 200, 200]` at 15-minute steps with `start_timesteps 0` / `end_timesteps 4`, which is 4 usable steps and infeasible every time.

This also refines something I said on that thread. I described the warning as simply not firing for sequence loads. That holds only when the operating hours and timesteps happen to be unset, so it is not a property of sequence loads at all, it is a coincidence of whatever value was left sitting in a field that does nothing here. It cuts the other way too, as the table below shows.

## What actually happens on master (f380180b)

The window check at [optimization.py:3695-3708](https://github.com/davidusb-geek/emhass/blob/f380180b/src/emhass/optimization.py#L3695-L3708) builds `min_steps` from the operating hours or operating timesteps, then `validate_def_timewindow` warns when `(end - start) < min_steps` ([:5856](https://github.com/davidusb-geek/emhass/blob/f380180b/src/emhass/optimization.py#L5856)). `sequence_length` is already in scope in the same loop iteration (computed at [:3525](https://github.com/davidusb-geek/emhass/blob/f380180b/src/emhass/optimization.py#L3525)) and is never consulted.

Four runs against master with the config above, horizon 10 steps:

| operating hours / timesteps for that load | "Available timeframe is shorter" warning | solver |
|---|---|---|
| both unset or 0 | **never fires** | Infeasible |
| `operating_timesteps = 5` | fires | Infeasible |
| `operating_hours = 1.25` (= 5 steps) | fires | Infeasible |

The first row is the natural configuration, because for a sequence load those parameters do nothing. I checked that directly: with the window widened so the problem solves, `operating_hours = 0`, `operating_hours = 100`, `operating_timesteps = 100` and `operating_timesteps = 5` all produce a byte-identical schedule (`P_deferrable0 = [2200, 2200, 2200, 200, 200, 0, 0, 0, 0, 0]`). The energy-constraint block that consumes them is skipped for sequence loads at [:3633-3638](https://github.com/davidusb-geek/emhass/blob/f380180b/src/emhass/optimization.py#L3633-L3638). They reach nothing except the arithmetic in that warning.

So the user who configures nothing meaningless gets no warning, and the user who happens to have left an unrelated value lying around gets one, for the wrong reason.

Two adjacent things that look like they should cover this and do not:

- The sequence branch has its own warning at [:3532-3537](https://github.com/davidusb-geek/emhass/blob/f380180b/src/emhass/optimization.py#L3532-L3537), but it compares `sequence_length` against the whole horizon `n`, not against the per-load window. In the case above `y_len = 10 - 5 + 1 = 6`, comfortably positive, so it stays quiet. Confirmed silent in all four runs.
- `validate_def_timewindow` is also called during the per-tick parameter update at [:4667](https://github.com/davidusb-geek/emhass/blob/f380180b/src/emhass/optimization.py#L4667) and [:4674](https://github.com/davidusb-geek/emhass/blob/f380180b/src/emhass/optimization.py#L4674), but the warning is discarded into `_` there. The only paths that log it are the problem build and the infeasible-retry rebuild, so even when it does fire it arrives as a byproduct of the failure rather than ahead of it.

## Proposed fix

Give the check the quantity that actually governs feasibility for a sequence load. In `_add_deferrable_load_constraints`, where `is_sequence_load` and `sequence_length` are both already in scope:

```python
if is_sequence_load:
    min_steps = sequence_length
elif def_total_timestep and def_total_timestep[k] > 0:
    min_steps = ceil(def_total_timestep[k])
else:
    min_steps = ceil(def_total_hours[k] / self.time_step)
```

For a sequence load `sequence_length` is the whole story, so the operating hours and timesteps should drop out of the check entirely rather than being folded in with a `max()`. I tried the `max()` version first and it keeps the false positive I described above: a window of exactly 5 steps for a 5-step sequence solves `Optimal`, but with `operating_hours = 10` left set it would still warn, because 40 steps is larger than 5. Those parameters are meaningless to the solve, so they should not be meaningless-except-when-larger to the warning.

That one function backs both logging call sites (the build at :5230 and the infeasible-retry at :5429), so it covers first runs and warm-started ones without touching the discarded calls in the parameter-update pass.

Two things I would want your opinion on before building:

1. The current warning text says "shorter than the specified number of hours to operate", which would be misleading for a sequence load. I would give the sequence case its own wording naming the sequence length and the usable window, since that is the number the user has to act on.
2. Whether this should stay a warning. It is a guaranteed infeasibility, known before the solve, and today it takes the whole optimization down with it. An early, explicit message naming the offending load is a large improvement over a bare `Infeasible`, but I have not proposed changing the failure behaviour itself here, only making it legible.

One small thing I noticed while in there, not introduced by this and not really part of it: `def_total_hours[k]` is read unguarded at :3706, while `def_start_timestep` and `def_end_timestep` get explicit `None` normalisation a few lines further on at :4557-4561, with a comment about partial `set-config` calls corrupting `params.pkl` into shapes like `[None, 0]`. `pad_list` only fills missing trailing entries, so an interior `None` in the hours list would raise a `TypeError` here. Happy to fold a defensive line in while touching this path, or leave it alone if you would rather keep the change narrow.

Out of scope: the sequence-continuation question from #1047 (a sequence is re-placed from scratch each run, with no notion of being partway through), and the exclusive-vs-inclusive `end_timesteps` documentation wording. Both are separate.

Happy to build this with tests if the shape works for you.

LesIT1:
Fixes #1081, agreed there before building.

## The bug

A sequence deferrable load is placed in one piece, so it needs at least `len(sequence)` usable timesteps inside its `[start_timesteps, end_timesteps)` window. The pre-solve check in `_add_deferrable_load_constraints` instead derived `min_steps` from the operating hours or operating timesteps, which do nothing for sequence loads (their energy constraint is skipped). Two wrong outcomes, both shown in the issue against master:

- window genuinely shorter than the sequence, hours/timesteps left at 0: no warning, the whole optimization just dies Infeasible (this is what @RikBast hit on #1047)
- window fine, a stale `operating_hours` value lying around: a false "Available timeframe is shorter" warning

## The fix

Where `is_sequence_load` and `sequence_length` are already in scope, the check now uses the quantity that actually governs feasibility:

```python
if is_sequence_load:
    min_steps = min(sequence_length, n)
elif def_total_timestep and def_total_timestep[k] > 0:
    min_steps = ceil(def_total_timestep[k])
else:
    min_steps = ceil(def_total_hours[k] / self.time_step)
```

Not `max()` with the hours-derived value: that keeps the false positive (a 5-step window for a 5-step sequence solves Optimal but would still warn if `operating_hours = 10` is set). Covered in the issue.

One refinement over the snippet I posted there: the length is clamped to the horizon. A sequence longer than the horizon is already truncated to `n` steps with its own warning, so `n` is all the window ever needs to fit. Without the clamp, that corner logged "Optimization will fail" with the untruncated length right next to the truncation warning, and then solved fine. There is a test pinning the clamp.

`validate_def_timewindow` gets an `is_sequence` keyword (default False) so the sequence case warns with wording a user can act on, naming the sequence length and the usable window. The non-sequence message is byte-identical to before, the clamping outputs do not depend on `min_steps` at all, and the six other call sites (which all discard the warning) are left alone.

- New warning: a sequence load with a window shorter than its sequence now warns before the solve (previously silent unless an unrelated hours value happened to trip it).
- Removed warning: a sequence load with an adequate window no longer warns just because `operating_hours` or `operating_timesteps` hold a larger stale value.
- Incidental: the sequence branch no longer reads `operating_hours` at all, so a `None` entry there can no longer raise a `TypeError` for a sequence load. The unguarded read I flagged in the issue still exists for non-sequence loads; kept out of this PR since you did not weigh in on it, happy to do it separately.
- Non-sequence loads: unchanged, warning text unchanged. Still a warning, not an error: the failure behaviour itself is untouched, per the issue.
- Known limit, same as the old check: it compares window width only, so a sequence with padding zeros at its edges can technically fit a narrower window than its full length. The warning can overclaim there; it errs on the loud side for a degenerate config.

Also not included: the `end_timesteps` inclusive/exclusive doc wording, which the issue already scoped out.

## Tests

- `test_sequence_load_short_window_warns_with_sequence_wording`: the #1047 shape (5-step sequence, 4-step window, hours at 0) now warns naming both numbers. Fails on master.
- `test_sequence_load_exact_window_with_stale_operating_hours_no_warning`: a 5-step window for a 5-step sequence with `operating_hours = 10` set solves Optimal, runs the sequence, and does not warn. Fails on master (false positive).
- `test_sequence_longer_than_horizon_window_check_clamped_to_horizon`: 12-step sequence at a 10-step horizon with a full-horizon window keeps the truncation warning, gets no window warning, and runs the truncated sequence.
- `test_nonsequence_load_short_window_warns_with_original_wording`: the existing hours-based warning for a standard load, original text. Passes on master and here.

Full suite green locally on top of f380180b.

## Summary by Sourcery

Correct deferrable-load window validation so sequence loads are checked against the number of timesteps required by their power sequence.

Bug Fixes:
- Derive sequence deferrable-load window validation from the sequence length, preventing missed warnings for undersized windows and false warnings from stale operating-hour settings.
- Provide sequence-specific warning text that identifies the available window and required sequence length.

Enhancements:
- Clamp the sequence length used for window validation to the optimization horizon while preserving existing behavior for non-sequence loads.

Tests:
- Add coverage for undersized sequence windows, stale operating-hour values, horizon-truncated sequences, and unchanged non-sequence warning behavior.

### DIFF
--- a/src/emhass/optimization.py
+++ b/src/emhass/optimization.py
@@ -3692,20 +3692,25 @@ class Optimization:
 
             # Time Window Logic
             # Calculate Valid Window
-            if def_total_timestep and def_total_timestep[k] > 0:
-                def_start, def_end, warning = Optimization.validate_def_timewindow(
-                    def_start_timestep[k],
-                    def_end_timestep[k],
-                    ceil(def_total_timestep[k]),
-                    n,
-                )
+            if is_sequence_load:
+                # A sequence load is placed in one piece, so the window must fit the
+                # whole sequence. Operating hours/timesteps play no role for sequence
+                # loads (their energy constraint is skipped above), so they must not
+                # drive this check either. Clamped to the horizon: a sequence longer
+                # than the horizon is truncated to n steps (warned above), so n is
+                # all the window ever needs to fit.
+                min_steps = min(sequence_length, n)
+            elif def_total_timestep and def_total_timestep[k] > 0:
+                min_steps = ceil(def_total_timestep[k])
             else:
-                def_start, def_end, warning = Optimization.validate_def_timewindow(
-                    def_start_timestep[k],
-                    def_end_timestep[k],
-                    ceil(def_total_hours[k] / self.time_step),
-                    n,
-                )
+                min_steps = ceil(def_total_hours[k] / self.time_step)
+            def_start, def_end, warning = Optimization.validate_def_timewindow(
+                def_start_timestep[k],
+                def_end_timestep[k],
+                min_steps,
+                n,
+                is_sequence=is_sequence_load,
+            )
             if warning is not None:
                 self.logger.warning(f"Deferrable load {k} : {warning}")
 
@@ -5822,7 +5827,7 @@ class Optimization:
 
     @staticmethod
     def validate_def_timewindow(
-        start: int, end: int, min_steps: int, window: int
+        start: int, end: int, min_steps: int, window: int, is_sequence: bool = False
     ) -> tuple[int, int, str]:
         r"""
         Helper function to validate (and if necessary: correct) the defined optimization window of a deferrable load.
@@ -5831,10 +5836,16 @@ class Optimization:
         :type start: int
         :param end: End timestep of the optimization window of the deferrable load
         :type end: int
-        :param min_steps: Minimal timesteps during which the load should operate (at nominal power)
+        :param min_steps: Minimal number of usable timesteps the window must contain:
+            derived from the operating hours for regular loads, or from the
+            (horizon-clamped) sequence length for sequence loads
         :type min_steps: int
         :param window: Total number of timesteps in the optimization window
         :type window: int
+        :param is_sequence: True when min_steps was derived from a sequence load's
+            power-sequence length rather than from operating hours/timesteps. Selects
+            the wording of the short-timeframe warning.
+        :type is_sequence: bool
         :return: start_validated: Validated start timestep of the optimization window of the deferrable load
         :rtype: int
         :return: end_validated: Validated end timestep of the optimization window of the deferrable load
@@ -5852,9 +5863,16 @@ class Optimization:
             start_validated = max(0, min(window, start))
             end_validated = max(0, min(window, end))
             if end_validated > 0:
-                # If the available timeframe is shorter than the number of timesteps needed to meet the hours to operate (def_total_hours), issue a warning.
+                # If the available timeframe is shorter than the number of timesteps the load needs (min_steps: from the operating hours, or the sequence length for a sequence load), issue a warning.
                 if (end_validated - start_validated) < min_steps:
-                    warning = "Available timeframe is shorter than the specified number of hours to operate. Optimization will fail."
+                    if is_sequence:
+                        warning = (
+                            f"Available timeframe ({end_validated - start_validated} timesteps) is "
+                            f"shorter than the load's power sequence ({min_steps} timesteps). "
+                            "Optimization will fail."
+                        )
+                    else:
+                        warning = "Available timeframe is shorter than the specified number of hours to operate. Optimization will fail."
         else:
             warning = "Invalid timeframe for deferrable load (start timestep is not <= end timestep). Continuing optimization without timewindow constraint."
         return start_validated, end_validated, warning
--- a/tests/test_optimization.py
+++ b/tests/test_optimization.py
@@ -1636,6 +1636,123 @@ class TestOptimization(unittest.IsolatedAsyncioTestCase):
         # The 2-step, 1000 W sequence ran exactly once: total power 2000 W.
         self.assertAlmostEqual(res["P_deferrable0"].sum(), 2000.0, delta=1.0)
 
+    def test_sequence_load_short_window_warns_with_sequence_wording(self):
+        """Issue #1081: a sequence (list-valued power) deferrable load's window
+        check must be sized against the length of its own power sequence, not
+        against operating hours/timesteps (which are meaningless for it, since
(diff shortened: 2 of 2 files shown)


LesIT1:
Checked the UnboundLocalError one and I don't think it's there: start_validated and end_validated are both initialised to 0 at the top of the function before any branching, so the invalid-timeframe path returns (0, 0, warning). That part predates this PR anyway, the diff only touched the warning selection. I also ran the start > end case directly and it returns cleanly.

Declining the two style points as well. The two message strings are each used once and the selection is three lines, so I'd rather keep them next to the condition than hoist them into constants. And the test substrings are deliberate: the warning naming the sequence length and the usable window is what #1081 asked for, so if that wording drifts I want the tests to notice.

davidusb-geek (approved):

PR #1107, opened 2026-09-03
feat: SOC-dependent charge power ceiling

### LINKED ISSUE #807: [Feature Request] Dynamic Battery Charge Power Limit Based on Forecasted SoC
Currently, EMHASS uses a fixed maximum battery charging power during optimization (for example P_batt_char_max = 8000 W). While this works in many cases, it does not reflect real-world battery behavior where the allowable charging power often depends on the battery State of Charge (SoC), temperature, voltage, or Battery Management System (BMS) constraints.

In my setup, the battery can technically charge at 8000 W, but only at lower SoC levels. As the battery fills up, the maximum accepted charge power gradually decreases. Because EMHASS assumes a constant maximum charging power, the optimizer may schedule charging too aggressively near high SoC levels, resulting in unrealistic plans that the inverter/BMS cannot actually follow.

Proposed Feature

Allow the optimizer to dynamically adjust the maximum battery charging power based on the forecasted SoC during each optimization timestep.

For example:

SoC 0–60% → 8000 W
SoC 60–90% → 6000 W
SoC 90–100% → 2500 W

This could be implemented as:

Piecewise power curve based on SoC
User-defined lookup table
Optional template/function for advanced users
Similar support for discharge limits as well
Why This Matters

Benefits would include:

More realistic optimization results
Better alignment with actual battery/BMS behavior
Improved charging schedules before sunrise or low-price windows
Less overestimation of available charging power at high SoC
Better battery longevity by respecting taper charging behavior
Example Use Case

If EMHASS predicts the battery will reach 92% SoC at 08:00, the optimizer should automatically know that full 8000 W charging is no longer possible, and instead use the reduced charge limit for subsequent timesteps.

Additional Note

Many hybrid inverters and batteries already internally taper charge current based on SoC and temperature. Integrating this behavior directly into the optimization model would significantly improve scheduling accuracy.

Sjeiz:
Closes #807, reported by @gieljnssns.

## The problem

A BMS derates its charge limit as the battery fills. The optimizer plans against
a single flat `battery_charge_power_max`, so it schedules charging the battery
cannot deliver: the plan reaches its target SOC in fewer timesteps than the
hardware needs, starts too late, and arrives under-charged.

Measured on a Dyness Tower T17 behind a GoodWe ET, from the current limit the BMS
publishes to the inverter over three months of five-minute samples:

| SOC | BMS limit | power |
|---|---|---|
| below 50% | 22 A | ~10.8 kW |
| 50-70% | 18 A | ~8.9 kW |
| 70-90% | 9 A | ~4.5 kW |
| above 90% | 5 A | ~2.5 kW |

The same levels appear when charging from the grid at night, so this is the
battery side, not the available PV.

One run here planned a fill from 15% to 90% in six 15-minute steps; against the
measured taper that takes just over nine.

## Why the existing runtime route does not cover it

`battery_charge_power_max` is already a `cp.Parameter` so that "SoC-derated
values arriving via runtimeparams update without invalidating the
OptimizationCache". A caller can therefore derate it per run.

That covers the current timestep, not the trajectory. `update_battery_power_limits`
takes `float(charge_list[k] or 0)`, one value per battery for the whole horizon,
while the derating depends on the SOC the plan itself produces:

- A battery at 20% that must fill plans every timestep at full power, including
  the ones the plan puts at 90% SOC. That is the original problem, unchanged.
- A battery at 85% caps a later refill starting from 30% at the top-band limit, a
  factor of four too low, and concludes the fill takes far longer than it does.

Only the first timestep is executed, but the decision taken there - start charging
now or later - follows from the rest of the horizon.

Letting `battery_charge_power_max` take a per-timestep series instead, the way
`weight_battery_charge` does, does not help: that series would have to be derived
from the SOC at each timestep, and the SOC trajectory is the optimizer's own
output. The limit depends on a decision variable, not on a timestep index, so it
has to be expressed inside the model.

## What this adds

Empty is the default, so behaviour is unchanged for every existing config.

`battery_charge_power_derating`, an optional `plant_conf` parameter: a list of
`[soc_threshold, power_max]` pairs, both percentage/100, ascending by SOC. Above
`soc_threshold` the charge power is capped at that much of
`battery_charge_power_max`; below the first threshold the flat maximum applies, so
the table lists only the derating steps. The measured table above becomes
`[[0.5, 0.84], [0.7, 0.42], [0.9, 0.23]]`. The planned charge trajectory then
follows the BMS limits more closely.

A cap relative to the maximum keeps the table valid when
`battery_charge_power_max` is retuned, and keeps that parameter usable at
runtime, since the ceiling multiplies the existing `cp.Parameter`.

To build a table for another battery, read the charge-current limit the BMS
reports to the inverter (here `sensor.goodwe_battery_charge_limit`) and take the
median per SOC band. Measure the current, not the power: the limit is regulated
in amperes, so the power moves with pack voltage and a curve fitted on power
fits two things at once. Levels that look like outliers in power resolve into
steps in current.

## Implementation

`optimization.py`: one additive block per battery, after the SOC recursion. A
boolean indicator per derating step, pinned to the SOC with the same big-M form
the recovery block below it uses (`cap` as M, `recovery_margin` so the two sides
cannot both hold), lowers the ceiling as the battery fills. The limit follows the
SOC a step *starts* at, so charging within a step cannot tighten that step's own
ceiling. The problem stays DPP, so the compiled problem is still reused.

`utils.py`: `check_batt_charge_derating` normalises the value against
`number_of_batteries`. Shape, not length, tells shared from per-battery: two
batteries sharing a two-row table would be split into one row each. Shape and
content are both checked, because a misordered table cannot raise the ceiling
above the flat maximum and would therefore behave like no table at all. A fault
is logged with the offending row and the table dropped rather than raised: unlike
the other battery parameters this one refines a ceiling that already exists, so a
typo costs the refinement, not the optimization.

## Solver cost

Ten runs per side on the same 96-step profile with a deferrable load active, at
the default `lp_solver_mip_rel_gap` of 0.01: `optim_solve` 0.2 s on stock
v0.18.2, 0.3 s with the table.

## Tests

`tests/test_battery_charge_power_derating.py`: 16 config-layer cases (shape,
per-battery broadcast, the N-rows-N-batteries collision, content faults, runtime
override) and 3 optimization cases (the ceiling binds at the SOC the step starts
at, the fill is spread over more timesteps, and the feature is inert without the
table).

## Summary by Sourcery

Model SOC-dependent battery charge-power ceilings so charging plans better match BMS derating behavior.

New Features:
- Add optional SOC-dependent battery charge-power derating tables so optimization plans follow BMS charge limits as batteries fill.

Bug Fixes:
- Prevent charge schedules from reaching target SOC too quickly by accounting for charge-power tapering across the planned SOC trajectory.

Enhancements:
- Validate and normalize shared or per-battery derating tables, logging and ignoring invalid values while retaining the flat charge-power ceiling.
- Keep the derating model compatible with runtime charge-power updates and reusable optimization problems.

Documentation:
- Document the new battery charge-power derating configuration, supported shapes, thresholds, and fallback behavior.

Tests:
- Add configuration and optimization coverage for table validation, broadcasting, runtime overrides, SOC-based limits, tapered charging, and unchanged behavior without derating.


Sjeiz:
Thanks — one taken, one I think is a false positive.

**Big-M gap: fixed in b606bf5.** You are right that `[threshold - recovery_margin, threshold)` satisfied neither indicator. The margin is copied from the recovery block, where it keeps two one-way switches from both firing; here it only carves a hole out of the feasible set. The bands now meet at the threshold, so every SOC falls in exactly one of them, and at the threshold itself the band below applies:

```python
constraints.append(soc_at_step_start >= threshold - cap * (1 - above[i]))
constraints.append(soc_at_step_start <= threshold + cap * above[i])
```

**NaN: already rejected.** The range check is negated, so a comparison that evaluates false is what triggers the rejection:

```python
if not 0 <= value <= 1:
    return f"row {position} has {name}={value}, expected a value between 0 and 1 ..."
```

`0 <= nan <= 1` is `False`, so `not False` is `True` and the row is refused with that message before it reaches the ordering checks or the model. A `math.isfinite` guard would be equivalent here, not additive.

Sjeiz:
Second one implemented anyway, in 4fbc874: `math.isfinite` before the range check.

It is redundant today, as above. The reason to add it is that the current rejection leans on nan comparing false against everything, which is not obvious from reading the code — rewrite `not 0 <= value <= 1` into `value < 0 or value > 1` and nan starts passing. The explicit guard survives that.

Sjeiz:
Correction on the first point: I reverted it in 99ace3b, because CI showed the suggested shape breaks the feature.

Letting the bands meet at the threshold makes both indicators feasible at the threshold itself, and the solver then has an incentive to hold the SOC exactly there and take the higher ceiling. `test_charge_power_respects_the_limit_for_the_soc_it_starts_at` caught it immediately: charging at the flat 5000 W maximum on a step starting at or above 0.5, where the table allows 0.4 of it.

The underlying constraint is that a strict inequality has no exact MILP form. Separating "below the threshold" from "at or above it" needs an epsilon, and that epsilon is precisely the interval the review flagged. Here it is `max(cap * 1e-6, 1e-3)`, so 0.018 Wh on a 17.76 kWh pack — one part in a million of capacity, and `soc_init` arrives rounded to three decimals, so it cannot land inside it.

So the finding describes a real property of the formulation, but it is the standard trade-off rather than something to remove. Happy to widen or narrow the epsilon if you would rather see it derived differently.

gieljnssns:
Could you also add this for discharging?

Sjeiz:
Based on what criteria?
En let’s first wait if David accepts this PR 😊

Regards,
Erik

davidusb-geek:
> Based on what criteria? En let’s first wait if David accepts this PR 😊 Regards, Erik From

Hi thanks for developing this. It was a long standing feature request that is useful for better modeling those last moments when approaching SOCmax.

I currently don't see how this is useful for discharging

gieljnssns:
This is how my discharge current looks in a template
```
        {% if soc < 5 -%}
          {% set discharge_soc = 0 -%}
        {% elif soc < 10 -%}
          {% set discharge_soc = 5 -%}
        {% elif soc < 20 -%}
          {% set discharge_soc = 5 -%}
        {% elif soc < 30 -%}
          {% set discharge_soc = 18 -%}
        {% elif soc < 40 -%}
          {% set discharge_soc = 18 -%}
        {% elif soc < 50 -%}
          {% set discharge_soc = 22 -%}
        {% elif soc < 60 -%}
          {% set discharge_soc = 22 -%}
        {% elif soc < 70 -%}
          {% set discharge_soc = 22 -%}
        {% elif soc < 80 -%}
          {% set discharge_soc = 22 -%}
        {% elif soc < 90 -%}
          {% set discharge_soc = 22 -%}
        {% elif soc < 100 -%}
          {% set discharge_soc = 22 -%}
        {% else -%}
          {% set discharge_soc = 22 -%}
        {% endif -%}
```

scruysberghs:
@gieljnssns  
Where/how do you use the result of that discharge_soc template? amps?

gieljnssns:
It's a part of my template
```
    - name: "Battery P Discharge Max"
      unique_id: 634b09f5-0a4b-4098-9c02-0720f7a31326
      unit_of_measurement: "W"
      device_class: power
      state: >
        {% set soc = states('sensor.battery_state_of_charge') | float(states('sensor.emhass_soc')|float) -%}
        {% set temp = states('sensor.battery_temperature') | float(10) -%}
        {% set voltage = states('sensor.battery_voltage') | float(501) -%}
        {% set remaining = state_attr("sensor.pv_power_forecast_list_interpolated", "remaining") | int -%}

        {#- Bepaal de ontlaadstroom op basis van SOC -#}
        {% if soc < 5 -%}
          {% set discharge_soc = 0 -%}
        {% elif soc < 10 -%}
          {% set discharge_soc = 5 -%}
        {% elif soc < 20 -%}
          {% set discharge_soc = 5 -%}
        {% elif soc < 30 -%}
          {% set discharge_soc = 18 -%}
        {% elif soc < 40 -%}
          {% set discharge_soc = 18 -%}
        {% elif soc < 50 -%}
          {% set discharge_soc = 22 -%}
        {% elif soc < 60 -%}
          {% set discharge_soc = 22 -%}
        {% elif soc < 70 -%}
          {% set discharge_soc = 22 -%}
        {% elif soc < 80 -%}
          {% set discharge_soc = 22 -%}
        {% elif soc < 90 -%}
          {% set discharge_soc = 22 -%}
        {% elif soc < 100 -%}
          {% set discharge_soc = 22 -%}
        {% else -%}
          {% set discharge_soc = 22 -%}
        {% endif -%}

        {#- Bepaal de ontlaadstroom op basis van temperatuur -#}
        {% if temp < -10 -%}
          {% set discharge_temp = 0 -%}
        {% elif temp < -5 -%}
          {% set discharge_temp = 6 -%}
        {% elif temp < 0 -%}
          {% set discharge_temp = 9 -%}
        {% elif temp < 5 -%}
          {% set discharge_temp = 14 -%}
        {% elif temp < 10 -%}
          {% set discharge_temp = 22 -%}
        {% elif temp < 40 -%}
          {% set discharge_temp = 22 -%}
        {% elif temp < 45 -%}
          {% set discharge_temp = 22 -%}
        {% elif temp < 50 -%}
          {% set discharge_temp = 18 -%}
        {% elif temp < 55 -%}
          {% set discharge_temp = 18 -%}
        {% else -%}
          {% set discharge_temp = 0 -%}
        {% endif -%}

        {#- Neem het minimum van beide -#}
        {% set max_current = [discharge_soc, discharge_temp] | min -%}

        {#- GoodWe GW8K-BT omvormer limiet: max 22A ontlaadstroom -#}
        {% set inverter_max_discharge = 22 -%}
        {% set final_current = [max_current, inverter_max_discharge] | min -%}

        {#- Converteer naar Watt (P = U x I) -#}
        {% set power = final_current * voltage -%}
        {#- Limiet op 8000W AC (omvormer max) -#}
        {% set limit =  [power, 8000] | min | round(0) -%}
        {{ limit if remaining < 30 else 8000 }}
```

davidusb-geek:
I still don't get it @gieljnssns. 
I think that you are mixing two concepts that should be regarded separetly:
- On one hand the new set of hard contraints that are added in this PR to account for typical charging strategies behavior, ie. charge power decreases as you approach SOCmax
- On the other hand how you actually control your battery discharge based on the provided optimal SOC trajectory by emhass, ie. the set of automations you shared

So I don't think that there is any missconception with this  PR.
Merging...

davidusb-geek (approved):

PR #1117, opened 2026-09-11
fix: stabilize results UI (#1104)

WartoWiedziec:
Fixes #1104.

Successful optimization actions return HTTP 200 with a plain-text body. The results UI only recognizes 201, then attempts to parse that text as JSON. The exception leaves the loading indicator active and prevents the results from refreshing.

This change:
- Accepts all successful HTTP 2xx responses without parsing their bodies as JSON.
- Handles HTTP and network failures and clears the loading indicator in finally.
- Centers the existing Plotly containers.
- Synchronizes the cloned first header cell during horizontal scrolling without applying both sticky positioning and counter-scrolling.
- Adds regression tests for action response handling.

Thanks to @MMicieli for identifying the precise JSON-parsing failure and suggesting the regression test. This follows the maintainer's invitation in #1104. The configuration issue #1115 is being handled separately.

Validation:
- JavaScript syntax check passed.
- 11 regression tests passed using Node's built-in test runner.
- With the original source, 6 tests failed and 5 passed.
- No new runtime dependencies.

Outstanding before marking ready for merge:
- [ ] Chrome and Firefox: chart centering, horizontal and vertical table scrolling, resizing and repeated result refresh.
- [ ] Smoke test on the actual EMHASS page.
- [ ] Project CI checks.

Browser validation remains pending because browser engines were unavailable in the preparation environment and their download timed out. The full Python test suite was not rerun.

AI-assisted implementation and tests. Submitted as a draft pending browser validation.

## Summary by Sourcery

Stabilize optimization result updates and scrolling presentation across successful actions, failures, and repeated refreshes.

Bug Fixes:
- Handle all successful HTTP 2xx action responses without parsing response bodies as JSON, refresh results reliably, and clear loading state after HTTP, network, or refresh failures.
- Improve sticky table headers during horizontal scrolling and repeated result refreshes without accumulating listeners or conflicting positioning.
- Center Plotly result charts within their containing rows.

Tests:
- Add regression coverage for successful, failed, and network action responses, result refresh failures, loading-state cleanup, and sticky-table lifecycle and geometry behavior.

Issue #1104, opened 2026-09-01
Web UI regressions: endless optimization spinner, plot alignment and sticky table header


WartoWiedziec:
Description

I found three reproducible issues in the EMHASS web interface:

The loading spinner does not stop after a successful Day-ahead Optimization.
Optimization plots remain aligned to the left instead of using or centering within the available page width.
The sticky optimization-results table header becomes misaligned during horizontal scrolling.

The same behavior occurs in both Chrome and Firefox, so it does not appear to be browser-specific.

Environment
EMHASS version: 0.18.1
Installation type: Home Assistant add-on
Home Assistant installation: HAOS
Browsers tested: Chrome and Firefox
1. Endless spinner after Day-ahead Optimization
Steps to reproduce
Open the EMHASS web interface.
Click Day-ahead Optimization.
Wait for the optimization to complete.
Actual behavior

The optimization completes successfully and the results are generated, but the loading spinner continues indefinitely.

The EMHASS logs confirm that the optimization finishes successfully. The action endpoint returns HTTP 200, while the current frontend code in static/script.js treats only HTTP 201 as a successful response.

This appears to be a frontend regression related to PR #925, which intentionally changed JSON API endpoint responses from HTTP 201 to HTTP 200, while the corresponding frontend success check was apparently not updated.

Expected behavior

The spinner should stop after any successful HTTP 2xx response and the optimization results should be displayed or refreshed.

Using response.ok instead of checking exclusively for status 201 would support the current HTTP 200 response as well as other valid 2xx responses.

The spinner should also be cleared when:

the server returns an HTTP error,
the response cannot be parsed,
a network error occurs.
2. Optimization plots are aligned to the left
Actual behavior

On a wide screen, the Plotly charts occupy only the left part of the available area and leave a large unused area on the right.

This is especially visible when the EMHASS interface is displayed in a wide Home Assistant panel.

Expected behavior

The plots should either:

use the available width responsively, or
remain at their current maximum width but be centered within the page.

A small CSS adjustment to the Plotly graph container appears sufficient. It would not require changing the chart data or the optimization backend.

3. Sticky table header becomes misaligned during horizontal scrolling
Steps to reproduce
Open the latest optimization results table.
Verify that the sticky header is initially aligned correctly.
Scroll the wide table horizontally.
Actual behavior

[...]

PR #1118, opened 2026-09-11
fix: preserve nested configuration arrays (#1115)

### LINKED ISSUE #1115: Web UI list view cannot save battery_charge_power_derating (array.array.float is flattened to a list of strings)
**Describe the bug**
battery_charge_power_derating is defined as array.array.float in param_definitions.json, but the configuration page's list view cannot produce a nested list. It renders one plain text box per entry and, on save, pushes every box value into a single flat list of strings. The result is rejected by check_batt_charge_derating() in utils.py, so the table is silently dropped and charging falls back to the flat battery_charge_power_max.

**Two failure modes, both from the list view:**

1. Two boxes with 0.5 and 0.84 are saved as ["0.5", "0.84"]:
ERROR in utils: battery_charge_power_derating: must be a list of [soc_threshold, power_max] pairs, got ['0.5', '0.84']. Ignoring it: charging is limited by the flat battery_charge_power_max instead.
2. One box with [0.5, 0.84] is saved as ["[0.5, 0.84]"]:
ERROR in utils: battery_charge_power_derating: must be a list of [soc_threshold, power_max] pairs, got ['[0.5, 0.84]']. Ignoring it: charging is limited by the flat battery_charge_power_max instead.

There is no way to add a nested row in the list view. The plus button only adds another flat text box.

**To Reproduce**
1. Open the EMHASS web UI, go to Configuration, list view.
2. Under Battery, find "Charge power derating" (battery_charge_power_derating).
3. Press +, enter 0.5 in the first box and 0.84 in the second box. Save.
4. Trigger any optimization (for example naive-mpc-optim).
5. Observe the error above in the log. The derating table is ignored.

Repeat with a single box containing [0.5, 0.84] for the second failure mode.

**Expected behavior**
Either of:
- The list view lets the user add rows of [soc_threshold, power_max] pairs and saves them as nested lists of numbers, e.g. [[0.5, 0.84]].
- Or the list view parses the array.array.float text box as JSON on save, the way it already does for the object input type.

At minimum, the config UI should not write a value that the backend immediately rejects.

**Screenshots**
Image
Image
Image

**Root cause (from reading src/emhass/static/configuration_script.js)**
- In the render switch, case "array.array.float" only sets type = "text" and a JSON placeholder. It then falls into the generic array branch, which emits one <input> per element.
- In the save loop, only input === "object" goes through JSON.parse. Everything else hits the default: branch of the switch (input.type) and does param_values.push(input.value), so text boxes become strings and the parameter is saved as param_values flat.
- The backend validator check_batt_charge_derating() correctly requires current[0] to be a list or tuple, so the flat list is rejected.

**Workaround**
Switch the configuration page to box (JSON) view and set the value there, e.g.
"battery_charge_power_derating": [[0.5, 0.84], [0.7, 0.42], [0.9, 0.23]]
Saving from box view works. Saving from list view afterwards flattens it again, so the list view must be avoided while this parameter is set.

- Home Assistant Supervised

**Your hardware**
- OS: [Linux]
- Architecture: [amd64]

**EMHASS installation type**
- [Add-on]

**Additional context**
EMHASS version: [0.18.3]. The parameter and its validator were added for issue #807, so this affects every release that includes that change. The docs in docs/config.md describe the nested format correctly. The gap is only in the list view of the web UI.

WartoWiedziec:
Fixes #1115.

The configuration list view flattens array.array.float values into text inputs and saves strings. This corrupts battery_charge_power_derating, so the backend rejects the table and falls back to the flat charging limit.

This change:
- Renders the complete nested value in one JSON input.
- Preserves empty arrays, shared 2-D tables and per-battery 3-D tables.
- Removes generic +/- controls for this input type.
- Escapes rendered values and blocks malformed JSON, flat arrays and nonnumeric values before submission.
- Saves an empty array when the field is blank.

Backend validation of thresholds, row sizes and battery counts remains authoritative. No backend, optimization or schema changes are included. The related OpenAPI representation gap is outside this UI fix.

Thanks to @anton4 for the report and @MMicieli for checking the supported shared and per-battery formats. This follows the maintainer's request in #1104. The results-page fixes are submitted separately in #1117.

Validation:
- JavaScript syntax check passed.
- 13 regression tests passed using Node's built-in test runner.
- With the original source, 12 tests failed and 1 passed.
- Tests cover rendering, serialization, clearing the field, invalid input, HTML escaping and unchanged flat string-array serialization.
- No new runtime dependencies.

Outstanding before marking ready for merge:
- [ ] Chrome and Firefox: list/JSON view round-trip and correct +/- controls.
- [ ] Actual configuration save/reload with shared and per-battery tables.
- [ ] Project CI checks.

Browser validation remains pending because browser engines were unavailable in the preparation environment and their download timed out. The full Python test suite was not rerun.

AI-assisted implementation and tests. Submitted as a draft pending browser validation.

## Summary by Sourcery

Preserve nested battery derating configuration tables through the UI without changing backend validation or schema behavior.

Bug Fixes:
- Preserve battery charge power derating tables as nested numeric arrays when configuration values are rendered and saved.
- Reject malformed, flat, nonnumeric, or non-finite battery derating values before submission while allowing blank fields to clear the table.

Enhancements:
- Render battery derating tables in a single escaped JSON input without generic array controls.
- Add regression coverage for shared and per-battery table round trips, clearing, validation, escaping, repeated serialization, and unaffected array parameters.

Tests:
- Add Node-based UI regression tests covering nested configuration array rendering and serialization.

### DIFF
--- a/src/emhass/static/configuration_script.js
+++ b/src/emhass/static/configuration_script.js
@@ -323,6 +323,7 @@ function buildParamContainers(
     let array_buttons = "";
     if (
       parameter_definition_object["input"].search("array.") > -1 &&
+      parameter_definition_object["input"] !== "array.array.float" &&
       section != "Deferrable Loads" &&
       !(section == "Battery" && BATTERY_ARRAY_PARAMS.includes(parameter_definition_name))
     ) {
@@ -509,6 +510,14 @@ function buildParamElement(
   //definitions default value is used if none is found in the configs, or an array element has been added in the ui (deferrable load number increase or plus button pressed)
   //check if a param value is saved in the config file (if so overwrite definition default)
   let value = checkConfigParam(placeholder, config, parameter_definition_name);
+
+  // Keep the complete nested value in one JSON input, including per-battery tables.
+  if (parameter_definition_object["input"] === "array.array.float") {
+    const json = typeof value === "string" ? value : JSON.stringify(value ?? []);
+    const escaped = json.replaceAll('&', '&amp;').replaceAll('<', '&lt;')
+      .replaceAll('>', '&gt;').replaceAll('"', '&quot;');
+    return `<input class="param_input" type="text" placeholder="[]" value="${escaped}">`;
+  }
 
   //generate and return param input html,
   //check if param value is not an object, if so assume its a single value.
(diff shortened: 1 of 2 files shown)


WartoWiedziec:
@sourcery-ai review

davidusb-geek (approved):

PR #1123, opened 2026-09-14
fix: scope deferrable load UI cardinality (#1116)

### LINKED ISSUE #1116: Configuration list view: number_of_deferrable_loads grows/shrinks every .param in the section, including non-per-load ones, and dynamically-added rows don't pick up a true default
While reading through `configuration_script.js` on current master (`f4fd0cf`) I noticed two related gaps in how the "Deferrable Loads" section grows/shrinks its rows when `number_of_deferrable_loads` changes. Both come from the same place — the handler for that field doesn't have an allow-list of which params are actually per-load — so I'm putting them in one issue rather than two.

### 1. `heat_topology` gets grown/shrunk along with the genuinely per-load params

The `number_of_deferrable_loads` case in the header-input switch does this:

```js
case "number_of_deferrable_loads":
  param_list = param_container.getElementsByClassName("param");
  ...
  for (const param of param_list) {
    plusElements(param.id, param_definitions, "Deferrable Loads", {});
  }
```

`param_list` is every `.param` element currently rendered in the "Deferrable Loads" section — that includes `heat_topology`, whose `param_definitions.json` entry is `"input": "object"` (a single declarative graph, `default_value: null`, documented in `docs/heat_topology.md`), not a per-deferrable-load array. `plusElements()` doesn't check the input type before appending — it just calls `buildParamElement()` and inserts the result into that param's own `.param-input` container. So incrementing `number_of_deferrable_loads` appends an extra text input inside `heat_topology`'s single-object box, and decrementing removes it again (symmetric, since `minusElements()` always removes the *last* input). The extra input is silently ignored on save (`saveConfiguration()` only reads `param_values[0]` for `input === "object"`), so this doesn't corrupt a saved config, but the list view visibly grows a second (then third, ...) input under what's meant to be one JSON graph object as the load count goes up, which looks like a real defect from the UI side.

I compared this to the existing `Battery` section, which handles the equivalent problem for `number_of_batteries` with an explicit allow-list:

```js
const BATTERY_ARRAY_PARAMS = [ ... ];
...
case "number_of_batteries":
  param_list = Array.from(param_container.getElementsByClassName("param"))
    .filter((p) => BATTERY_ARRAY_PARAMS.includes(p.id));
```

So there's already a precedent in the codebase for scoping this kind of grow/shrink to only the params that are actually cardinality-N. I'm not proposing a `DEFERRABLE_ARRAY_PARAMS` constant as the only acceptable shape — an allow-list mirroring `BATTERY_ARRAY_PARAMS` is one small option, deriving which params are per-load from the schema (e.g. a flag on the `array.*` definitions) is another, and there may be other params in that section with the same problem that I haven't checked as carefully as `heat_topology`.

### 2. A dynamically-added `array.boolean` row doesn't get its declared `true` default

Separately, in `buildParamElement()`:

```js
case "array.boolean":
case "boolean":
  ...
  placeholder = parameter_definition_object["default_value"] === "true";
```

Every boolean/`array.boolean` entry in `param_definitions.json` has a real JSON boolean as `default_value` (e.g. `is_electric_load: true`), never the string `"true"` — so this comparison is `true === "true"`, which is always `false`, for every boolean parameter, not just `is_electric_load`.

This doesn't show up on the *initial* page render for a param that's already in the served config (which is the normal case — `config_defaults.json` has `is_electric_load: [true, true]` for the default `number_of_deferrable_loads: 2`): `checkConfigParam()` overrides the broken `placeholder` with the real value from `config` before the input is rendered, and the checkbox's `checked` state is then fixed up by a one-time pass at the end of `buildParamContainers()`:

```js
let checkbox = SectionContainer.querySelectorAll("input[type='checkbox']");
checkbox.forEach((answer) => { answer.checked = (answer.value === "true"); });
```

But a row added later via the `number_of_deferrable_loads` `+` button goes through `plusElements()`, which calls `buildParamElement()` with an explicitly empty config (`{}` — there's a comment noting this is intentional, so a freshly-added row starts from the schema default rather than copying an existing row's saved value). With no config entry to override it, the row falls back entirely to the broken `placeholder` computation above, so `is_electric_load`'s declared default of `true` never reaches it. On top of that, the `checkbox.forEach` fixup above only runs once, at the point the whole section is first built — `plusElements()` doesn't re-run it — so even if the placeholder bug is fixed, a dynamically-added checkbox still wouldn't get its `checked` state set from `.value` unless that pass (or an equivalent) also runs after an add.

I've traced this at the source level rather than reproduced it in a browser, so I'd treat the description above as the more reliable part and the "does it visibly render unchecked" framing as something worth confirming in a browser before relying on it.

### Non-goals
- Not proposing a specific fix shape for either part — happy to send a PR once there's rough agreement on direction, or to split this into two issues if you'd rather track them separately.
- Not claiming other `array.*` params in "Deferrable Loads" have the same `heat_topology`-style problem — I only checked that one because it's the one whose semantics (single object, not per-load) make the mismatch obvious.
- Not touching `optim_conf`/backend behaviour — this is UI-only, the same way #1115 is.

### To reproduce (heat_topology growth)
1. Open the Configuration page, list view, "Deferrable Loads" section.
2. Note `heat_topology` renders one text input.
3. Increase `number_of_deferrable_loads` by 1 (e.g. 2 → 3).
4. Observe `heat_topology` now has two text inputs instead of one.

MMicieli:
Fixes #1116.

## Summary

The Configuration list view treated every `.param` in **Deferrable Loads** as cardinality-N when `number_of_deferrable_loads` changed. That incorrectly grew/shrank `heat_topology`, which is one `object` input rather than a per-load array. The same path also mishandled real JSON boolean defaults for dynamically-added `array.boolean` rows.

A pre-submission audit found one additional part of the same cardinality contract: EMHASS supports `number_of_deferrable_loads = 0`, but the list UI enforced `min="1"`, generic row removal retained the final row, and zero-row arrays were omitted on save instead of being serialized as `[]`.

This PR keeps the fix within the existing frontend contract:

- adds `DEFERRABLE_ARRAY_PARAMS`, mirroring the existing `BATTERY_ARRAY_PARAMS` / `number_of_batteries` pattern;
- grows/shrinks only genuine per-load fields and leaves `heat_topology` as exactly one object input;
- allows `number_of_deferrable_loads` to reach 0 while leaving `number_of_batteries` at its existing minimum of 1;
- gives `minusElements()` an optional minimum-input floor, defaulting to the existing 1, with only the Deferrable count-managed path requesting 0;
- serializes a rendered zero-row array as `[]`, keeping count 0 coherent through save/reload;
- recognizes real JSON boolean defaults and renders checkbox `checked` state immediately for dynamically-added rows;
- keeps generic array controls and battery cardinality behavior unchanged;
- keys generic array controls off the explicit Deferrable/Battery cardinality allow-lists instead of a section-wide Deferrable assumption.

No backend, optimizer, schema/default, OpenAPI, Home Assistant, or device-control behavior is changed.

## Regression coverage

`tests/test_configuration_ui_deferrable.py` is pytest-discovered and uses Node against the real `configuration_script.js`, `param_definitions.json`, and configuration-list template.

Independent tests cover:

- Deferrable count minimum 0 and Battery minimum remaining 1;
- the Deferrable cardinality allow-list and `heat_topology` exclusion;
- grow and shrink-to-zero touching only managed per-load fields;
- generic `minusElements(param)` retaining one row;
- explicit `minusElements(param, 0)` removing the final row;
- zero-input array serialization as `[]`;
- dynamically-added true/false boolean defaults;
- saved boolean values overriding defaults;
- `heat_topology` remaining single-input across grow/shrink.

`tests/test_schema_contract.py` receives only the minimal existing-probe harness update needed because `buildParamContainers()` now references both cardinality allow-list constants.

The dedicated `heat_topology` regression test mutates simulated per-parameter row counts. Managed fields must move 2 → 3 → 0, while `heat_topology` must remain exactly one input.

The strengthened test was also run unchanged against unmodified upstream `f4fd0cfd54f08d82d7cc4cec310b9e110894e6d1` and fails there for the original defect:

`heat_topology must stay at 1 input after growing to 3 (2 !== 1)`

## Review remediation

The initial review correctly identified that the dedicated `heat_topology` regression test used no-op grow/shrink mocks, meaning that specific test could remain green even if `headerElement()` incorrectly targeted `heat_topology`.

The test now:

- mutates simulated row counts through the mocked grow/shrink operations;
- asserts managed per-load fields move 2 → 3 → 0;
- mutates `heat_topology` too if it is incorrectly routed through those operations;
- asserts `heat_topology` remains exactly one input.

No production code changed as part of this review remediation.

## Validation

Validated on final commit:

`609dc04e3cfca808127c899cc9dd9d07a3ab5e33`

Local validation:

- `node --check src/emhass/static/configuration_script.js` — pass;
- focused #1116 regression tests — 12 passed, 0 failed;
- `tests/test_schema_contract.py` — 14 passed, 0 failed;
- `uvx ruff check .` — pass;
- `uvx ruff format --check --diff` — pass;
- `git diff --check` — pass;
- full `uv run pytest` suite — 1200 passed, 34 xfailed, 0 failed.

The full local suite used an isolated temporary directory so it did not inherit unrelated machine-global Solcast daily-counter state. No source or persistent runtime data was modified for that isolation.

## Upstream GitHub Actions

All checks completed successfully on `609dc04e3cfca808127c899cc9dd9d07a3ab5e33`, including the test matrix, CodeQL analysis, code-quality checks, Docker/image builds and OSV security scans.

## Scope / compatibility

- no backend or optimizer changes;
- no `param_definitions.json`, `config_defaults.json`, or OpenAPI changes;
- no new dependency or frontend framework;
- no Home Assistant or production EMS changes;
- no #1115/#1118 nested-battery-array behavior included;
- nullable `cost_forecast_per_deferrable_load` editor behavior remains a separate follow-up.

PR #1118 is semantically independent but currently edits the same `buildParamContainers()` condition for `battery_charge_power_derating`. If it merges first, this branch will be rebased, the two exclusions combined, and validation rerun before merge.

## Summary by Sourcery

Fix deferrable-load configuration handling for per-load cardinality, zero-load configurations, and dynamically added boolean defaults.

Bug Fixes:
- Correct deferrable-load configuration cardinality so only per-load fields resize with the configured load count while heat_topology remains a single input.
- Allow zero deferrable loads and preserve zero-row deferrable arrays as empty lists when saving.
- Render dynamically added boolean fields with the correct default checked state.

Enhancements:
- Scope generic array controls to explicit deferrable-load and battery cardinality lists while preserving existing battery and generic array behavior.

Tests:
- Add regression coverage for zero-load handling, cardinality scoping, heat_topology stability, array serialization, and boolean defaults.


davidusb-geek:
Hi @MMicieli, could you take a look at that current conflict? Thank you

davidusb-geek (approved):

PR #1129, opened 2026-09-15
fix: correct battery derating example and Plotly centering selector

WartoWiedziec:
This PR addresses two follow-up points raised by @MMicieli after #1117 and #1118:

* Correct the battery derating validation example from `[[0.5, 1000]]` to `[[0.5, 0.84]]`, consistent with the documented per-unit values.
* Update the Plotly centering selector to `#template > .table_div .plotly-graph-div`, allowing an intermediate wrapper between the results container and the graph.

The changes affect one validation-message example and one CSS selector. Optimization logic and configuration validation remain unchanged.

Validation:

* JavaScript syntax check passed.
* All 18 existing nested-configuration tests passed locally after the example correction.
* The corrected selector matched a representative HTML fixture containing an intermediate wrapper; the previous selector did not.
* Visual validation in a running EMHASS instance remains pending.

The earlier CI run failed on `test_perform_naive_mpc_optim` on all three platforms, comparing `0.5000623695643236` with `0.5` to 7 decimal places. This has been reported in a separate comment for comparison against the base branch. The CodeCov failure has not yet been diagnosed.

Related to #1104 and #1115. Thanks @MMicieli for identifying these follow-up points.


WartoWiedziec:
Hi @davidusb-geek,

The Python test jobs on Linux, Windows and macOS all fail on the same test:

`tests/test_optimization.py::TestOptimization::test_perform_naive_mpc_optim`

The assertion compares `0.5000623695643236` with `0.5`, requiring agreement to 7 decimal places. The absolute difference is approximately `6.24e-5`. All other tests pass, apart from the reported skips and expected failures.

This PR only corrects the example in a frontend validation message, from `[[0.5, 1000]]` to `[[0.5, 0.84]]`; it does not change optimization logic.

Could you check whether the same failure occurs on the base branch? I have left the assertion unchanged, since the logs alone do not establish whether this is a solver tolerance issue or an optimization regression.

The CodeCov job also failed, but I have not confirmed whether it has the same cause.

Thanks!

davidusb-geek:
Sometimes some runners will fail randomly, but everything seems good now

